Issue #4, opened 2018-02-02
Components inside table


jungrafael:
## Issue
Could we have a way of adding our own components to the table? We could just add it to the data variable. With the same API, we would be able to add text or a component.

## Your Environment

| Tech         | Version |
|--------------|---------|
| mui-datatables  | 1.0.7 |
| node        | 8.7.0 |
| OS      | Ubuntu 17.10 |

jungrafael:
okay, this is already possible!
![image](https://user-images.githubusercontent.com/5156163/35735313-8be4a196-0824-11e8-83b0-7b0f96979546.png)

Should we update the README, to tell about this awesome possibility?

gregnb:
Hah, I never tested it for that purpose. Well that is interesting! Let me look into more test cases to make sure that this works like you would intend it to before putting it in the README

jungrafael:
Hey, I've just found a drawback with this approach: we're unable to search items which are a JSX component, since it calls toString() on the component and this returns [object Object].
Should I reopen this issue to change the way the search/filtering is made?

jungrafael:
Hello @gregnb, I've just found a drawback with this approach: we're unable to search items which are a JSX component, since it calls toString() on the component and this returns [object Object].
Should I reopen this issue to change the way the search/filtering is made?

jungrafael:
Since I needed to solve this quite quickly, I've created a quick fix here: https://github.com/jungrafael/mui-datatables/tree/text-search-on-components
This could be of help for fixing this issue :)

gregnb:
Would you want to add some unit tests and send me this as a PR? Nice work @jungrafael

gregnb:
Hey @jungrafael so I took your function out and imported it into another project just to test. Take for example the following component

```
class Test extends React.Component {
 render() {
   return (
      <div>
        <div>this is the text to print</div>
      </div>
   );
 }
}
```

All I'm going to see for a return from your function is:
```
[object HTMLDivElement]
```

To be honest I'm not sure how you're going to extract or return the text of a component without rendering it first

jungrafael:
Hey @gregnb, this use case would not actually supported. This function I implemented just extracts the text inside nested elements, through navigating to its children. So you would need to use this function this way:
```javascript
getElementText(
  <div>
    <div>this is the text to print</div>
  </div>
);
```
This is exactly my use case for it on this library, since it's being used on an array of arrays of React Elements.
[...]

PR #10, opened 2018-03-12
Made all functionality work with components, updated example to show

tgBryanBailes:
Made it so that DOMServer renders a component, then uses textContent to strip out all of the text for search, filter, and CSVExport. Updated example to demonstrate using a component for inline editing, adhering to Material guidelines.


gregnb:
Hey Bryan, thanks for opening this PR. I'm on vacation currently but I'm skimming through your commits so I'll try to get to this when I get some free time. I've never used DOMServer in the client so I'm wondering what your experience has been?

I know it's allowed in browser but I'm looking out on the web for people's experiences and not finding much. I think it's a good solve and I thought about it myself as a way to get the pure text of any component but felt like ok this should only be on the server side, you know?

tgBryanBailes:
HI gregnb, I did all testing within the browser and all seemed to work fine as long as you stick to `renderToString()` and `renderToStaticMarkup()`

gregnb:
hey @jungrafael I know you did some work on #4 to extract text out of components.  What do you think of this here? I'm going to test this when I get some free time over the next week but this could be a better solve?

gregnb:
@tgBryanBailes Took a little bit of time to start going through this and I've found a couple of things that are a little concerning. 

-------------------

I've tried playing around with the text extraction and to me there's a noticeable lag every time I hit a key in the search box using the demo example.  This data set is small so each key should respond quickly: 

**Steps to recreate:**
1) Load your example
2) Click search
3) Monitor performance and enter word "business analyst"

**ReactDOMServer version:**
screen shot 2018-03-13 at 12 53 08 pm

**Current version without ReactDOMServer:**
screen shot 2018-03-13 at 12 52 01 pm


This issue needs to be investigated some more

-------------------

Also, the results returned when searching for "Business Analyst" do not seem correct:
screen shot 2018-03-13 at 12 42 39 pm
 
-------------------

Typography component is gone from the title in the toolbar:

**Actual:**
screen shot 2018-03-13 at 12 22 48 pm

**Expected:**

screen shot 2018-03-13 at 12 25 55 pm

While it's a minor issue I believe we need a separation of concerns for addressing the two issues: text extraction when components are provided for cells and unused vars/cleanup work. Some of the unused vars you found were good but it would be better to review this as **two separate PR's**.

tgBryanBailes:
Hi gregnb, I believe that the missing Typography component was a result of bumping the version of Material-UI to the latest beta, not removing unused vars. I also have some ideas for improving the performance of the text parsing. As for the search results, I do not currently have the select component updating the underlying data set, so if values are changed they will not appear correctly in the search results. I assumed that any changes to underlying data would happen in a data store such as Redux in the real world, so did not implement it in the example. I will work through these issues today, so expect a commit.

gregnb:
@tgBryanBailes Ah, ok my mistake then. I'll be in between flights today but when I get some downtime I'll pull down the latest changes and test them

jungrafael:
Hi, this seems to be an interesting approach, since it allows to search the text inside rendered components as well. I'll take a better look at it, but my only concern is only with performance.

gregnb:
@jungrafael Same but @tgBryanBailes has made some updates. I'm going to test some more to see the performance increase but this could be a solution to your ticket as well

gregnb:
hey @tgBryanBailes so I pushed some changes into your branch.  Your changes worked nicely addressed the issues when searching but I found another performance issue. When the data set gets increased to something larger (in my example, 1800 rows) it takes a noticeable amount of time to load. A lot of time gets spent in getText()

Try running what I pushed up and maybe even increasing the data set even more to 5000.  Then try and change getText to not use DOMServer to see the difference in speed.  Not sure how we could address this problem. If you can find a way, I think we're really close to merging this

tgBryanBailes:
You can go ahead and close this out. I am going to take your latest version and have a new idea for implementing this functionality. Expect a commit from me later today.

gregnb:
Ok @tgBryanBailes sounds good!

PR #26, opened 2018-04-03
Component Functionality

tgBryanBailes:
added renderComponent and renderValue options with an example; updated sortCompare to work with booleans, dates, and unicode; removed unused vars


gregnb:
Ok so I've started reviewing this PR and the 'renderValue' property is similar to a story I was going to do to allow custom displays of values.

I want to change `renderValue` to `format`:
```
  {
        name: "Age",
        options: {
          filter: false,
          format: (value) => value + "some custom string"
        }
   },
```

This piece to me is good to go with that change.  

------


With `renderComponent` I'm wondering what is going to be the process if I select an item from this drop down like so:

screen shot 2018-04-03 at 8 02 24 pm

If I select a new value say "Mesa" then our change={event => value} will fire but what about updating the table with that data to reflect the changes? I could see this working if you take the changed value and update the data store outside the library in the user space. Then that updated data store will trigger a complete re-render of MUIDataTable as the provided data={dataStore} prop has changed.

In the previous implementation you were trying to extract that string value from the component itself by rendering it but that was an expensive operation. The positive side on that was the library was able to always grab the latest value from a component.  The current implementation seems to be more along the lines of a controlled component where we'll update the data store and look to trigger a complete re-render of the datatables library.  Do I have that correct? 

I would also simplify `renderComponent` to just `component` but yeah would love your feedback on my comments

tgBryanBailes:
`renderValue` has a bit of a dual purpose, as shown in the example with the Active component. Not only can it format your data, but it also provides a searchable/filterable value for rendered components. That said, calling it `format` is fine by me if you feel it is appropriate for both use cases.

I have updated the component example to show the values updating. This does trigger a render as the provided data={dataStore} prop has changed. I am not so sure there's another way to do it, but am open to suggestions.

gregnb:
With the updated example this is a lot more clear 👍 so we're on the same page. I think this is great because it's following the along the lines of a controlled component. Honestly, there's no other good way todo this because rendering each component to extract text is just way too expensive.  It also fits along nicely with all the callback functions I'm adding so people can have more control.

The only thing I'm hung up on is the names right now for `renderValue` and `renderComponent`. I'm going to approve this PR but I want to think on that a little bit longer tonight before I do. After we're set on the names providing good information in the README will be key

Otherwise this is great. Thank you for this submission. The example will be a nice addition

gregnb:
@tgBryanBailes how about instead of `renderComponent` and `renderValue` it just becomes `display` with a little twist to handle all cases:

```
  // this provides a hook into updating the data source to an outside component
  // 
  display: (index, value, setValue) => {
    return (
      <SomeCustomComponent updateData={setValue}>
        {value}
      </SomeCustomComponent>
    );
  },

  // also, if a person choses to return just a string then they can do this
  display: (_, value) => value.toLocaleString('en-US', {style: 'currency', currency: 'USD'})

```

tgBryanBailes:
What exactly is `setValue` in this scenario?

gregnb:
`setValue` would be a function that would update the data store inside MUIDataTable. It would look something like this:

```
  updateDataCol(row, index, value) {
    // update the data store inside the library
  }

  getDisplayData(columns, data, filterList, searchText) {
    let newRows = [];

    for (let index = 0; index < data.length; index++) {
      if (this.isRowDisplayed(columns, data[index], filterList, searchText))
        newRows.push(columns.map((column, colIndex) =>
          typeof column.display === "function" 
            ? column.display(index, data[index][colindex], this.updateDataCol.bind(null, index, colindex)
            : data[index][colIndex]
        ));
    }

    return newRows;
  }
```

tgBryanBailes:
The only issue I see is if the component manipulates the value (like the Active component does in the example), how would we sort or search for that value? Maybe `setValue` could handle it, I am just not sure how that would happen without digging into it (which I won't be able to so for a couple of days). I just think that will end up being the "gotcha" with this setup.

gregnb:
Ok, I'm going to take a look at this in the morning it's getting late for me. I will take your example and see if there are any issues trying to implement. If it turns out to be fine I will add some documentation to the readme on how to use this all and merge in the PR

If not, we'll revisit

tgBryanBailes:
Commit what you have. Perhaps I can help get it going sooner.

gregnb:
pushed it up @tgBryanBailes

gregnb:
So just to circle back I converted everything into one method called 'display' and with that you get a function in the 3rd argument called 'setValue' -- now, if a user passes this function along to the component it can run that function as well so that the tableData updates.

Biggest issue though is what about the initial data value? I don't want to call setValue() on each componentDidMount for the individual table cells.  I can't think of a smart way todo this yet and maybe there may not be.

tgBryanBailes:
I hope this better outlines the "gotchas" I was speaking about yesterday.

Definitions:
- _Initial Value_ is the value provided in the data array
- _Component Value_ is a custom component provided for display
- _Display Value_ is the manipulated value of the Initial Value (provided by the `display` prop, either the result of a `display` function, or the manipulated value provided if a component is displayed)

Sorting:
- Initial Value is an integer, Display Value is an integer formatted by locale. Should sort by the Initial Value.
- Initial Value is a boolean, Display Value is "Yes" or "No". Should sort by the Display Value.
- Initial Value is a JS date object, Display Value is the date formatted as mm/dd/yyyy. Should sort by the Initial Value.
- Initial Value is a string, Display Value is a different string. Should sort by the Display Value.

Filtering/Searching/CSV Export:
- Should always utilize the Display Value

Components:
- How do we provide a Display Value and a Component Value to a single `display` prop in a way that makes sense?

gregnb:
Ok, thanks for going through and explaining everything. I will time box trying to work on this to next Wednesday. If nothing pans out, I will reset the head on your branch to your last commit and add in documentation for renderValue and renderComponent

gregnb:
Got it. I made some small changes and pushed up. Still need to work on it but I'm seeing sorting work for:

- currency format
- yes/no booleans
- cities component. if I select a new value it then reflects and if i sort again it sorts by that new value

The demo is updated if you wanna give it a look. I'll try to circle back tomorrow for the other items. Going to really see if this can be done 

screen shot 2018-04-05 at 9 54 45 pm

tgBryanBailes:
It seems to work rather well. Since your concern was an expensive re-render on update, I should point out that your `updateDataCol` method currently triggers a full re-render. The only way I can think of to prevent that would be to set the value in the `MUIDataTableBodyCell` component, then update the state there instead of updating the entire `data `array.

tgBryanBailes:
I noticed that the tooltip for Print doesn't work anymore. After digging over what could have possibly caused this regression, I realized that even your live demo has this issue. So, not sure when this was introduced, but it wasn't part of this PR.

gregnb:
Ok just catching up on this thread. I’ll have a couple hours tonight to put in to try to finish this off. I think it’s close.

Regarding the ToolTip if you want to take a look I would love a hand in a separate PR. If not, no biggie I’ll look into it

tgBryanBailes:
The code looks correct to me. I think it has to do with your react-to-print component. But, I'll dig more and see what I can find.

gregnb:
Made a couple quick changes:
- added lodash deepClone to properly clone some nested arrays (columns, data, filterData).  
- filterData now reflects changes from a custom rendered component (ie: change a city, that value shows in filter list)

I think this is really close to merging. Thoughts?

tgBryanBailes:
I noticed earlier that updates didn't work correctly after sorting. But, I will test out your latest commit on Monday and let you know what I find.

gregnb:
Sounds good! Enjoy the weekend

tgBryanBailes:
You too!

tgBryanBailes:
Yep, a pretty big issue with the Active Switch. Sort by Active, then try changing Justin Mann to "Yes".

tgBryanBailes:
The last remaining issue I see is that the Active filter shows True/False for the values instead of Yes/No. I will spend some time today trying to find a solution.

gregnb:
Ok I’ll have time tomorrow to get back to this just a lot going on right now. I’m thrilled about your other PR as well can’t wait to test it out. 

Hopefully square this one away tomorrow night and merge it in

gregnb:
Ok, checking out the latest commits and first couple quick thoughts:

>Made it so that if your component provides a "value" prop, then it will use that value for the Filtering/Searching instead of the one provided. This provides for the use case where the component manipulates the data and shows something different AND fixes the issue where filtering/searching weren't working if customRender returns a string.

1. We are asking the user to know that the only way to change this is by providing a prop on their own component that we’ll extract out 
2. What if the component provided to be rendered looks like:
		
		<ParentWrapper>
		  <AnotherWrapper>
			<RealInput value={“Finally value is here”}>
			</RealInput>
		  </AnotherWrapper>
		</Parent>

How would we extract it?

Also, if you load up the example and then try to toggle the last active switch on the last row the page dies because this line 280 in MUIDataTable.js:

```
        filterData[row].sort(collator.compare);
```

filterData[row] is undefined because it's checking for false but we have applied the display data of either Yes/no so it never finds it.  


---

So overall, I'm wondering is there a simple way to provide the customRender a way to have a value for sorting/filtering and a value for displaying.  What if

```
customRender: (index, value, updateValue) => {
}
```

became

```
customRender: (index, value, updateData) => (
            <FormControlLabel
              label={value ? "Yes" : "No"}
              value={value ? "Yes" : "No"}
              control={
                <Switch color="primary" checked={value} value={value ? "Yes" : "No"} />
              }
              onChange={event => {
                updateData(dataValue, optionalDisplayValue);
              }}
            />
          )
)

```

That way, if someone needed further control to change the display value they can provide both right there? and we don't need to search for a value prop because we can't always depend on it being the root component provided.

gregnb:
Sorry, and just to be clear on that use case of the value prop:

A developer has a column that is a boolean and would also like to render all the cells as a component like a Switch using customRender.  Now, we would like to display "Yes" or "No" but keep the boolean value for the sortable data.  This allows us to use the search and type "No" or "Yes" and get filtered results. Along with the filtered drop down having the ability to select "No" or "Yes".

tgBryanBailes:
For the `updateData` solution to work, you would have to call it immediately as well. Otherwise, the initial value won't have the display value.

gregnb:
So your original commit came in with this roughly:

```
          renderValue: (value) => value ? "Yes" : "No"
          renderComponent: (index, value) => (
             <FormControlLabel
                label={value ? "Yes" : "No"}
                control={
                  <Switch color="primary" checked={value} value={value ? "Yes" : "No"} />
                }
                onChange={event => true}
              />
          ),
```

Where I'm stuck is how can I provide that initial value to kickstart the boolean in the example above? If we didn't want to rely on extracting the 'value' prop how could this be done?

I like having one function (customRender) to modify what is being displayed and then callback functions provided in the arguments to trigger updates.  It could still be useful incase someone did not want to manage their own updates to the data store and wanted the library to take care of it via updateValue.  But, having to add this initialValue property is bringing a step closer to what you have above 

```   
          initialValue: (value) ? "Yes" : "No",  // stuck here. Now i'm having to add a property 
          customRender: (index, value, updateValue) => (
            <FormControlLabel
              label={value ? "Yes" : "No"}
              value={value ? "Yes" : "No"}
              control={
                <Switch color="primary" checked={value} value={value ? "Yes" : "No"} />
              }
              onChange={event => {
                updateValue(event.target.value === "Yes" ? false : true);
              }}
            />
          )
```

Still thinking 🤔

gregnb:
> The thinking behind the value prop was that it works with all of Material UIs input controls and, worst-case scenario, it could accommodate any use case by creating a simple wrapper component that provides a value prop and passes it down accordingly. It is a bit strange, but works across the board.

Honestly, I'm thinking you're right. After looking more closely at this value is pretty standard.  I'm going to go with this solution.

gregnb:
Made a couple minor tweaks & pushed up. I'm testing this out and it looks good. If you can give it a go on your end let me know if you see any snags.

If we're good I would like to start looking to merge this in

tgBryanBailes:
In testing, I found that it is no longer responsive. Looking into why that would happen.

tgBryanBailes:
Need to get the coverage up. Other than that, it looks great.

gregnb:
Yeah, it's been on my radar.  So we're in agreement then what is here is good I'll run a coverage report when I get some time this week and try to patch in what is missing to bring it back to 90%

gregnb:
@tgBryanBailes I'm going to merge this PR into master. Thanks for all your hard work! It'll be in this week's release

PR #31, opened 2018-04-09
Filter multiselect

tgBryanBailes:
Added option to use `multiselect` as a `filterType`


gregnb:
This looks great. Merging in!

PR #34, opened 2018-04-16
The responsive: scroll option looks terrible

tgBryanBailes:
Made it look good and made headers and footers fixed.

### DIFF
--- a/examples/component/index.js
+++ b/examples/component/index.js
@@ -108,7 +108,7 @@ class Example extends React.Component {
     const options = {
       filter: true,
       filterType: 'dropdown',
-      responsive: 'stacked',
+      responsive: 'scroll',
     };
 
     return (
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <title>MUIDatables Example</title>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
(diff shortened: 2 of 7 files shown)


gregnb:
Right, good find. This came about because I have to create this additional "hidden" td that then shows itself on a stacked view.  Ok, fine, but when I run this branch I see another issue

fabe43f8-ba6c-42dc-y36c-8ae18639db0b

We now have a scrollbar that appears

tgBryanBailes:
Hmmm, that doesn't happen for me in Chrome. Which browser are you using?

gregnb:
Browser: Chrome Version 65.0.3325.181 (Official Build) (64-bit), 
OS: MAC 10.13.2

I get the same result in Safari

screen shot 2018-04-17 at 6 54 07 pm

tgBryanBailes:
OK, was able to reproduce. It seems to be caused by the Tooltip. If you scroll to the last column (inner scrollbar) then hover over the column header, the scrollbar goes away.

gregnb:
What are your thoughts on a solution? Potentially remove tool tips is one I have in mind if we can’t fix the issue.  I know there’s still an outstanding issue on the material-ui board related to tooltip but I can’t remember what it was exactly

tgBryanBailes:
Just committed a fix using some simple CSS on the example.

gregnb:
can we work in this popper.js fix into the library? Rather not have it outside this is a very targeted fix that nobody will dig up to add to their own pages

tgBryanBailes:
I'm not sure how to do an attribute selector in JSS. I'm not sure there is a way...

tgBryanBailes:
OK, figured out how to do an attribute selector in JSS. I didn't even know that it could be done.

PR #41, opened 2018-05-04
Fix csv export for unicode data

tux-tn:
Calling CSV Download button on Unicode data will cause a InvalidCharacterError exception because window.btoa supports characters inside the range 0x00 to 0xFF. 
This PR fixed the problem by encoding extended characters to their ASCII representation


gregnb:
Awesome thanks for fixing this issue!

Issue #45, opened 2018-05-07
“View Columns” and "Filter Table" icons not show with HMR (react-hot-loader)


shubham81407:
## Config

```
/**
 * Data Table
 */
/* eslint-disable */
import React from 'react';
import MUIDataTable from "mui-datatables";

// page title bar
import PageTitleBar from '../../../components/PageTitleBar/PageTitleBar';

// rct card box
import RctCollapsibleCard from '../../../components/RctCollapsibleCard/RctCollapsibleCard';

// intl messages
import IntlMessages from '../../../util/IntlMessages';

class DataTable extends React.Component {
  render() {
    const columns = ["Name", "Title", "Location", "Age", "Salary"];
    const data = [
      ["Gabby George", "Business Analyst", "Minneapolis", 30, "$100,000"],
      ["Aiden Lloyd", "Business Consultant", "Dallas", 55, "$200,000"],
      ["Jaden Collins", "Attorney", "Santa Ana", 27, "$500,000"],
      ["Franky Rees", "Business Analyst", "St. Petersburg", 22, "$50,000"],
      ["Aaren Rose", "Business Consultant", "Toledo", 28, "$75,000"],
      ["Blake Duncan", "Business Management Analyst", "San Diego", 65, "$94,000"],
      ["Frankie Parry", "Agency Legal Counsel", "Jacksonville", 71, "$210,000"],
      ["Lane Wilson", "Commercial Specialist", "Omaha", 19, "$65,000"],
      ["Robin Duncan", "Business Analyst", "Los Angeles", 20, "$77,000"],
      ["Mel Brooks", "Business Consultant", "Oklahoma City", 37, "$135,000"],
      ["Harper White", "Attorney", "Pittsburgh", 52, "$420,000"],
      ["Kris Humphrey", "Agency Legal Counsel", "Laredo", 30, "$150,000"],
      ["Frankie Long", "Industrial Analyst", "Austin", 31, "$170,000"],
      ["Brynn Robbins", "Business Analyst", "Norfolk", 22, "$90,000"],
      ["Justice Mann", "Business Consultant", "Chicago", 24, "$133,000"],
      ["Addison Navarro", "Business Management Analyst", "New York", 50, "$295,000"],
      ["Jesse Welch", "Agency Legal Counsel", "Seattle", 28, "$200,000"],
      ["Eli Mejia", "Commercial Specialist", "Long Beach", 65, "$400,000"],
[...]

PR #46, opened 2018-05-07
Feature/update mui icons

baer:
As of Material-UI beta-41, the material-ui-icons package was moved under the namespace `@material-ui`. See the [beta-41](https://github.com/mui-org/material-ui/releases/tag/v1.0.0-beta.41
) release notes for details. 

This PR updates the `material-ui-icons` package to the new `@material-ui/icons` and updates all code references to the name package name.

Note: This was a breaking change for `material-ui` and will therefore be a breaking change for `mui-datatables`. Not sure of the best way to handle this.


gregnb:
Thanks for the PR @baer this should close #45

gregnb:
@baer To your point, it's putting this package in a weird spot here. A breaking change based on a dependency should the semver go up a whole major? seems excessive. What are your thoughts?

baer:
Humm, well for this project I see 3 problems:

*The Semver numbers depend on non semver projects*
`mui-datatables` relies on material-ui beta which may or may not have breaking changes between versions. Depending on the 1.x line was a calculated risk and I think was the right choice but it puts you in awkward situations when there are breaking changes like this one.

*_Real_ Semver vs. _Semver_ in the wild*
Semver is meant only as a system for alerting users to major/minor/patch changes but a lot of projects use it to mean both that _and_ to indicate changes in philosophy or major rebuilds/ refactors. To me, if you're going to say that a project uses semver, then use semver and don't be afraid of major revs. I feel strongly about this in projects I'm involved with but I respect that a lot of people do it differently.

*Semver for UI*
So that's by-the-book semver but `mui-datatables` has another issue and that's that semver doesn't really apply to UI. In my experience, I think that semver for apps just doesn't work but for libs like this you can treat the UI as a black box and do semver based on the integration points. If you start assuming that people are going to rely on specific pieces of DOM then you'll end up with a major on almost every release.

All this is to say that I think that major is just fine in this case :). Also, thanks for all the great work!

gregnb:
@baer Thanks for the valuable feedback! Really great stuff and gave me some things to think about. 

I'm thinking releasing this a major would be fine if Material-ui already pulled the trigger on V1.  One thought I have is what if they squeeze in another big breaking change before the V1 version comes out on May 17th. Then I'm in that same position again.  I'm thinking of maybe just doing a patch here to help some people who are stuck with this issue.  Then on the 17th when V1 finally comes out doing a major release to sync with that.

baer:
If you do a patch, this will break anybody who has `^` in their package.json the next time they run `npm install` on a fresh repo OR if they run `npm update` on a current one.

baer:
You could do a 2.x.x-beta-1 as a pre-release. That wouldn't break existing users but would tide folks over who need this until next week. Also, I didn't know it was going live next week. That's super exciting.

gregnb:
@baer Ok, thanks for pointing this out. With that said I'm going to go ahead with releasing another major and keep my fingers crossed Material-ui doesn't change too much before the 17th.  Appreciate your help!

baer:
For sure. Either approach is fine and TBH now that I know it's only a week away I'd be fine to just wait it out. FWIW, I've been using the 1.x line since beta 20something and this is the only major I've seen.I'm crossing my fingers for ya.

gregnb:
Yeah it's been pretty smooth sailing.  At work we're on about the same version

```
    "material-ui": "^1.0.0-beta.26",
    "material-ui-icons": "^1.0.0-beta.5",
```

and just waiting until next week to get it over with :)

Issue #27, opened 2018-04-04
More details on how onRowsSelected relates to data in grid


JordanKadish:
## Issue

I understand that the  onRowsSelected callback returns the index of which checkbox was last ticked/unticked, as well as the indices of all currently selected items. When I tick a checkbox and then sort the rows in a different way, the index of the row changes.
Example:
Row 1: bob, smith
Row 2: john, apple

If I click the box for row 1, it will console log: [0], [0]
If I then say sort by surname, ascending, and untick the ticked row 1 box, it will console log: [1],[1]

With this in mind, how am I supposed to get information out of a selected row (eg make a json obj out of the row when tickbox is ticked) and how can I use the array of all currently selected items when I have stuff like server side pagination only sending the current page to screen (IE storing which items have been previously selected on the server). 

Thoughts: Assuming there is a way to get a json obj of the row currently being ticked/unticked, it would be easy to just store a selectedItems array server side, I am just not sure how to pull a row to then send to server
## Your Environment

| Tech         | Version |
|--------------|---------|
| mui-datatables  |    lateast     |
| node        | latest         |
| OS      | arch linux         |

JordanKadish:
More clarity (since I feel some is needed):

I bound my callback function this {this} in constructor
Then in the method, I tried to use the array index args being sent relate to the state data. This doesn't work, so I'm just not sure how to use the data being displayed on the screen to get a row's data that way. I have tried relating the indices to this.state.data, but when I sort the on screen grid in any way, the on screen data won't match the state data indices...

gregnb:
Hey @JordanKadish, you're like the biggest package supporter! Not sure where I would be without all your feedback :). 

Ok, I'm seeing your issue now. While it's nice to get the index of what is currently selected what is really required is the index of where it is in relation to the provided datastore.  What if we modified onRowsSelected to return the following (imagine if we checked off both rows):

[{ index: 0, dataIndex: 0 }],
[{ index: 0, dataIndex: 0 }, { index: 1, dataIndex: 1 }],

gregnb:
Also, if that is something that will help would you want to take this work up?

gregnb:
Ok, the way i'm seeing it if data looks like this:

```
const data = [
["bob", "smith"],
["john", "apple"]
];
```

Steps:
1) Click on row #1 
   -> onRowsSelect returns [{ index: 0, dataIndex: 0} ], [ {index: 0, dataIndex: 1}]
2) Click on row #2
[...]

PR #59, opened 2018-05-18
new feature: render custom selected rows toolbar + dataIndex on selectedRows callback

### LINKED ISSUE #47: Question add/remove or modify Icons on selected rows
Good morning, 

Is there a way to add new Icons when selecting a row. Currently there is a delete Icon. I'd like to add an edit or more info icon somewhere and pop a Modal to extract row information. It seems only way to extract row info is by selecting the row and using an `onRowsSelect` function. Is there another way to extract the row information and display it in a popup Modal? if so I can nest a custom button using `customRender ` function.

Thanks for all your work.
Ray

gregnb:
When merged this should close #47 #27 

**TODO:** 
- Add prop for selectedRows to support controlled components
- Fix unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ The component accepts the following props:
 |:--:|:-----|:--|:-----|
 |**`page`**|number||User provided starting page for pagination
 |**`filterList`**|array||User provided filter list
+|**`rowsSelected`**|array||User provided selected rows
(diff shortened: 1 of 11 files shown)


RaySheikh:
There is an extra s on **rows(s) selected**
image

pamminnie:
Hi @gregnb is it possible to include row data in? As I wanted to select multiple rows to perform certain actions. Currently, there is only index and dataIndex which does not help to identify the row data. It would be great if I can include our own ID / Data.

PR #61, opened 2018-05-20
Update ReadMe to note that strings in data must be non-null or will cause a crash (issue 54)

teadams:
The one change is 
|**`data`**|array|Data used to describe table. Must be an array of non-null simple strings 

The rest are white space.  I'm using Atom.  Is there something I should set up on the editor to not cause the white space additions?


teadams:
I figured it out.  Will submit a new pull request

Issue #54, opened 2018-05-16
Small suggestion for documentation - data can not contain nulls


teadams:
I hooked up the table to the result of a database which contained nulls.  It caused on error which I easily fixed. Small suggested to note in the document that the data cannot contain nulls.

This UI is absolutely incredible. thank you for making it available.

teadams:
also recommend you give a link to the examples of custumRender from the main documentation. Very useful!
https://github.com/gregnb/mui-datatables/blob/master/examples/component/index.js

gregnb:
One of my short term goals is to finish up the docs setup with links to tons of examples. Trying to balance finishing up the features needed with documentation. It can be tough.

Would you like to pitch in and help me finish the docs site?

teadams:
Sure

teadams:
will do this tomorrow

gregnb:
Ok great @teadams , but actually do not list anything about this null issue. It's actually a bug I should address in which null values should just be treated as empty values to be rendered for display.

What I do need help on is expanding the docs portion of the website and filling more useful examples out and then having a drawer menu appear when clicking on the hamburger menu for the docs website.

teadams:
Hi Greg,  I'm working on a project "CRUD-mui-datatables".  This uses your component as a base and adds the "create/update/delete/retrieve" all as one component.  In addition to publishing on Git, I'm going to have a sample with a working backend connected to AWS so people can install and instantly see it working with a real server.

This should generate some really good examples for how to use MUI-datatables in a real world context.

gregnb:
@teadams Cool sounds good. I am open to reviewing this when you're finished but again what I am really asking for is finishing up the docs site located here: http://www.material-ui-datatables.com

There is a docs/ folder in this project and I would love to have the hamburger menu pop out a drawer much in the way http://www.material-ui.com does and show have links to running examples from the current examples/ folder

teadams:
Sure.  

Is this project associated with Material UI? (I see you use the log)

gregnb:
Closing because mui-datatables@2.0.0-beta-20 now accepts nulls

PR #62, opened 2018-05-20
Doc changes for #54 (non-nulls in data and customRender examples)

teadams:


gregnb:
@teadams so as I mentioned in the issue thread I am going to fix this bug to allow nulls. I will keep the other Readme update to list to the example. Thank you for the PR!

PR #80, opened 2018-06-07
Add missing viewColumns option on README.md

nihey:
Just noticed it was on the code but not on `README.md`, thought I could give my 2 cents on that.


gregnb:
Great catch, man! Thanks for the PR!

Issue #75, opened 2018-06-05
MUIDataTable cannot customize  count


rst66:
There are no prop or option to customize the total number of rows, which is  the ``count`` prop
of ``TablePagination`` . 

This attribute is the ``length`` of the data, but in my case, only one page of data is pulled from backend service, with  the ``total`` attribute. I think this is the common case, so it will be very necessary to be able to customize the ``count``.

## Expected Behavior
add ``count`` in ``options`` prop of MUIDataTable

## Current Behavior
no such option

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    1.0.0-beta.41     |
| MUI-datatables  |  1.1.4       |
| React        |  16.3.1       |
| browser      |     chrome    |

adetwiler:
I am having the same issue, I think the real enhancement here would be to support remote datasources in general. Such as sorting, filtering, searching, pagination, etc.

Currently this only seems to support an entire datasource rather than a paginated datasource either from a remote or local source. It would be nice to get further support on that rather than it just being limited to needing the entire dataset.

jstgermain:
Having the same exact issue, and I have come to the same conclusion as @adetwiler. This has halted development and forced me to download the git repo instead of the NPM module. It seems like I have it really close to working with remote data, but issues in filters, searching, and pagination. This module is amazing, but it seems odd that it wasn't built to handle remote data with an offset and limit instead of a full data set. I look forward to seeing further releases and hope to not have to fork my own version. This has really spread up development till the remote data issue came into play.

gregnb:
I will add count property to the options prop. Hope this will address all your issues

floscom:
yeah would be great!

Issue #73, opened 2018-06-01
rows click event


JoonHyunJeong:
i'm trying to handle event when row clicked but it seems there is no such options in mui-datatables.
How can i implement it?

gregnb:
I will address these shortly as a part of a larger improvement

gregnb:
I am going to provide a callback named `onCellClick` which will look like so:

```
function(tableMeta) => void 

tableMeta: {
  rowIndex: number,
  columnIndex: number,
  columnData: array, // Columns Options object
  rowData: array, // Full row data
  tableData: array, Full table data
  tableState: {
    announceText: null|string,
    page: number,
    rowsPerPage: number,
    filterList: array,
    selectedRows: {
      data: array,
      lookup: object,
    },
    showResponsive: boolean,
    searchText: null|string,
  },
}

```

That will provide you every bit of information

gregnb:
This is now available in the latest release. A callback function named `onCellClick`

laxmikamala:
how to get particular row data onRowClick in mui-datatables

PR #85, opened 2018-06-26
add count support for options + onCellClick callback

gregnb:
This should solve #75 and also #73 when merged 

@rst66 @adetwiler @jstgermain

### DIFF
--- a/src/MUIDataTableBody.js
+++ b/src/MUIDataTableBody.js
@@ -124,6 +124,7 @@ class MUIDataTableBody extends React.Component {
               options={options}>
               colIndex={0}
               rowIndex={0}
+            >
               <Typography variant="subheading" className={classes.emptyTitle}>
                 {options.textLabels.body.noMatch}
               </Typography>
--- a/src/MUIDataTableBody.js
+++ b/src/MUIDataTableBody.js
@@ -18,6 +18,8 @@ class MUIDataTableBody extends React.Component {
   static propTypes = {
     /** Data used to describe table */
     data: PropTypes.array.isRequired,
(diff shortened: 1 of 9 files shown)


src/MUIDataTableBody.js
```diff
@@ -115,6 +122,8 @@ class MUIDataTableBody extends React.Component {
             <MUIDataTableBodyCell
               colSpan={options.selectableRows ? columns.length + 1 : columns.length}
               options={options}>
```
jstgermain: this closing HTML bracket should be after the `rowIndex`

gregnb: Nice catch. Thank you!

gregnb:
@jstgermain I haven't given that much thought but if you want to spearhead that initiative and provide and example I would love to include it into examples/ 

If you're stuck maybe at least create a test repo with some mock data and we can work from there? See if we can come up with a viable solution.

I will be merging this PR later tonight and doing a release then

jstgermain:
@gregnb Awesome! I am a dev for a small project, and @adetwiler is actually the CTO for the project. We would love to work on this with you. If you could contact me via text message on my business line, (480) 757-9953, maybe we can setup a hangout to show you what we are dealing with and come up with a solution together. We are interested in assisting in the development of this project. It is amazing and would love to think of other edge cases to crush.

gregnb:
@jstgermain Sure man, l'll touch base with you sometime tomorrow as I have a pretty full day with meetings. I'll try to assist as much as I can

jstgermain:
@gregnb awesome. We look forward to it.

PR #88, opened 2018-06-26
Feature/options elevation

benjaminbours:
Hello, just added an option "elevation" relying to the [Paper React Component](https://material-ui.com/api/paper/).


gregnb:
Hey @benjaminbours, thanks for the PR but this is going to need some more work. See if you can find a nice way to affect a change to the Paper component by leveraging CSS overrides instead I would accept the PR.

What I mean by that is maybe apply a className to the Paper component in MUIDataTable.js. Then, if a developer wants to override it they can do so that way.

jstgermain:
@benjaminbours with your change, are you going to make it so you can set the elevation all the way to 0?

benjaminbours:
@gregnb  Sure i understand !

https://material-ui.com/api/paper/ Here, we can see there is a props classes where I can override or extend styles applied to the component.

Does it sounds good for you if I replace the "elevation" options by a "paperClass" options where developer can write the style wanted in an object ? Like you do in JSS.

Or do you mean a real className like a string, and developer apply style to that class in CSS ?

@jstgermain You are right, i forget to limit value, Paper component accept an elevation 0, but with what i wrote, you can set it below 0 and create a bug i think, so thank you to notice it :)

jstgermain:
@benjaminbours awesome man. I ended up just adding custom style originally to remove elevation. This will be a nice addition since it was something I was looking to move towards anyways. Great job. Glad to help for the big too.

gregnb:
@benjaminbours I was thinking more along the lines of something like this:

```
<Paper elevation={4} ref={el => (this.tableContent = el)} className={classes.paper}>
```

then if someone wants to override it would look like this:

```
import React from "react";
import MUIDataTable from "mui-datatables";
import { createMuiTheme, MuiThemeProvider } from 'material-ui/styles';

class Example extends React.Component {

  getMuiTheme = () => createMuiTheme({
    overrides: {
      MUIDataTable: {
        paper: {
          // let a user do whatever they need todo to override 
        }
      }
    }
  })

  render() {
  
    return (
      <MuiThemeProvider theme={this.getMuiTheme()}>
        <MUIDataTable title={"ACME Employee list"} data={data} columns={columns} options={options} />
      </MuiThemeProvider>
    );

  }
}

```

Since the package is now using MUITheme's all styling modifications should be done through it

jstgermain:
I did my styling directly on MUIThemes itself with overrides, but slightly different. @gregnb solution seems best IMO.

benjaminbours:
Yep it seems it's the proper way :) I did the change, and checked in exemples/customize-styling, and it works.

PR #101, opened 2018-07-20
[MUIDataTable] fix className prop

kulakowka:
Add className prop in MUIDataTable


gregnb:
Hey @kulakowka thanks for the PR but the way to solve this issue was to adjust from className to classes={{ tooltip: classHere }} 

I've made the change and submitted the PR since other people were complaining

Issue #103, opened 2018-07-23
Update to @material-ui/icons v2.0.0 


jkeruzec:
Can you provide support for updated material-ui/icons V2.0.0 ?

`Module not found: Can't resolve '@material-ui/icons/FileDownload' `

gregnb:
Resolved thanks to @terenced I will be doing a release later tonight

kalyan542:
The same issue is coming in "@material-ui/icons": "^3.0.2"

LeoLozes:
I don't really understand the release process of this project, because the releases and tags listed in github are not in line with the releases in npm. Anyway. I lost one hour trying to figure out this one. 
It seems that the last version released in npm is the `2.0.0-beta-55`, except it's not the source of the master branch, hence the FileDowload import that hasn't been there for months. So I recommend using the version `2.0.0-beta-54` which looks like the source matches the version.

Thanks for that.

gregnb:
@LeoLozes First off, watch your tone. This is a free project if you don't like it you can totally leave Bud. 

As for this issue with FileDownload there is no import of it in either -54 or -55 release so what are you even talking about?

LeoLozes:
You are right about my tone, I'm sorry, I spent way too much time on this one this morning. 

BTW something must have been wrong with the version 55 because it has been pulled out npm when I was trying to document this issue properly:

```
npm ERR! No compatible version found: mui-datatables@2.0.0-beta-55
npm ERR! Valid install targets:
[...]

PR #104, opened 2018-07-23
Upgrade Material-UI Icons to v2.0.0

terenced:
Resolves issue #103

### DIFF
--- a/docs/pages/index.js
+++ b/docs/pages/index.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import Typography from "material-ui/Typography";
 import IconButton from "material-ui/IconButton";
 import Tooltip from "material-ui/Tooltip";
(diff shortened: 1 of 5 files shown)


gregnb:
@terenced Thanks for the PR. I would be happy to accept if you can revert the download icon to FileDownload

terenced:
@gregnb In v2.0.0 of the icons, `FileDownload` has been replaced by `CloudDownload`, that is why I switched.

https://github.com/mui-org/material-ui/blob/master/packages/material-ui-icons/src/CloudDownload.js
https://github.com/mui-org/material-ui/tree/master/packages/material-ui-icons

gregnb:
Huh, wow no kidding looks like they did remove it. Oh well I did like that icon. Merging in now.
Thanks for the info!

Issue #114, opened 2018-07-31
support for external pagination handling


jmarucha:
When the data is provided from external source (REST API, database, etc) the internal handling of pagination is unnecessary. The option `externalPagination`, default to false can be provided to turn on and off the array splicing for rendering the table, and paging is handled via events onChangePage and onChangeRowsPerPage.

PR #115, opened 2018-07-31
[#114] added externalPagination option

jmarucha:
Added new option that disables built-in pagination for the cases where paginated
data is provided from external source.


gregnb:
hey @jmarucha thanks for the PR but I just recently added a property called 'customFooter' which will allow you to completely override that section. Unfortunately I will have to decline the PR because the functionality is now there in the latest version

PR #123, opened 2018-08-05
Enable sorting by column only when sort is enabled for column

### LINKED ISSUE #121: Disabling sort for a single column doesn't seem to work
The documentation says you can enable/disable sort for the whole table, or for each column individually. Currently only enabling/disabling sort for the whole table seems to have any effect.

## Expected Behavior

According to the documentation, you should be able to enable or disable sorting for a particular column with the options, e.g. if the columns are:
`const columns = [
 {
  name: "Name",
  options: {
   filter: true,
   sort: false
  }
]
`
then sorting should be disabled on the name column (even if sort is true in the options passed to the table).

If sort is true in the table options (default), then sort should be enabled/disabled depending on the options for each column.
If sort is false in the table options, then sort should be disabled for all columns.
## Current Behavior

If sort is true in the table options, then sort is enabled for all columns.
If sort is false in the table options, then sort is disabled for all columns.

See here for the standard example modified to attempt to make the name column have sort disabled.
https://codesandbox.io/s/oo5kjmvz2q
However, you can still sort by name (in the non-stacked view) by clicking on the Name header.

## Fix?
I think I can fix this by changing the dataTableHeadCell to consider both props.options.sort and props.sort, I can submit a pull request possibly I think?

kirstenland:
Should fix #121

### DIFF
--- a/src/MUIDataTableHeadCell.js
+++ b/src/MUIDataTableHeadCell.js
@@ -45,6 +45,8 @@ class MUIDataTableHeadCell extends React.Component {
     sortDirection: PropTypes.string,
     /** Callback to trigger column sort */
     toggleSort: PropTypes.func.isRequired,
+    /** Sort enabled / disabled for this column **/
+    sort: PropTypes.bool.isRequired,
   };
 
   handleSortClick = () => {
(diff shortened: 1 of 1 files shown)


gregnb:
Thanks for the PR!

PR #130, opened 2018-08-13
Add Feature: onRowClick

Skn0tt:
mui-datatables is a great project! It helps me a lot at [EntE](https://gitlab.com/skn0tt/EntE).

There was one feature missing for me:

`onRowClick` gets called when a row is clicked.
It's provided the full `data` of that row.


gregnb:
@Skn0tt Thanks for the feedback, man! Looks almost good to accept but can you make the following change:

I would like the callback parameters to look like this:

```
onRowClick(rowNumber, rowData)
```

Codedrainer:
how can i use onRowsSelect ?

gregnb:
@Skn0tt Can you make those changes?

Skn0tt:
I'll return from my vacation in a week, then I'll get around to do this. Cheers :)

Skn0tt:
Should be working now. I changed the signature to `(dataIndex: number, data: string[], rowIndex: number) => void`.

gregnb:
Looks almost right can we match what I had above?
```
onRowClick(rowNumber, rowData)
```

make it so that the 2nd parameter is just an object that contains 
```
{ dataIndex: number, rowIndex: number} 
```

It will match what I have in another place

Skn0tt:
What exactly should difference between `rowNumber` and `rowIndex` then be?
Wouldn't it be better to do:

```
onRowClick(rowNumber, { dataIndex: number, data: string[] })
```

gregnb:
Ok, a mistake on my end. Can we change it to match the following? Sorry for the inconvenience.
I'm looking to probably update some of these other callbacks into the future to provide just an object that has the cell information (dataIndex and cellIndex/rowIndex)

```
onRowClick(rowData, rowMeta)
```

make it so that the 2nd parameter is just an object that contains 

```
{ dataIndex: number, rowIndex: number} 
```

What do you think?

Skn0tt:
That seems better.
I'll implement it!

gregnb:
Awesome, thanks man! Again, my bad!

titulus:
Good job! This feature was last who blocks me from using this repo. :+1:

Issue #156, opened 2018-08-23
Display empty table if no data is given


jmarucha:
Behaviour on displaying empty dataset is weird.

## Expected Behavior

When I provide `data=[]` prop, a table with "Sorry, no matching records found" message is provided, similarly to manner when search or filtering shows nothing.

## Current Behavior

When I provide `data=[]` prop, no table is rendered.

| Tech         | Version |
|--------------|---------|
| Material-UI  | 1.4.1 |
| MUI-datatables  | 2.0.0-beta-19 |
| React        | 16.4.1 |

jmarucha:
edit: it's the same on MUI-datatables 2.0.0-beta-

gregnb:
I can't replicate this issue. See sandbox below:

https://codesandbox.io/s/n0xk6pz9o0

gregnb:
This is intended behaviour. If a user provides empty data nothing will render.  I'm leaving this up to the user to decide how they want to handle situation where data is not available yet (say a loading state)

brianpmaher:
>  I'm leaving this up to the user to decide how they want to handle situation where data is not available yet

Well, the user could also decide not to display the table.  Allowing the empty table to display gives the user more options for how to display when the data set is empty.

PR #159, opened 2018-08-23
fix issue #156

benjaminbours:
Fix the weird behaviour when displaying empty dataset.


gregnb:
This is intended behaviour because I'm leaving this up to the user to decide how they want to render situation with no data provided

Issue #145, opened 2018-08-20
Trigger the close search button


gadanihiman:
I just want to ask simple issues here, because i can't trigger the close button for search, you just give callback onSearchChange for trigger when the textSearch changed.

I just wonder to have a trigger callback when i close the search placeholder here.

Or maybe i want to hide the close button for that search placeholder

Thanks so much!

gregnb:
So you're asking for a callback function when a user clicks the 'x' on the search or hits the escape key? Sure sounds reasonable.

I would probably name it something like 

```
 onSearchClose
```

gadanihiman:
That's sounds good, i need that. Thanks so much!

PR #161, opened 2018-08-24
#145 Add callback function when search input is closed

benjaminbours:
Right now, no params are returned. Don't know if it's relevant to return some.


gregnb:
Looks good. Thanks for the PR!

PR #165, opened 2018-08-25
#156 Display empty table if no data is given

brianpmaher:
#156 Displays the table even when data is empty.

I understand this is fixing an issue marked as closed but I wanted to open the idea up for discussion.

I believe that allowing the users to decide what to do when the data is empty provides more flexibility.  Users who want to display some message when there is no data would need to do a conditional check on their end anyway, so it's a pretty simple change to also hide the data table for the same case if that's the behavior they want.


benjaminbours:
I totally agree with you @brianpmaher 👍

gregnb:
Sure no problem. We can leave this PR open for discussion and see where the community weighs in

jmarucha:
In my opinion:
- Writing a HOC that doesn't display a table if no data is given is extremaly easy (literally 3 lines, as ones that @brianpmaher removed)
- Writing a HOC that displays a MUI-datatable header, footer and message "Sorry, no data" like if a user added filters that returns nothing is difficult (and would probably be copy-pasting of MUI datatables

Therefore, if the user wants current behaviour, he can easy implement it himself, and if he wants something like empty MUI-datatable, he will have to do quite a lot of pointless work.

+1 for this PR

brianpmaher:
Well articulated @jmarucha

gregnb:
Thanks for all the feedback so far guys. I'll leave this open for another week for any differing opinions but it seems like we're going to merge it come next week

geekashu:
Hi,

For some reason, it is still not working for me. When an empty array is passed to the MUIDataTable whole table disappears for me. I am using `"mui-datatables": "^2.0.0-beta-28"` 

If it helps, here is the code where the problem is occurring. I am using DRF framework to return the data via Redux Saga and Axios. 

```
class Institutes extends React.Component {
     state = {
         page: 1,
         rowsPerPage: 10,
         searchText: null
     };

    componentDidMount() {
        const {page, rowsPerPage, searchText} = this.state;
        this.props.getInstitutes({page, rowsPerPage, searchText})
    }

    formatTableData = (data) => {
        return (data.results) ? data.results.map((row) => {
            return Object.keys(row).map(function (key) { return row[key]; });
        }) : [];
    };

    changePage = (params) => {
        const {rowsPerPage, searchText} = params;
        let page = params.page + 1;
        if (searchText === "" ) {
            page = 1;
        }
        this.props.getInstitutes({page, rowsPerPage, searchText})
    };

    render() {
        const {rowsPerPage} = this.state;
        const {match, data} = this.props;
        const columns = ['id', 'name', 'email', 'institute_code', 'timezone'];

        let page = 1;
        if (data.next === null) {
            page = data.previous + 1
        } else if (data.next !== undefined) {
            page = data.next - 1
        }

        const options = {
            filter: true,
            filterType: 'dropdown',
            responsive: 'stacked',
            serverSide: true,
            count: (data.count === undefined) ? 0 : data.count,
            page: page - 1,
            rowsPerPage: rowsPerPage,
            rowsPerPageOptions: [10, 25, 50],
            onChangeRowsPerPage: (numberOfRows) => {
                this.setState({rowsPerPage: numberOfRows});
            },

            onServerRequest: (action, tableState) => {
                switch (action) {
                    case 'changePage':
                    case 'changeRowsPerPage':
                    case 'search':
                        this.changePage(tableState);
                        break;
                }
            }
        };

        console.log("=========================");
        console.log(this.formatTableData(data)); // HERE DATA IS EITHER EMPTY OR FILLED WITH DATA ARRAY
        console.log("=========================");

        return (
            <div className="app-wrapper">
                <ContainerHeader title="Institutes" match={match}/>

                <div>

                    <MUIDataTable
                        title={"Employee List"}
                        data={this.formatTableData(data)}
                        columns={columns}
                        options={options}
                    />

                </div>
            </div>
        );
    }
}

```

nidheeshdas:
This is just wrong. Because when I'm searching for something which does not exist, the whole table disappears and there is no way to get back. It's like hiding the whole form if one field is invalid. 

Please revert this PR, so that user can decide what to do when there is no data.

gregnb:
Are you guys running into issues sure you are running the very latest version? Here's a screenshot of following sandbox when searching for text that does not exist:

https://wy2rl1nyzl.codesandbox.io/

screen shot 2018-09-15 at 8 15 29 am

PR #176, opened 2018-08-30
Allow sorting methods to be extensible

ccurtin:
This update would allow for custom and advanced sorting and wouldn't cause any breaking changes. This could probably be tightened up though.

One of the problems that I was personally experiencing is that when you sort filtered data--where the column data is all the same--the row order becomes randomized which isn't really an effect you'd want in more complex tables.

Included an example of how it'd be used in the `options` below. Sorts by the value of the column "data" and then by the date (a custom column of mine).

```javascript
      sortedData: (data, col, order, props, state) => {
        return data.map((row, sIndex) => ({
          data: row.data[col],
          date:
            row.data[props.columns.findIndex((x) => x.key == 'completedAt')],
          position: sIndex,
          rowSelected: state.selectedRows.lookup[sIndex] ? true : false
        }))
      },
      sortCompare: (order) => {
        return (a, b) => {
          if (a.data === null) a.data = ''
          if (b.data === null) b.data = ''

          let val_1 = a.data,
            val_2 = b.data,
            date_1 = a.date,
            date_2 = b.date

          return (
            (val_1 < val_2
              ? -1
              : val_1 > val_2
                ? 1
                : date_1 < date_2
                  ? -1
                  : date_1 > date_2
                    ? 1
                    : 0) * (order === 'asc' ? -1 : 1)
          )
        }
      }
```


ccurtin:
I totally agree that it's not ideal and a little confusing given that there's some internal workings that a user really does not need to know about. But the only concern with not allowing `sortedData` to be somewhat editable, is that it only allows sorting by a single data point. If we look at the little example I posted, in my case I needed to add another key(`date`) to the sortedTable to first sort by the initial data in the column and the secondarily by the date column. I know not the ideal fix above, but any suggestions for allowing such a feature?

gregnb:
What about something like this? if a user provides an override sort function we'll use it. This way, a user can do whatever they need to with the data as long as they return us the sorted data set we'll go ahead and build the rest for the internal usage

```
  sortTable(data, col, order) {

    let dataSrc = data;

    if (this.options.customSort) {
      // this should return the sorted data array provided by the user
      dataSrc = this.options.customSort(data, col, order);
    }
  
    let sortedData = dataSrc.map((row, sIndex) => ({
      data: row.data[col],
      position: sIndex,
      rowSelected: this.state.selectedRows.lookup[sIndex] ? true : false,
    }));

```

ccurtin:
Much cleaner!

Here's the updated example of the custom sorting option in use with the new changes which sorts by the column data first, and then by date(a column called `completedAt`)

```javascript
customSort: (data, col, order) => {
  const primarySortKey = col
  const secondarySortKey = columns.findIndex((x) => x.key == 'completedAt')

  return data.sort((a,b) => {
    if (a.data[primarySortKey] === null) a.data[primarySortKey] = ''
    if (b.data[primarySortKey] === null) b.data[primarySortKey] = ''

    // default sorting by column values.
    let val_1 = a.data[primarySortKey],
      val_2 = b.data[primarySortKey],
      // adding in custom sorting of another column
      date_1 = a.data[secondarySortKey],
      date_2 = b.data[secondarySortKey]
      
    return (
      (val_1 < val_2
        ? -1
        : val_1 > val_2
          ? 1
          : date_1 < date_2
            ? -1
            : date_1 > date_2
              ? 1
              : 0) * (order === 'asc' ? -1 : 1)
    )
  })
},
```

gregnb:
Looks good! Let’s throw in some unit tests and I’ll accept the PR

ccurtin:
I will get to this in the next week or two. Trying to wrap up a big project. But I can put an example together, but I shamefully might need a hand with writing a proper unit test.

Let's also not forget to conditionally include the default sort so that the customSort method doesn't get overridden:

```javascript
    if (!this.options.customSort) {
      sortedData.sort(this.sortCompare(order));
    }
```

gregnb:
Sure I'll keep a watch on your progress

gregnb:
hey @ccurtin just checking in any updates? let me know if you need a hand

ccurtin:
Hey @gregnb, I added in an example and updated the README. Would you be able to help me with the unit tests?

gregnb:
Hey @ccurtin thanks for the PR! I Simplified the example because it's too hard for a beginner to follow and cleaned up the README. Merged

ccurtin:
@gregnb Perfect, thanks!

Issue #154, opened 2018-08-22
Make rowsSelected  options observable like data


jkeruzec:
I'd like to programmaticaly set selected rows for the datatable. 

I'm using options and rowsSelected field with an array of row index.

When data changes or for the first time I pass on render function it's OK.
Except that when this rowsSelected changes within my render function and data is not changing, datatable is not refreshed.

So is it possible to make the option.rowsSelected obvservable so that grid change according to this new option data ?

Thanks !

jkeruzec:
I think I found the bug  : 
 First,  with this 

```js
componentWillReceiveProps(nextProps) {
    if (this.props.data !== nextProps.data || this.props.columns !== nextProps.columns) {
      this.initializeTable(nextProps);
    }
  }
```

data must be changed in a immutable way, in other word the array must be copied and reference must have change in order to go on initializeTable.
I used Array.from ... and I have to adapt with my code to test it.. 

Second : 

```js
setTableOptions(props) {
    const optionNames = ["rowsPerPage", "page", "filterList", "rowsPerPageOptions"];
    const optState = optionNames.reduce((acc, cur) => {
      if (this.options[cur] !== undefined) {
        acc[cur] = this.options[cur];
      }
      return acc;
    }, {});

    this.validateOptions(optState);
    this.setState(optState);
  }
```
When I set option.page = 0 to go back to initial page when I re render my data after an insert for example on datatable. 
this.options[page] with 0 as data is evaluated to false... with this.options[cur] !== undefined it's fine !

I think, I will have to add rowsSelection also in here to take into account selected rows passed as options.

jkeruzec:
close with #182

PR #182, opened 2018-09-03
Fix reloading options + add rowsSelected as reloaded

jkeruzec:
Add Reload for rowsSelected option
Fix reload if page is set to 0

According to #154


gregnb:
Thanks for the PR!

Issue #184, opened 2018-09-03
Server Side Support doesn't work well with selecting


kirstenland:
When using server side support, you get problems when trying to select items on the second page.

## Expected Behavior

When using Server Side support (different for without server side support):

When you click on the select all button, all rows on that page should be selected. [important]

When you try to select a row, the index given to handleRowSelect etc. should give the index for its position on the current page, not for where it would be if the previous pages were loaded. [less important, but probably the correct behaviour]

## Current Behavior

When using Server Side support:

When you click on the select all button, the rows don't seem to be selected (as in the check boxes are not full), but the handleRowSelect is given indices for its position on the current page.

When you try to select a row, the index given to handleRowSelect etc. gives the index for its position for where it would be if the previous pages were loaded.

## Steps to Reproduce (for bugs)

https://codesandbox.io/s/xvnq8456pq

The link above is the server side example provided by MUI-datatables, with the option selectableRows: true added.

1. Go to the second page of data
2. Click the select all button at the top
3. Note how nothing has been selected

## Your Environment

| Tech         | Version |
|--------------|---------|
| MUI-datatables  | 2.0.0-beta-24 |

## Fix?
Adding the line 
`if (this.props.options.serverSide) return index;`
at the beginning of the getRowIndex function in MUIDataTableBody appears to fix the problem for me.

ffarman:
+1

gregnb:
Should be good in the latest version

kirstenland:
Thanks! It is working for me now. 😄

PR #185, opened 2018-09-04
fix serverSide selectedRows bug

gregnb:


gregnb:
When merged should resolve #184

PR #190, opened 2018-09-06
Clarification on onRowSelect option

### LINKED ISSUE #163: Clarification on onRowSelect option
From the README:
> Callback function that triggers when row(s) are selected. function(currentRowsSelected: array, rowsSelected: array) => void

What is the difference between `currentRowsSelected` and `rowsSelected`?

brianpmaher:
Fixes #163 

Clarifies the meaning of the second parameter to `onRowsSelect` to imply that it is all of the rows selected.


gregnb:
Looks good. Thanks for the PR!

PR #195, opened 2018-09-16
Add a onTableChange callback to get table state upon any action

jpstrikesback:
I'd like to be able to subscribe to the table state as a whole for any action regardless of the serverSide option, in order to treat the table as more of a controlled component.

Usecase: I want to allow a user to change the shape of the table and save the structure. I am using the table as form component of sorts.

I've started with piggybacking on the setServerRequest (renamed to setTableAction), and continued with selectRowUpdate, selectRowDelete and columnViewChange.

If this is helpful and desired, great. If you think there is a better route let me know :)

Cheers,
JP

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ const columns = [
    filter: true,
    sort: true,
   }
- },      
+ },
  {
   name: "Company",
   options: {
@@ -80,7 +80,7 @@ const columns = [
    filter: true,
    sort: false,
   }
- },  
+ },
  {
   name: "State",
   options: {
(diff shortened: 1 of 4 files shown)


gregnb:
Let me think about this for some time. I'm leaning towards having one property instead of two that do the same thing and make a breaking change considering the current version is still beta.

gregnb:
@jpstrikesback Let's go ahead and update the README in the serverSide section to reference the `onTableChange` callback. Remove any mention of the `onServerSide` throughout the project and also update the example using it

I'm ready to make that breaking change and I think this is a great PR

jpstrikesback:
Awesome, will do, thanks!!

gregnb:
Thanks for the PR!

jpstrikesback:
Cheers! Thanks for accepting it!

PR #205, opened 2018-10-04
Perform search only if serverSide is falsy

nidheeshdas:
Currently, during the search when the server side data is passed on to the datatable, it again filters the data using local search logic. In the computeDisplayRow() function, it doesn't check if the options.serverSide is false or not.


gregnb:
Thanks for the PR!

PR #221, opened 2018-10-17
IE export csv bug fixed

danosbi:
Hello @gregn,

Thank you for mui-datables react component, I just added a feature to download CSV file from IE 11, and I would suggest that to add this feature on #master.

Cordially


gregnb:
I'd be happy to accept but please remove the package-lock/package changes. Just the update for the CSV portion

PR #235, opened 2018-10-22
Feature/sticky header

gregnb:
Add support for sticky header column


rcfish:
Can you confirm that adding a background color of white is causing unexpected behavior with the rows selected shadow (no longer visible) and the background color of the checkboxes for each row (is also white instead of inherited)?

tonymckendry:
The fixed header option does not seem to be doing anything for me either.

jlei523:
Any plans to implement this feature?

tonymckendry:
@jlei523 - it was working again for me after a recent update

klappy:
This feature only appears to work when the responsive mode is set to 'stacked'. 

Has anybody found a work around to get the sticky headers without enabling 'stacked' responsive mode? I've been toying around with using stacked and overriding the rendering to be identical to scroll.

klappy:
With some further research, this now makes sense as to why sticky doesn't work in scroll mode.

Via Stacked a comment on this overflow accepted answer:
"Worth noting that this won't work if the table is contained inside an element with overflow: auto or overflow-x: auto."
https://stackoverflow.com/questions/44001954/css-only-sticky-table-headers-in-chrome

Here is where overflow is set to auto.
https://github.com/gregnb/mui-datatables/blob/529307099c96ec5d983f6e4cf0f58a010ebe1c2c/src/MUIDataTable.js#L26

I've been able to resurrect the sticky positioning via customizing the MUI Theme at the cost of needing to find a solution around the overflow of content.

klappy:
For anybody interested I've been able to make both the Toolbar and the table headers sticky with this:

```js
    MUIDataTable: {
      responsiveScroll: {
        maxHeight: 'unset',
        overflowX: 'unset',
        overflowY: 'unset',
      },
    },
    MuiToolbar: {
      root: {
        top: 0,
        position: 'sticky',
        background: 'white',
        zIndex: '100',
      },
    },
    MUIDataTableHeadCell: {
      fixedHeader: {
        top: '62px',
      },
    },
```

PR #249, opened 2018-11-01
Fixed Toolbar

### LINKED ISSUE #248: [Feature Request] - Fixed toolbar
Building on #235, it would be nice to also be able to specify that the toolbar should be fixed as well. 

It should be possible to have both the toolbar and the header row fixed, or one individually at at time.

GeeWee:
Fixes #248

This PR does two things:

- Refactors propTypes for MUIDataTable out into its own file. It seemed rather unwieldy, and I wanted to reuse the options part of the propTypes (instead of just specifying object)
- It implements a fixed toolbar as well as the header. This is complicated a bit by us having to manually calculate the offset (e.g. if both header and toolbar have `position:sticky` with `top:0` they will lie on top of eachother. We start the toolbar off at position 0, and if both are marked as sticky, we adjust the `top` value of the header row.

Feel free to let me know if you need anything changed, or don't like the propTypes change.


src/MUIDataTable.js
```diff
@@ -170,6 +112,7 @@ class MUIDataTable extends React.Component {
       serverSide: false,
       rowHover: true,
       fixedHeader: true,
+      fixedToolbar: true,
```
GeeWee: Should this be the default? Breaks backwards compat, but I think it's what people expect

src/MUIDataTableBody.js
```diff
@@ -106,6 +106,8 @@ class MUIDataTableBody extends React.Component {
                     dataIndex: dataIndex,
                   })}
                   checked={this.isRowSelected(dataIndex)}
+                  fixedToolbar={false}
+                  fixedHeader={false}
                 />
```
GeeWee: I had a proptypes warning here.

src/MUIDataTableBody.js
```diff
@@ -106,6 +106,8 @@ class MUIDataTableBody extends React.Component {
                     dataIndex: dataIndex,
                   })}
                   checked={this.isRowSelected(dataIndex)}
+                  fixedToolbar={false}
+                  fixedHeader={false}
```
GeeWee: I had a propTypes warning here for the fixedToolbar and fixedHeader, so I decided to fix it. These are the SelectCells not in the header row.

src/MUIDataTableHeadCell.js
```diff
@@ -5,50 +5,55 @@ import TableCell from "@material-ui/core/TableCell";
...
-    color: "rgba(0, 0, 0, 0.87)",
-  },
-  toolButton: {
-    height: "10px",
-    outline: "none",
-    cursor: "pointer",
-  },
-};
+    fixedHeaderAndToolbar: {
+      top: "64px",
```
GeeWee: 64 px is a `min-height` I found for the toolbar I found in my developer console. However, I can't seem to find it *anywhere* in the codebase. It works in all of my testing, but it seems we might have different min heights depending on the orientation. Is there any way to do this more consistently?

src/MUIDataTableHeadCell.js
```diff
@@ -15,6 +17,9 @@ const defaultHeadCellStyles = {
     zIndex: 100,
     backgroundColor: "#FFF",
   },
+  fixedHeaderAndToolbar: {
```
GeeWee: 64 pixels is something I pulled out of my developer tools. I couldn't find it anywhere in the codebase. Is there any way we can do this better?

gregnb:
Great PR but I would want the PropTypes back where they were. Other than that I would accept and merge it. Thanks!

gregnb:
Ok I'll take a look tonight. Sorry for the delay

src/MUIDataTableHeadCell.js
gregnb: Setting the table responsive mode to 'scroll' you'll see this approach won't work. The true answer has to be something along the lines of taking the height of the toolbar and using that to get the true value to set `top` to

screen shot 2018-11-09 at 7 13 39 am

src/MUIDataTable.js
gregnb: That's fine I would expect this to be default

src/MUIDataTableHeadCell.js
GeeWee: That's true. Do you have a good idea of how to go about this?

gregnb: Take the computed height of the toolbar and then pass that as a prop. You can use props within the styles object since I'm using a modified withStyles HOC.  With that you could do something like:

``` 
   fixedHeaderAndToolbar: {
     top: props.toolBarHeight
   }
```

gregnb: Hey @GeeWee , any update on this?

GeeWee:
Closing due to reasons in #248

LandazuriPaul:
Hi @gregnb!

Thank you very much for the amazing work with mui-datatables! It's really impressive and efficient for my use cases. 

I was wondering if we could reopen this PR. Because this feature would be very interesting to me. So if you don't have time for it, I wouldn't mind giving it a try and follow on @GeeWee's work.

GeeWee:
I think you can just fork the branch and open a new PR :)

PR #250, opened 2018-11-02
Fix for changing rows per page issue

### LINKED ISSUE #242: [Bug][Per page] Page is greater than the total available
There is an exception, when I try to change "per page" option in the MUITable at the last page (data length equals to 29, and we change per page from 10 items to 15 items, and expect total 2 pages, but got the exception: ```Provided options.page of `2` is greater than the total available page length of `1` ```)
Btw, if data length equals to 30, there are no exceptions in this case.

Such problem already have been solved at #48, but seems not at all.
 

## Expected Behavior

We don't have an exception, and can see 2 pages with our data

## Current Behavior

We have exception on display ```Provided options.page of `2` is greater than the total available page length of `1` ```

Live example: https://codesandbox.io/s/6vlml849zk

1. Go to last page (page number 3)
2. Change option "per page" from "10" to "15"
3. Boom!

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    1.4.0     |
| MUI-datatables  |     ^2.0.0-beta-32    |
| React        |    16.5.0     |
| browser      |    Chrome     |

DTupalov:
Fix `MUIDataTable` changeRowsPerPage method

After merge it will fix #242

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@ npm-debug.log*
 *.sublime-project
 *.sublime-workspace
 
+# vscode project
+.vscode
+
 # Runtime data
 pids
 *.pid
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -540,9 +540,17 @@ class MUIDataTable extends React.Component {
   };
 
   changeRowsPerPage = rows => {
+    /**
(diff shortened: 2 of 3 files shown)


gregnb:
Awesome. Thanks for the PR!

Issue #266, opened 2018-11-11
[Enhancement][RFC] Pass function for managing selected rows in customToolbarSelect 


DTupalov:
We have opportunity to use `mui-datatables` as controlled component to set selected rows with `rowsSelected` and `onRowsSelect`. Either I can use it in uncontrolled `mui-datatables`, when I select rows and it keeps in state of component. But if I use `customToolbarSelect`, and receive my selected rows, I don't have any chance to manage those selected rows except to make component from uncontrolled without `rowsSelected` to controlled with `rowSelected`. But it is antipattern, for example, React give a warning in console, if you change `defaultValue` to `value` in one lifecycle on `input` element.

So, I propose to pass some function, e.g. `setSelectedRows` to `customToolbarSelect`, that will give us opportunity to manage selected rows in uncontrolled component.

```js
options: {
  customToolbarSelect: (rowsSelected, setSelectedRows) => {
    // do something with rowsSelected
    ...
    // then flush selected rows
    setSelectedRows([]);
  }
}
```

## Expected Behavior

I can manage selected rows (e.g. flush them), when using `customToolbarSelect` in uncontrolled component

## Current Behavior

I can't do anything with selected rows, when using `customToolbarSelect` in uncontrolled component
## Steps to Reproduce (for bugs)

Official example with `customToolbarSelect`
https://codesandbox.io/embed/545ym5ov6p

1. Select row
2. Click on any icon
3. You can't flush selections, or do anything with it, except to make component controlled and pass to it `rowsSelected`. But it is antipattern.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    1.4.0     |
| MUI-datatables  |   2.0.0-beta-38     |
| React        |    16.5.0     |
| browser      |   Chrome      |

gregnb:
Sounds very reasonable. If you want to take this PR up with unit tests I'm more than happy to accept

DTupalov:
OK, I'm in :)

PR #280, opened 2018-11-19
[Feature] Custom toolbar select with imperative setSelectedRows

DTupalov:
Added two additional parameters to `customToolbarSelect`: 
- `displayData` - an array of actual displaying data in the table. It's may be used for calculating next selected rows
- `setSelectedRows` - a function, that receives `selectedRows` (array of numbers - indexes of selected rows). It's may be used to set imperatively selected rows after user click on action button from your customToolbarSelect.

Also, I've updated documentation and example `customize-toolbarselect` (need to deploy to firebase).

This PR will cover #266


gregnb:
Looks good man! Thanks for the PR!

PR #281, opened 2018-11-20
Use MUI theme breakpoint media queries

### LINKED ISSUE #269: Use MUI breakpoint tools
The responsive breakpoints are hard-coded rather than relying on MUI Theme breakpoints. This results in off-by-one incongruencies.

## Reproduction

Let's say I want to hide the table and use cards for mobile. I might use the following styles which build on the MUI configured breakpoints:

Cards:
```
  cards: {
    [theme.breakpoints.up('md')]: { // generates: @media (min-width: 960px)
      display: 'none'
    }
  }
```
Table:
```
  table: {
    [theme.breakpoints.down('sm')]: { // generates: @media (max-width: 959.95px)
      display: 'none'
    }
  }
```

## Expected

At 960px, the desktop table is shown. 

## Unexpected

At exactly 960px, the table is shown but in the non-desktop view.

## Recommendation

Instead of using hard-coded styles ([example](https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTableBodyRow.js#L10) that specifies `max-width: 960px`), consider using the breakpoint tools provided by MUI to avoid off-by-one breakpoint errors.

mpareja:
Using the MUI theme breakpoints prevents off-by-one incongruencies with
applications rendering datatables using the default breakpoints. It also
enables application developers to customize the MUI breakpoints without
jarring experience when using `mui-datatables`. Closes #269.

NOTE: There were a couple places with very small media queries which don't have a MUI equivalent. I left those media queries intact: [MUIDataTablePagination](https://github.com/gregnb/mui-datatables/blob/0297bb6b21effdf1a6d4264e22b8b1307020140b/src/MUIDataTablePagination.js#L16) and [MUIDataTableToolbar](https://github.com/gregnb/mui-datatables/blob/0297bb6b21effdf1a6d4264e22b8b1307020140b/src/MUIDataTableToolbar.js#L77) (specifies no styles?)

### Testing Procedure

I manually verified that each of the updated media queries worked as expected in Chrome. (Thankfully, otherwise I might not have noticed the [conditional spread of responsive styles](https://github.com/gregnb/mui-datatables/blob/0297bb6b21effdf1a6d4264e22b8b1307020140b/src/MUIDataTableToolbar.js#L42).)

### DIFF
--- a/src/MUIDataTableBodyCell.js
+++ b/src/MUIDataTableBodyCell.js
@@ -3,13 +3,13 @@ import classNames from "classnames";
 import TableCell from "@material-ui/core/TableCell";
 import { withStyles } from "@material-ui/core/styles";
 
-const defaultBodyCellStyles = {
+const defaultBodyCellStyles = theme => ({
   root: {},
   cellHide: {
     display: "none",
   },
   cellStacked: {
-    "@media screen and (max-width: 960px)": {
+    [theme.breakpoints.down("sm")]: {
       display: "inline-block",
       backgroundColor: "#FFF",
       fontSize: "16px",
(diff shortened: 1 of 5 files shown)


gregnb:
Thanks for the PR!

Issue #275, opened 2018-11-16
Download CSV feature not working with customBodyRender 


diegochavez:
Hello
The CSV export using an older version such as beta-21 it's working the custom render, after upgrading to beta-38 it's render [object,object] instead of plain text

## Expected Behavior
current version beta-21
https://codesandbox.io/embed/xrvrzryjvp?autoresize=1&hidenavigation=1

## Current Behavior
latest version beta-38
https://codesandbox.io/s/m9z107j30x

## Steps to Reproduce (for bugs)
click download and check the columns is displaying the [object,object] instead of the text.

![export-csv-error](https://user-images.githubusercontent.com/35136/48618083-80341d80-e998-11e8-93db-cf8cbed9513d.png)

![image](https://user-images.githubusercontent.com/35136/48620301-e96b5f00-e99f-11e8-9df6-70b6dd07d2ee.png)

rudrabot:
I am also having the same issue
![screenshot 2018-11-24 at 10 10 45 pm](https://user-images.githubusercontent.com/35026134/48970754-22f23900-f036-11e8-8371-a759b60810f5.png)

diegochavez:
Solved with #287 :smile: :+1:

sriram4493:
Facing the same issue.
customHeadRender: ({ index, ...column }) => {
            return (
              <TableCell key={index} >
                Heading
                <TableCell >Subheading1</TableCell>
                <TableCell >Subheading2</TableCell>
                <TableCell >Subheading3</TableCell>
              </TableCell>
            )
          },
customBodyRender: ({ index, ...column }) => {
            return (
              rows.map(row => (
                <TableRow >
                  <TableCell >{row.a}</TableCell>
                  <TableCell>{row.b}</TableCell>
                  <TableCell>{row.c}</TableCell>
                </TableRow>
              ))

PR #287, opened 2018-11-24
I've tested based on issue #275 the download CSV

diegochavez:
Using data instead of displayData? there's a specific functionality missing if we use data? we can prevent the React elements?

### DIFF
--- a/src/MUIDataTableToolbar.js
+++ b/src/MUIDataTableToolbar.js
@@ -84,7 +84,7 @@ class MUIDataTableToolbar extends React.Component {
   };
 
   handleCSVDownload = () => {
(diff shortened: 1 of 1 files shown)


Legogris:
With this change I am able to set `toString` on data objects and make the result show up in the generated CSV.

gregnb:
Thanks for resolving this issue. Merged!

darkhallow:
where is the file "MUIDataTableToolbar.js" location?

mxdavis:
Still having this issue on nested objects. https://codesandbox.io/s/mutable-darkness-685n4?file=/src/index.js

PR #288, opened 2018-11-26
Pass full row rather than single value to getTableMeta rowData parameter when rendering rows

xantorres:


Badji:
Any news on this PR @gregnb ?

nathancharles:
Is there any update on this PR?

SergeyShiryaev:
The problem is important, actually. I have a lot of crutches to get around the problem of returning one value. This fix should fix the problem.

gabrielliwerant:
Can someone explain what was broken before that this PR fixes with a specific example?

patorjk:
After investigating this, this does appear to be fixed. Closing.

PR #301, opened 2018-12-03
Feature detect "Intl" and use plain string comparison when not available

Skn0tt:
First of all, thank you for creating this great package!

I recently discovered that in some browsers (like Android Web View or older versions of mobile Safari), there is no support for the `Intl` API. In order to prevent crashes on these devices, I propose to feature detect for `Intl` and fall back to a plain string comparison in cases where `Intl` is not available.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -15,15 +15,15 @@ import isEqual from "lodash.isequal";
 import textLabels from "./textLabels";
 import { withStyles } from "@material-ui/core/styles";
 
-const standardCollator = (a, b) => a.localeCompare(b);
+export const fallbackComparator = (a, b) => a.localeCompare(b);
 
-const getCollatorComparison = () => {
+export const getCollatzComparator = () => {
   if (!!Intl) {
     const collator = new Intl.Collator(undefined, { numeric: true, sensitivity: "base" });
     return collator.compare;
   }
 
-  return standardCollator;
+  return fallbackComparator;
 }
 
 const defaultTableStyles = {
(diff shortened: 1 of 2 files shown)


mantzuri:
This breaks for older browsers, now for iOS 9.3.5 the table does not work at all, maybe consider adding compatibility check before using the Intl.

Skn0tt:
@mantzuri maybe I misunderstand you, but this is the exact intention of the PR. Are there any other compatibility checks to make that I'm unaware of?

mantzuri:
Maybe it is unrelated to the PR, I didn't review it. 
I was about to open a new issue, related to `Intl`.
The thing is, I'm working on a website for quite a while using Datatables, and today I tried to use it on old iPad 2 (iOS 9.3.5). The website crashed due to `Intl`. I patched the problem as follow:

```
var v = function iOSversion() { // Checking for iOS version
	if (/iP(hone|od|ad)/.test(navigator.platform)) {
		// supports iOS 2.0 and later: <http://bit.ly/TJjs1V>
		var v = navigator.appVersion.match(/OS (\d+)_(\d+)_?(\d+)?/);
		return [parseInt(v[1], 10), parseInt(v[2], 10), parseInt(v[3] || 0, 10)];
	}
};
var isNew = v === undefined || v[0] > 9;
.....
if (isNew) {
	var i = new Intl.Collator(void 0, {...
.....
if (s.options.sortFilterList && isNew) {
     var l = new Intl.Collator(void 0, {...
```

I am sure that there might be a better way, but it seem to work for me

Skn0tt:
The exact problem you experienced is addressed by this PR and will fix it.

gregnb:
Approved. Thanks for sending this in @Skn0tt !

Issue #125, opened 2018-08-06
Feature Request: Custom filter list


jmarucha:
For the data provided from external source, such as database, providing custom list of avialable filters would be extremaly useful.

## Current Behavior

The filter options are based on the data passed to the component. For the data:
```
[1, 'a']
[2, 'b']
[3, 'a']
```
avialable filters on column 2 are 'a' and 'b'
## Expected Behavior

A user should be able to provide list of filters to override the possible filter options via custom column field. Passing
```
columns = [
 {
     name: "ID",
     options: {
       filter: false,
       sort: true,
  }
  {
    name: "value",
    options: {
      filter: true,
      filterOptions: [
        'a', 'b', 'c', 'd'
      ]
    }
  }
];
```
As columns option should make filtering by 'a', 'b', 'c', 'd' on column 'value' avialable regardless of options.

gregnb:
There is an option to allow you to specify the filter list. Take a look at example located here:
https://github.com/gregnb/mui-datatables/blob/master/examples/serverside-options/index.js

gregnb:
Sorry, misunderstood a provided filterList for filterOptions. Ok, I'll get back to you on this.

toolaugh:
I have the same problems like @jmarucha . I want to custom filter list menu options . and wonder how to custom ?

toolaugh:
I also want to custom styles of filter view like width and height ? Let me know how to how to use ? @gregnb

phelion:
Hi, what is the progress with this request? I'd really need to add custom filter options, because I don't always query all the data from the database, and then the values what are not in the list, but it's possible to exist, not showing in the filter options.

gregnb:
This is now available in the latest release. There are two column options `filterList` and `filterOptions`.  I have removed the `filterList` from the root level options and now they must be provided per column.

Look at the README for an example. Thanks for all your patience.

dafapramudya:
Hi @gregnb how can I custom filter type on each column? for ex: I have 3 columns: username(filterType: 'textField'), date(filterType: 'multiselect'), age(filterType: 'dropdown').

I already tried this on each column, but didn't work:

```
column: [
      {
        name: 'Username',
        options: {
          filter: false,
          sort: false,
          filterType: 'textField',
        },
      },
      {
        name: 'Date',
        options: {
          filter: true,
          sort: false,
          filterType: 'multiselect',
        },
      },
      {
        name: 'Age',
        options: {
          filter: true,
[...]

PR #303, opened 2018-12-04
#125 Add custom filter list options

lukas-zech-software:
This adds 3 new options for the column configuration:

`customFilterRender: (column, index, props) => ReactNode`
This provides the possibility to override the default rendering of the `FilterInput`
For example this can be used to render a DatePicker instead of a DropDown

`customFilterFn: (filterValue, columnValue) => boolean`
This enables the user to provide an own filtering implementation. Given the value of the column and the value of the filter, the function return either true or false if it want the row filtered out or not.
This also make filtering for greater/lower than or range filter possible

`customFilterValueRender: (columnValue) => ReactNode`
This makes it possible to define how the filter value is rendered in the active filter list.
Necessary eg. for custom date rendering.

All of these methods are showcased in the new example `customize-filtering` where I added a Datepicker to filter a column of date.

I could not find where to update the docs with the new method, if you point me to it, I'll gladly update them.

If anything should be changed, please let me now


gregnb:
Nice work 👍 Can you please look at https://github.com/gregnb/mui-datatables#column-options and update the README explaining these methods. Then I'll do a code review from there.

I like it overall and will look to merge it in but I'm going to see when I get some free time if there are any improvements I can make

lukas-zech-software:
I updated the docs with the new methods and commented the example as well.

Its a functional solution but not a very nice one. The APIs of the functions could propably be way nicer but I think that would need more internal refactoring.

Also the new functionality is only implemented for filterType dropdown but could be easily expanded to the other, given the time

lukas-zech-software:
I updated the API for the `customFilterRender` function.

Its way cleaner now.

One big problem remains, when the user returns a React element and not plain text in the `customFilterValueRender` function.

As the `filterList` doesn't not hold the actual filter values but just the Elements rendered in the filter list, the user must manually extract the value again, which is not a big problem but counter intuitive.

I documented that in the ReadMe and in more detail in the example.


Could probably be improved by holding the actual filter values in the state as well, but that would be a bigger refactoring

lukas-zech-software:
Ok, so I further improved the changes by refactoring the internal handling of the filter list.

Before they were simple strings, now they are stored as the full ReactComponents that will then be rendered  in the filter list.
This allows full controller over the filter chips e.g. attaching own events, rendering Links or even complex controls like buttons or dropdowns. So you could for example change a filter directly in the chip or link to the site of a filtered category.

This also allows to store the raw value of the filter as prop instead of relying on the rendered text to display the correct value.
The formatting / displaying of the value is now decoupled from the value. The value could now actually be a complex object.

Also the multiselect can now do the custom filter rendering, checkbox ist still missing though.

gregnb:
Ok, I still haven't had any free time yet. In fact, I'm working on a weekend. I will try to get to this sometime soon. My overall goal with a review is to see if we can somehow condense the API surface of these new features. I'm sure that won't  be an easy task but I need to at least try.

gregnb:
Ok, I will try to get it when I have some down time. Please bare with me. Thanks for all the work in the mean time

lukas-zech-software:
No worries, just let me know if if I should change anything on this PR.

I updated the PR on my fork with another API proposal which I find even nicer.
Once you had the time review the proposal, let me know if you are okay with the API approach and I'll start, if you're not, its fine either.

alxbbnk:
It will be useful for my use cases.  
Do you plan to write tests on this functionality ?

lukas-zech-software:
@alxbbnk Not until @gregnb has approved the general approach.

Tests in enzyme are very time intensive and I do not want to invest that for a feature that still might be rejected.

gregnb:
Yes please do not start any work yet. This is going to take some time I have a general vision for refactoring this project that is playing into this decision as well.

tspvivek:
@lukas-zech-software  While trying to merge your code with modifications i have done for my project specific requirement, I noticed that in `filterUpdate` function definition the line for finding the `filterPos`  is returning -1 always while removing the filter. 

I think the `filterpos ` should be populated using below line

`const filterPos = filterList[index].findIndex(x => x && x.props.rawValue[0] === filterValue[0]);`

instead of 

`const filterPos = filterList[index].findIndex(x => x && x.props.rawValue === filterValue);`

gregnb:
@lukas-zech-software sorry for the delay. I've done some refactoring of the project and I want to move forward with what you've done here instead of the proposal. 

Just need to get this PR back in working order and add some unit tests for these changes and I will merge.

Thanks again!

src/MUIDataTable.js
```diff
@@ -41,6 +41,11 @@ const TABLE_LOAD = {
   UPDATE: 2,
 };
 
+// Wrapper Component to keep the rawValue
+function FilterValue(props) {
```
gregnb: Can we move this to line 669? since it isn't referenced anywhere else I would rather have it in the class

lukas-zech-software:
@tspvivek 
sorry for the late answer

`rawValue` can never be an array, it is the exact value from your data. At least as long as you did not have an array of arrays in your data.

I will investigate the issue
Can you send me a probe of your data so I can test it?

src/MUIDataTable.js
lukas-zech-software: This is a React functional component, I think it its quite unusual to have that as a property on a class, isn't it?

gregnb: Let's separate it out in a different file then

gregnb: Actually don't worry about this

gregnb:
@lukas-zech-software Can we get some unit tests in? I would love to approve and merge this after

lukas-zech-software:
@gregnb I'll try to get at this soon, but I'm currently stuck in a huge amount of work.

Any help would be very welcomed @alxbbnk ?

gregnb:
Yes open call to anyone who can pitch in on unit tests for this PR

VagnerNico:
@lukas-zech-software i have a problem in columns with customBodyRender in witch i render images or icons. So far i could not figure out how to customize the filters to match my jsx elements.
I remember having a similar issue on jQuery's DataTables and the fix whas customize filters and use spans hidden with same text.

lukas-zech-software:
@VagnerNico 
sound like you need to implement a custom `customFilterRender` function to show the filters you want.

The `customFilterRender` function always gets passed the real data value and does not need to extract the value from the jsx itself

see https://github.com/lukas-zech-software/mui-datatables/blob/master/examples/customize-filtering/index.js#L30 or the readme

VagnerNico:
@lukas-zech-software, i followed your example in this sandbox [](https://codesandbox.io/s/vnnxmok6wl) for birthday and the filter field was not a text input.
I don't know if i misunderstood the feature, but the idea is be able to change the input type, values and behavior for specific column filter, am'i correct?

lukas-zech-software:
@VagnerNico Have you installed the package from my fork?

This feature is not yet merged and not available on the official version of the package.

If you want this feature you will need to install it like this
```json
// package.json
"dependencies": {
  "mui-datatables": "git://github.com/lukas-zech-software/mui-datatables.git#temp-release",
}
```


If you want the help get this feature in the official version, you can help out and write unit tests fo the new feature

VagnerNico:
Ok. I will try to write some tests.

JRamos1993:
Any news on this?

JRamos1993:
I would if I could, 0 experience at unit testing atm.

simonkaspersen:
Is this planned to be implemented soon?

lukas-zech-software:
@simonkaspersen  as soon as somebody helps out on the unit tests @gregnb requested.

I'm currently working on two full-time projects and have no time left to invest in this

lukas-zech-software:
I just merge the latest master, which also gets more and more time consuming the longer this stays open ...

So either some help out on the test, we skip them or close this PR as I currently cannot invest the time for proper unit tests, especially not with enzyme

I've using my branch in a project for 3 Months now, so I'm pretty sure is stable, but of course its only tested for my use case

lukas-zech-software:
The latest changes on master are just too incompatible with this PR.

Keeping a branch open so long inevitably makes it unmergable ...

simonkaspersen:
This was sad. Did just discover this library and this PR, and can't use it without more customizable filter and search.

lukas-zech-software:
@simonkaspersen 
Sadly, the internal workings of this library suffer from supporting string, arrays, objects and ReactElements all in the same methods just seperated by a bunch of `if` statements.

And that was just to much for me to keep updated and merged for almost for month ..
If you want just the customizable filter you can try the version from my branch
```
// package.json
"dependencies": {
  "mui-datatables": "git://github.com/lukas-zech-software/mui-datatables.git#temp-release",
}
```
but of course this is now stale ...

I was thinking about porting and refactoring this library to Typescript for a while, but this will take a month or so


If you want to try to finish the merge of this branch, you're very welcome to to so
On my master branch https://github.com/lukas-zech-software/mui-datatables I already merged and fixed all syntax errors and fixed a bunch of logic error, but the filtering logic is now incompatible and must be refactored.
The part is covered with tests though, so it should be easy to check if the refactoring works

PR #307, opened 2018-12-11
add custom styling methods for rows and cells

### LINKED ISSUE #304: Customize row theme based on data
Hi,
I already use the customization of the theme like this:
`var theme = createMuiTheme({
      overrides: {
        MUIDataTableBodyCell: {
          root: {
            fontSize: ".9rem",
            fontWeight: "bold",
          }
        },
        MUIDataTableBodyRow: {
          root: {
          }
        }
      },
    });`

but now I need to customize the row based on content like : column "date_end" is different from "" so backgroundColor: red.. I have found how do it possible based on the index but not we the data. Someone can help me? Thanks

lukas-zech-software:
fixes #304 

This adds 2 functions
On table options:
`getRowClassName`
 which is called for each row and adds classes returned from it to the `MUIDataTableBodyRow`  object .
Allows styling of each row based on its data

And `getRowClassName` on column options, which does the same for cells

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,11 +163,11 @@ The component accepts the following props:
 |**`onChangeRowsPerPage`**|function||Callback function that triggers when the number of rows per page has changed. `function(numberOfRows: number) => void`
 |**`onSearchChange`**|function||Callback function that triggers when the search text value has changed. `function(searchText: string) => void`
 |**`onSearchOpen`**|function||Callback function that triggers when the searchbox opens. `function() => void`
-|**`onFilterChange`**|function||Callback function that triggers when filters have changed. `function(changedColumn: string, filterList: array) => 
-void`
+|**`onFilterChange`**|function||Callback function that triggers when filters have changed. `function(changedColumn: string, filterList: array) => void`
 |**`onColumnSortChange`**|function||Callback function that triggers when a column has been sorted. `function(changedColumn: string, direction: string) => void`
 |**`onColumnViewChange`**|function||Callback function that triggers when a column view has been changed. `function(changedColumn: string, action: string) => void`
 |**`onTableChange`**|function||Callback function that triggers when table state has changed. `function(action: string, tableState: object) => void`
+|**`getRowClassName`**|function||Is called for each row and allows to return custom className for this row based on its data. `function(row: array) => string`
 
 ## Customize Columns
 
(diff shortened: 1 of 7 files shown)


gregnb:
I like it but one concern:

1) Someone will probably ask how can we do inline styling not just classses at some point. How could we offer something clean in one call for this? 

I would also change these to match the other fn calls. 

customRowStyle={}
customCellStyle={}

Let's talk about some approaches here before doing any more code

lukas-zech-software:
we could rename the methods or maybe group them like so
```javascript
var options = {
  rendner: {
    customBodyRender...
    customFilterRender...
  },
  styles:{
    customRowStyle={}
    customCellStyle={}
  }
}
```

For the question how to combine className and inline style in one method we could introduce a `customRowProps` method wich gets the row and returns an object wich will then be spreaded in the props of the `TableCell`. This would basically allow to overwrite each and every prop, so could also apply custom handlers etc.
But this would be dangerous as the user could overwrite necessary internal props.
The props should probably be filtered and eg. classNames must be merged, so maybe this should be limited styles and className

lukas-zech-software:
I factored that into my new API proposal, it fits in quite naturally there.
Please have a look https://github.com/lukas-zech-software/mui-datatables/pull/1/files#diff-83c315224b438550e5d9d961b017b5e1R49

Basically it would look like

```jsx
<MUIDataTable>
         <MUIDataTable.Row>
          {/*called for each row can do formatting or even wrapping in custom components*/}
          {(value, children) => (
            <MUIDataTableRow className={classnames({'isActive': value[1].isActive})}
                             style={{color: value[1].isActive ? "red" : "blue"}}>
              {children}
            </MUIDataTableRow>
          )}
        </MUIDataTable.Row>
{...}
</MUIDataTable>
```

lukas-zech-software:
Another pattern I just came across are `propertiesGetter functions`

A function cann be provided as proper for each element, like cell or row. The function gets passed the current data und returns an object that will then be spread as props on the element.
By this you can edit every property of the element
They would be way easier to introduce and would look like this


```jsx
 rowPropGetter={(row) => {
  return {
    className: row.value ===1 ? "rowClass": "",
    style: { color:'red' },
    onSomeEvent={...}
   };
}}
```
```jsx
 cellPropGetter={(cell) => {
  return {
    className: row.value ===1 ? "rowClass": ""
   };
}}
```

gregnb:
I love this idea, but shouldn't we call it a setter since it is setting (spreading) props onto the cell/row?

setRowProps
setCellProps 

what do you think?

lukas-zech-software:
I think your proposal fit quite well

`setRowProps`
`setCellProps`
.. etc

This should be implement for all elements.

If you're fine with this approach, I'll refactor this branch to use this pattern

gregnb:
yep lets do it. I'll approve this change. the other PR i'm still thinking about how to best go about it. such a big sweeping change

lukas-zech-software:
I updated Row and Cell to the new API.

Feels cleaner and allows for more flexibility.

I moved the spread to the top of the properties so all internal properties will always overwrite the user provided

gregnb:
I love how simple it is and will solve lot of customization issues. Great work. I’ll be merging this tonight if you have any other changes or improvements you wish to make to this PR let me know and i’ll Hold off.

I’ll be doing a release shortly after

PR #308, opened 2018-12-11
Add support for object data source for row

RahulKumarSaini:
Add support for array of object as data source.


src/MUIDataTable.js
```diff
@@ -347,15 +360,18 @@ class MUIDataTable extends React.Component {
...
     let isSearchFound = false;
     let displayRow = [];
 
-    for (let index = 0; index < row.length; index++) {
-      let columnDisplay = row[index];
-      let columnValue = row[index];
+    displayRow = rowDataSource === "object" ? {} : [];
 
-      if (columns[index].customBodyRender) {
+    Object.keys(row).forEach((key, index) => {
```
lukas-zech-software: That will probably not work reliably as the order of properties on an object is not defined.
https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties

That means, that the index might not match to the provided column options object.

lukas-zech-software: Note: The order is same within one implementation of a javascript engine, eg. in Chrome it might always work.
But in FireFox, Safari, Edge or IE it will be different for each.

Thats why this approach seems to be working during development, but is not cross browser compatible

lukas-zech-software: You can use this code to create an array with an defined order from your object

´´´javascript
const dataObjects = [{prop1:1,prop2:2,prop3:3},{prop1:1,prop2:2,prop3:3}]
const dataArray = dataObjects.reduce<any[]>((all, data) => {
      all.push([
        data.prop1,
        data.prop2,
        data.prop3,
      ]);
      return all;
    }, []);
´´´

This then simply work with the current API

RahulKumarSaini: Yes, you are right we can covert the array of object as you suggested. But I have also fix my implementation without using Object.keys. 
Please have a look

RahulKumarSaini:
I have also fix my implementation without using Object.keys.
Please have a look

RahulKumarSaini:
@gregnb Can you please review this pull request I have add support for array of objects as data source as mention in this #308

gregnb:
This PR will not work anymore. The project was refactored you are applying changes to files that have been renamed.  I am open to supporting array of objects so please make those changes and then add in a unit test that shows we are still supporting the older implementation of providing Array of Arrays

gregnb:
The latest release now supports data as objects. Example: https://github.com/gregnb/mui-datatables/blob/master/examples/data-as-objects/index.js

Issue #305, opened 2018-12-06
Table sort is lost if data changed on a re-render


joaoaguiam:
If user has sorted a column and the table data changed afterwards, the table is render with the new data but the sort is not taken into account.

## Expected Behavior
When the table data is updated, the re-rendering of the MUI DataTables should take into account the previously selected sort by the user.

## Current Behavior
Currently, in this situation, the rows are back to the initial order and not sorted correctly. However, the column header still indicates the sort icon correctly (based on the previously clicked column).

## Steps to Reproduce (for bugs)
I am storing the table data in the state and derive it from the props. If the props change, the data in the state is updated. The MUIDataTable component is retrieving the data directly from the state on the render method.

## Your Environment
| Tech         | Version |
|--------------|---------|
| Material-UI  |      1.4.3   |
| MUI-datatables  |      2.0.0-beta-38   |
| React        |    16.5.0     |
| Browser      |      Chrome   |

hwatersiv:
I have encounter this same issue. Anyone know if there is a work around until this is resolved?

cwkeller:
I was using react.  I created a wrapper component that would save off the state of the table and re-apply it during render.  Because of a sort issue (#273) I was also sorting the table and saving that to updatedColumns.  The important code is below.

```
const options = {
      filter: true,
      selectableRows: false,
      filterType: 'dropdown',
      responsive: 'stacked',
      rowsPerPage: this.state.tableState ? this.state.tableState.rowsPerPage : 10,
      onTableChange: (action, tableState) => {
        this.state.tableState = tableState;
      },
      filterList: this.state.tableState ? this.state.tableState.filterList : filterList,
    };
    return (
      <MUIDataTable
        title={title}
        data={dataSorted || data}
        columns={updatedColumns}
        options={options}
      />
    );
```

gregnb:
Bump. Can anyone confirm?

cwkeller:
Looks like it is still doing it for me with 2.0.0 beta 48 but I could be doing something wrong.

the-noob:
FWIW sorting doesn't work at all for me with updated props. I have an async action retrieving the listing and when it completes, while the table is updated, the sorting doesn't work at all.
2.0.0.beta-50

joaoaguiam:
For me, the version 2.0.0-beta-42 fixes the issue.

asharafshahi:
[...]

PR #316, opened 2018-12-14
Fix issue with sort not taking into account on re-render

joaoaguiam:
Fixing issue #305


gregnb:
Great thanks for fixing this isssue!

PR #318, opened 2018-12-15
Fix `onFilterChange` line in README

betoharres:
I spotted a little mistake on the README on the line about `onFilterChange`, the commit just fixes the breakline before `avoid` word.


gregnb:
Awesome thanks for catching this!

PR #319, opened 2018-12-17
Custom column hint tooltip

Legogris:
This adds the option to add a custom hint tooltip to column headers by setting the `hint` property on column options.
image


gregnb:
Nice, great idea! Merged

Legogris:
Sweet, thanks @gregnb :)
Any idea when we can expect the next npm release?

PR #321, opened 2018-12-19
Add ability to override column header

fantapop:
Some background here.  Column header can sort of be overridden by using the customHeadRender option on the column object however if you use this, you lose a lot of out of the box functionality including sorting, adjustable column widths and column hint.  Also, customHeadRender is not applied in the responsive view so if you change the names of the columns, they'll revert when on mobile.


gregnb:
More specific, this PR is an attempt to provide the following to each column:
 id, displayName

is that correct?

src/MUIDataTable.js
```diff
@@ -268,6 +269,8 @@ class MUIDataTable extends React.Component {
 
         columnOptions = {
           name: column.name,
+          label: column.label,
```
gregnb: Lets change this to 

```
label: column.label ? column.label : column.name
```

gregnb: Also, instead let's make `label`  always defined that way we can remove all these guards we have in the other files checking if label || name.

That will require changing: https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L296
to this:
```
        columnOptions = { ...columnOptions, name: column, label: column };

```

gregnb:
Ok I've left some comments and the project structure overall changed so if you can make those adjustments I'll review again

fantapop:
Great. Will do.

fantapop:
@gregnb I've made these updates.  Take another look when you have a chance.

gregnb:
Looks good @fantapop can you put in some unit tests to confirm that if label is provided that it renders that label and if not provided it uses name? i'll be ready to approve after

fantapop:
Yeah. I was looking for an example test which is checking for what is
rendered. Is there one I can model it after?

russell-ingram:
Any idea when this will be merged? Was looking for a fix for this exact issue before using mui-datatables to replace some functionality in my existing app.

fantapop:
I’ll try to figure out some tests this evening.

fantapop:
@gregnb added some tests here.  Let me know if there's a better way.

fantapop:
@gregnb have you had a chance to look over this one? Thanks

gregnb:
Sorry for the delay. Just took a look and it's great. Thanks again for your hard work! Merged

PR #331, opened 2018-12-26
Add jsonMode option

mgolkardev:
Add jsonMode option for support array of object data

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,68 @@ const options = {
 
 ```
 
+Or customize columns with JSON data:
+
+```js
+
+import MUIDataTable from "mui-datatables";
+
+const columns = [
+ {
+  name: "Name",
+  field: "name",
+  options: {
+   filter: true,
+   sort: true,
+  }
+ },
+ {
+  name: "Company",
+  field: "company",
+  options: {
+   filter: true,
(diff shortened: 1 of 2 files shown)


gregnb:
So people have been asking about accepting array of objects for a long time now. I will be willing to accept this PR with some changes:

1) I don't want an explicit option so can we just detect if a user is passing in an array of arrays (current implementation) and if they are not then handle the array of objects
2) Change 'field' to 'id' and document this in the README. Specify how `name` is the label and `id` is the true column id
3) Something in the README to show that it accepts both array of arrays and array of objects

abrhim:
I could really use the JSON mode in my project and I am willing to finish this up. @mgolkardev, I saw that you worked on it Jan. 5th. What remaining work is there?

apercas:
If @abrhim couldn't work on this I'll be glad to help and move this forward.

JRamos1993:
This PR as it is will prevent other functionalities like Sorting/CustomBodyRender, etc from working correctly.

gregnb:
The latest release now supports data as objects. Example: https://github.com/gregnb/mui-datatables/blob/master/examples/data-as-objects/index.js

PR #337, opened 2019-01-04
Fix sorting when customSort doesn't sort inplace

Psykar:
When using the `customSort` prop, if the sort function provided doesn't sort `data` in place but returns a copy of `data` sorted, the original (unsorted) `data` is used to pick indexes... which results in no sorting happening in the table.

This PR ensures that the sorted list is used when reading indexes.


gregnb:
Hi, thanks for the PR but can you go into details as to what the bug was, what you expected, and what this is solving. Thanks

gregnb:
Great, thanks for fixing this bug!

Issue #342, opened 2019-01-05
Text Field Filter Type


loganasherjones:
First, thanks for the great library. I would be interested in implementing this for you if you agree with the feature.

## Expected Behavior

**Use Case:** When working with some external data sources, its impractical or impossible to know all possibilities for all columns. For example, in my application, I would have to do a full table scan to figure out all the possible values for each of the columns. However, we do have a filter API that would allow people to type text and we can search for that text on a particular column.

**Suggestion:** Add a `textField` option to the `filterType`. This would render a text field in the filter toolbar that would allow users to type what they want and filtering would occur the same way the search works today, only specified down to a column.

This closely resembles #294 but is designed to update the filters, not the search functionality.

gregnb:
@loganasherjones Thanks for the kind words. 

Sounds reasonable but just be mindful of the filter badge that appears after a user selects something to filter by. We only want to create one badge tied to this column and update or remove complete when they have cleared the text field.  

Your use case is useful for when the table goes `serverSide` but what about people who would potentially use this without serverSide? in that case all the values are provided.  I would like to see examples of both in your implementation.

PR #346, opened 2019-01-07
Adding "textField" filter type (#342)

loganasherjones:
Adds a `textField` type to the `filterType` options. 

This PR allows users to select `textFeild` for the filter type, which will allows users to start typing to filter a particular column.

Here is what it looks like when the filter is active:

filter_active

Here is what it looks like when the filter tab is open:

filter-open


gregnb:
Great. Thanks for the PR @loganasherjones

PR #300, opened 2018-12-02
Changed fixed colors to mui palette colors

idobh2:
First of all - awesome component! thanks 👍 

This PR aims to introduce some dynamic color styling based on the material-ui palette object.
One good use case for it is using a dark mui palette, like partially mentioned in #272, which results in a "broken" table header, selection cells and label (sue to fixed color styling).

So with these changes, the following theme passed to a wrapping MuiThemeProvider actually renders a fully dark themed table

```
createMuiTheme({
  palette: {
    type: "dark",
  },
})
```

One thing I did leave is the checkboxes coloring, which by default uses the "secondary" color. The stated color (#027cb5) seemed too accurate, so i figured someone really wanted this color. I can change this too, to either secondary or primary if required. or it can stay this way and be set through style overrides. your call.

PR #348, opened 2019-01-07
Updating colors to work with the MUI themes.

loganasherjones:
This is essentially a copy of #300 but updated to work with the
new layout. It also replaces all colors I could find to be sure
they refrence the themes and still look good on both light and
dark themes.

See #300 for what the screenshots look like as they are
an example. Also, I added a new theme example that will render
dark by default.


gregnb:
Approved! I will merge this in later with some other bug fixes and do another release

Issue #253, opened 2018-11-04
When serverSide is true. Empty data list doesn't display proper messages.


geekashu:
I am using the latest build `2.0.0-beta-37`. I tried this with `serverSide` example from this location https://github.com/gregnb/mui-datatables/blob/master/examples/serverside-pagination/index.js

Problem is when the data is empty and `serverSide: true` the table doesn't display any message. Instead, it just displays an empty table. 
![screenshot 2018-11-04 23 32 07](https://user-images.githubusercontent.com/920804/47968000-f0ce6680-e089-11e8-874d-42ed7c635ba3.png)

Ideally, it should display the table with a message as shown in the image below.
![screenshot 2018-11-04 23 33 41](https://user-images.githubusercontent.com/920804/47968016-1f4c4180-e08a-11e8-8543-4275adeaa2ac.png)

Would appreciate if you could guide me on how I can resolve the message issue with `serverSide: true`.

Also, how do I customize this message per table?

gabrielliwerant:
Closing, as I believe this issue has been resolved (as shown below). Additionally, examples for how to display loading indications have been added to the example here https://github.com/gregnb/mui-datatables/pull/583.

![capture](https://user-images.githubusercontent.com/1779564/57352824-8a0f3000-7134-11e9-9c04-2ea38faa2b58.gif)

PR #374, opened 2019-01-20
Fix bug #253

diegochavez:
When serverSide is true empty data list doesn't display proper messages. #253


gregnb:
Great. Thanks for the PR!

PR #377, opened 2019-01-21
Change title variant

lorena-cerbino:
h6 is no longer used on Typography variants: expected one of ["display4","display3","display2","display1","headline","title","subheading","body2","body1","caption","button"].


gregnb:
Hi @lorena-cerbino thanks for the PR but it seems like title will be deprecated:

https://material-ui.com/style/typography/#migration-to-typography-v2

PR #384, opened 2019-01-23
Added option "selectAllDisabled" to disable the checkbox "select all"…

Execomsrl:
For a project of mine i needed to disable the checkbox "select all" on table header. So i added a new boolean option "selectAllDisabled". I think it can be useful to other people too.


gabrielliwerant:
Being addressed here https://github.com/gregnb/mui-datatables/pull/882.

PR #390, opened 2019-01-25
bugfix: add dataIndex to onCellClick meta

gregnb:


Aaron-Lei:
I just noticed your bugfix about adding dataIndex to onCellClick meta, I'm wonder if it is possible to add it to 'column > customBodyRender func' too, 
When we wanna edit the value of each cell, the MUI-datatables provides us customBodyRender function , and the 'updateValue' in it can automatically refresh it to the table. but in my demand scenario, when I wanna format the list data in the table to a post request body,the property named "data" ref to the first loaded data must update itself corresponding to each cell updating, but I can not using setState when each cell updating, because In select mode, I do not want it rerender, that will leading the previous selected rows will refresh and the check status will dismiss, so I need the dataIndex to update the "data" property immutably.
Did I make sense? If not, please feel free to let me know. 
Thanks

Issue #408, opened 2019-02-08
Search issue


IvanGrimes:
## Expected Behavior
Search should not be applied to the excluded columns.

## Current Behavior
Search applied to the excluded columns.

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/00n97l290n

1. Click search icon
2. Type any value that match to user's id

PR #415, opened 2019-02-12
Exclude hidden columns from search results

rem4ik4ever:
#408 
Don't query for hidden column values


gregnb:
Great thanks for the PR!

rem4ik4ever:
@gregnb 
That was quick, thanks!

PR #416, opened 2019-02-12
Adjust TableBodyCell colSpan to exclude display:"false" columns as we…

jasontrigg0:
…ll as display:"excluded" columns


gregnb:
Great thanks for the PR!

Issue #299, opened 2018-12-02
default filter is throwing TypeError: Cannot read property 'toString' of undefined


dyoung522:
The following code snippet displays the DataTable properly, but then throws `TypeError: Cannot read property 'toString' of undefined` when the 'filter' icon is clicked.

Not sure if this is a bug, or I'm missing something.

```
    const tableOptions = {
      filter: true,
      pagination: false,
      download: false,
      selectableRows: false,
      print: false,
      viewColumns: false
    }

    const charColumns = [
      { name: 'ID', options: { display: false, filter: false, sort: false, download: true } },
      { name: 'Name', options: { display: true, filter: true, sort: true, download: true } },
      { name: 'Level', options: { display: true, filter: true, sort: true, download: true } },
      { name: 'Race', options: { display: true, filter: true, sort: true, download: true } },
      { name: 'Class', options: { display: true, filter: true, sort: true, download: true } },
      { name: 'Sex', options: { display: true, filter: true, sort: true, download: true } },
      { name: 'Owner', options: { display: true, filter: true, sort: true, download: true } }
    ]

    // Creates an array of arrays from Firestore data
    const charData =
      isLoaded(characters) && isLoaded(users)
        ? Object.keys(characters).map(id => {
          const char = characters[id]

          return isEmpty(char)
            ? []
            : [id, char.name, char.level, char.race, char.class, char.sex, users[char.owner].displayName]
        }).sort((a, b) => a[1] - b[1])
        : [['loading', 'Loading... please wait']]

    return (
        <Paper className={classes.root}>
          <MUIDataTable title={'Guild Roster'} data={charData} columns={charColumns} options={tableOptions} />
        </Paper>
    )
```
Here's the raw `charData`:

```
[Array(7)]
0: (7) ["5GM0mnQs2xZP1zCms9xK", "Squall", "5", "Tabaxi", "Monk", "Male", "Donovan C. Young"]
```

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  3.6.0  |
| MUI-datatables  |   2.0.0-beta-41 |
| React        |   16.6.3  |
| browser      |  Chrome  |
| etc          | react-redux-firebase v.2.2.4 |

dyoung522:
Interestingly, if I use the raw `charData` array (hardcoded) it works. So there must be something about the way I'm creating the `charData` above, but I haven't see any differences in the console. Both seem to produce the identical array or arrays.

[...]

PR #422, opened 2019-02-15
Allow for checking null and undefined.

jdetle:
Hi there, I think this was causing a bug for me when I was using the filter table functionality.

### DIFF
--- a/src/components/TableFilter.js
+++ b/src/components/TableFilter.js
@@ -207,7 +207,7 @@ class TableFilter extends React.Component {
                 </MenuItem>
                 {filterData[index].map((filterColumn, filterIndex) => (
                   <MenuItem value={filterColumn} key={filterIndex + 1}>
-                    {filterColumn !== null ? filterColumn.toString() : ''}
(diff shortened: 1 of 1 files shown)


gregnb:
@jdetle Thanks for the PR. Can you give me some detail as to what was going on?

jdetle:
I cant give a minimum reproducible example but I had a basic table and was pressing the filter button but this gave an error where filterColumn was undefined, but still being referenced. I think this should prevent that from ever happening.

aprakash-sovrn:
I believe this addresses https://github.com/gregnb/mui-datatables/issues/299, no?

jdetle:
Yeah #299 looks right.

jdetle:
🙏 thanks for the merge !

aprakash-sovrn:
Did this go out in beta.55? Because I am still having this problem.

chennanoka:
same here, updated to beta.55 still have the problem.

gregnb:
it did go out but i've also received another PR with one more fix. try .56

PR #431, opened 2019-02-20
fix setCellProps return type in README

AntoineDoubovetzky:
Hi !

I think there is a small mistake in the documentation :)


gregnb:
Nice catch! :)

PR #440, opened 2019-02-23
Fix download filtered data only

faran312:
This fix issues for download filtered data only. I think you have been fixed this before, but in #329 you've changed this, and become not worked. Give your feedback if i have some mistakes.

### DIFF
--- a/src/components/TableToolbar.js
+++ b/src/components/TableToolbar.js
@@ -85,8 +85,8 @@ class TableToolbar extends React.Component {
   };
 
   handleCSVDownload = () => {
-    const { data, columns, options } = this.props;
-    createCSVDownload(columns, data, options);
+    const { displayData, columns, options } = this.props;
+    createCSVDownload(columns, displayData, options);
   };
 
   setActiveIcon = iconName => {


gregnb:
I think this keeps going back and forth. We need to have some sort of option that toggles between showing all data or filtered data. Please reply back with an idea for that API can we can work from there.

Thank you for submitting.

shawnmitchell:
The issue seems to be that there are two filters being applied in `filteredData`: selected columns and visible rows. 

In my case, users want to download all rows in the table with columns filtered. 

Seems like the ideal solution might be a bitwise operator. 

```
const Columns = 1;
const Rows = 2;

downloadFilterOptions: (Columns | Rows)

// if (downloadFilterOptions === undefined) download all rows, all columns
// if (downloadFilterOptions === 1) download all rows, selected columns
// if (downloadFilterOptions === 2) download current visible rows, all columns
// if (downloadFilterOptions === 3) download current visible rows, selected columns
```

gabrielliwerant:
@shawnmitchell How about something like this for the API:

```
downloadOptions: {
  filename: 'tableDownload.csv',
  separator: ',',
  filterOptions: {
    useDisplayedColumnsOnly: true | false,
    useDisplayedRowsOnly: true | true
  }
}
```

Default for `filterOptions` would be `false`, in which case it works as it would normally.

- If `useDisplayedColumnsOnly: true` and `useDisplayedRowsOnly: false`, downloads only columns that remained through filtering, and all rows.
- If `useDisplayedColumnsOnly: false` and `useDisplayedRowsOnly: true`, downloads only rows that remained through filtering, and all columns.
- If `useDisplayedColumnsOnly: false` and `useDisplayedRowsOnly: false`, downloads all table data, as in the default case.
- If `useDisplayedColumnsOnly: true` and `useDisplayedRowsOnly: true`, downloads only columns and rows that remained through filtering.

I'm open to suggestions for shorter, but equally clear option names. I'm hoping that whatever we pick will be self-documenting enough to avoid confusion.

ali26Y:
can someone please merge this????

gabrielliwerant:
@ali26Y This PR can't get merged as is because it needs a more robust API. Some users will want displayed data only, others will want all the data. Feel free to submit a PR that provides an API similar to the one mentioned here and I will be happy to review it.

gabrielliwerant:
Options now available for behavior like this in `2.7.0`.

PR #441, opened 2019-02-25
Support for an initial state

nrdev88:
I implemented the possibility to pass an initial state to the MuiDataTable plugin. This initial state, does what it says, sets the already existed state. I had to re-arrange some of the logic to accept an initial state, since it wasn't build for that.

- Moved `page` and `rowsPerPage` from `options` to the new `initialState` prop
- Fallback support for `page` and `rowsPerPage` under `options` prop
- Implemented `searchText`, `filters`, `sort` and `columnOptions` (for visibility)
- Supports `serverside: true`
- Adjusted readme
- Added example


jpambrun:
Thanks for this contribution. I am trying it out.

 Given that you moved rowsPerPage and page to the new prop would it make sense to move count as well?

jpambrun:
Filters looks strictly equivalent to filterList. I think the name change is a bit confusing.

jpambrun:
It would be nice to have initialState members optional. It currently fails if you don't supply a sort member.

nrdev88:
> Thanks for this contribution. I am trying it out.
> 
> Given that you moved rowsPerPage and page to the new prop would it make sense to move count as well?

I did think about it, but somehow it didn't make sense for me. I just don't know why anymore, can't find any good reasons not too.

> Filters looks strictly equivalent to filterList. I think the name change is a bit confusing.

I find `filterList` confusing. But I can change that. You're right about consistency.

> It would be nice to have initialState members optional. It currently fails if you don't supply a sort member.
> 
> Edit1: I see that you are merging defaults, not sure what is going on. Hold on..
> Edit2: OK, I had set sort = null instead of sort.column and sort.direction.

True point. Should fix that :-)

jpambrun:
Just to be clear, I am not a contributor to this project.. just some random dude and this PR is just *exactly* what I needed this morning. I am already using it, it's awesome. Thanks!

gregnb:
hey @nrdev88 thanks for the PR. This is quite a large change. I'll need some time to think about this

Dragomir-Ivanov:
I would love to see this merged :+1:

dave-continuum-media:
Is there any update on this PR?  Would really love to see this added, but if there's any issue holding it up, could they be shared so we potentially address them and move this along? :)

Dragomir-Ivanov:
@gregnb Do you need any help merging this? This could be very helpful to hydrate the table, with previous saved state.

corroleaus:
@gregnb I'm also very interested in this change. Would be great to see it merged.

gregnb:
So my questions are why would we need an initialState? Is there something you can not already do by the available options? I don't want the package to get to the point where we have multiple options/props that do the same thing

jpambrun:
We need that so when the user hit back he/she can resume at the same page,
with the same filters/search/etc. We need a way to rehydrate the table.

gregnb:
@jpambrun To be more specific what options are not currently available as props that are you stopping from rehydrating the table?

nrdev88:
Currently you can’t set options for selected filter values, current sorted column, etc.

gregnb:
Anyone interested in this thread can we please list out the options they are missing when it comes to rehydrating the state. When we compile a list I'll look into it and see if someone isn't aware of an option and what is really missing that needs to be added.

@nrdev88 Selected filter values: https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js

Dragomir-Ivanov:
Current filter + sort column+order is needed for the exact same purpose. When table is unmounted I am going to save current table state, then when user hits the back button, he can pick where he left off.

matt-continuum-media:
@gregnb We have the same requirements as @corroleaus and have implemented what we need in a fork. We didn't want to duplicate functionality from this PR so haven't submitted our own PR yet but you can take a look: https://github.com/continuum-media/mui-datatables

paulmichaelxd:
@gregnb I also have the use case where I would like to rehydrate the table with previous user settings. For my specific requirements I needed everything addressed in this PR. I am currently just using 

To list it out:
* page (current page)
* rowsPerPage
* searchText
* filters (for all types)
* column sort (ASC or DESC, which column)
* column order
* column visibility

All effort on this is much appreciated.

Dragomir-Ivanov:
Any idea for the status of this PR?

karthick-vinod:
@paulmichaelxd @gregnb Correct me if I'm wrong.

> * page (current page)
> * rowsPerPage

can be hydrated using `page` and `rowsPerPage` of `options` prop

> * filters (for all types)
> * column visibility

can be hydrated using `filterList` and `display` column level options.

> * column order

can be achieved by ordering the columns as required in `columns` prop

> * column sort (ASC or DESC, which column)
> * searchText

These are the things which are actually missing.

gabrielliwerant:
@karthick-vinod Thank you for the summary!

We also have the column option `sortDirection` which sets the default sort order. I think this could be used for state hydration of the column order, right? If so, it seems that we only have one option needed, which would be `searchText`.

@nrdev88 What I would like to see, perhaps in a new PR (since this one has fallen out of date and has a lot of prettier changes), is an option added for `searchText` and maybe something for column order if `sortDirection` isn't enough for some reason. I think we should also leave the prettier changes to perhaps a separate PR, as it adds quite a lot of noise to the diffs and makes it difficult to review the essential features here. I would be happy to accept such a PR. What do you think?

karthick-vinod:
@gabrielliwerant You are correct.

* **searchText**: Please find my PR for searchText option here: https://github.com/gregnb/mui-datatables/pull/603

* **columnSort**: As `sortDirection` is a column level option, assigning sortDirection to multiple columns actually sorts only the last column though Arrow Icon could be seen next to multiple columns according to sortDirection option specified. This might be a bug.

gabrielliwerant:
@karthick-vinod Yep, saw your PR just after I commented! I left a review, but looks great!

gabrielliwerant:
Ok, so the `sortDirection` behavior of only affecting the final column I believe is intentional, as multiple columns cannot be sorted simultaneously, as this require reordering individual cell data among rows, thus scrambling the table. However, it is unclear what is happening from a UI perspective, as it does show both arrows.

What should probably occur is that only the final arrow should show, and there should be a console warning that multiple columns have been set to be sorted.

I opened an issue around this here https://github.com/gregnb/mui-datatables/issues/605.

gabrielliwerant:
This should be resolved via https://github.com/gregnb/mui-datatables/pull/603

PR #460, opened 2019-03-04
FIX onRowClick being triggered by row (de)selection

JReinhold:
At the moment, if you have an `onRowClick`-handler on a selectable row, the handler will be called when the user clicks the checkbox. I'm unsure if this is on purpose or not, but I have a hard time coming up with a scenario, where you want the click handler to be called on selection as well.

This PR fixes that, so the `onRowClick`-handler doesn't get called if the click happened on a select checkbox.

This is technically a breaking change for anyone that actually wants the onRowClick to be triggered on selection as well. If that is indeed a use case, maybe we should add a configuration of some kind, either like an option, or by returning a boolean from `onRowsSelected` that determines if the select should call `onRowClick` as well.

Right now, our temporary fix for this is:
```js
onRowClick={(row, _, event) => {
  if(event.target.type === 'checkbox'){
    return;
  }
  ...
}}
...
```

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -104,6 +104,14 @@ class TableBody extends React.Component {
     this.props.selectRowUpdate('cell', data);
   };
 
+  handleRowClick = (row, data, event) => {
+    // don't trigger onRowClick if the event was actually a row selection
+    if (event.target.id && event.target.id.startsWith('MUIDataTableSelectCell')) {
+      return;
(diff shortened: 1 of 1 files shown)


gregnb:
Thanks for the PR @JReinhold and you are correct there is no purpose to having it here. Merged

nthgol:
How do I install this version? @next?

Issue #443, opened 2019-02-26
[BUG]Data as objects Error


Aaron-Lei:
For example, when I use the last column for actions(buttons for adding,refreshing ..etc operations), the TypeError 'Cannot read property 'toString' of undefined' occurred. The code is as follows:

`import React from "react";
import ReactDOM from "react-dom";
import MUIDataTable from "mui-datatables";

class App extends React.Component {
  render() {

    const columns = [
      {
        name: 'name',
        label: 'Name'
      },
      {
        name: 'company',
        label: 'Company'
      },
      {
        name: 'city',
        label: 'City'
      },
      {
        name: 'state',
        label: 'State',
      },
      {
        label: 'Action',
        name: 'action',
        options: {
          customBodyRender: (value, tableMeta, updateValue) => {
            return <button>Add</button>
          }
        }
      }
    ]

    const data = [
      { name: "Joe James", company: "Test Corp", city: "Yonkers", state: "NY"},
     { name: "John Walsh", company: "Test Corp", city: "Hartford", state: "CT"},
     { name: "Bob Herm", company: "Test Corp", city: "Tampa", state: "FL"},
      { name: "James Houston", company: "Test Corp", city: "Dallas", state: "TX"},
    ];

    const options = {
      filterType: "dropdown",
      responsive: "scroll"
    };

    return (
      <MUIDataTable
        title={"ACME Employee list"}
        data={data}
        columns={columns}
        options={options}
      />
    );
  }
}

ReactDOM.render(<App />, document.getElementById("root"));`

But when I add action='anything' to the data structure, likes 
`const data=[{ name: "Joe James", company: "Test Corp", city: "Yonkers", state: "NY", action: "**anything**"}]`, It works fines, that means I have to change my data structure by adding the properties named the same as action columns names. It is not good specially in the case that the data is fetching from server-side, it can't to change server-side interfaces to meet the rules of mui-datatables.

Did that make sense to you? 
Any response will be appreciated.
@gregnb 

## Expected Behavior

The mui-datatables can avoid mapping the columns that not exist in the data structure
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

gabrielliwerant:
I believe I have a workable solution. Ok if I take this? @gregnb

gregnb:
[...]

PR #470, opened 2019-03-06
Missing data for custom action columns

gabrielliwerant:
Addresses https://github.com/gregnb/mui-datatables/issues/443.

There were two issues.

1. Missing fields in data were causing errors
1. Data arrays of arrays did not properly render `customBodyRender` when data was missing

Added an example under `custom-action-columns`. Not sure what the best name for this kind of use case would be, so I'm open to suggestions.

Let me know what you think.

UPDATED:

~~Added an API option for columns called `hasData` which defaults to `true`. When `false`, it signifies that a column should be skipped when checking for data. This is only relevant for the array of arrays data structure, but allows full use of empty columns or `customBodyRender` in those cases.~~

~~Not sure that the API option is super obvious, but I couldn't think of another way to distinguish data in the array of arrays structure as intentionally absent.~~

Added an API option for columns called `empty` which defaults to `false`. When `true`, it signifies that a column should be skipped when checking for data. This is only relevant for the array of arrays data structure, but allows full use of empty columns or `customBodyRender` in those cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ const columns = [
 |Name|Type|Default|Description
 |:--:|:-----|:--|:-----|
 |**`display`**|string|'true'|Display column in table. `enum('true', 'false', 'excluded')`
+|**`hasData`**|boolean|true|The column has corresponding data for display/mutation purposes
 |**`filterList`**|array||Filter value list [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
 |**`filterOptons`**|array||Filter options [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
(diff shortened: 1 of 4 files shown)


gabrielliwerant:
Just noticed an edge case here where no-data columns other than the last row cause the indexing to get messed up when transformed. Working on it and will push fixes here.

gregnb:
What if we named it `empty` instead? and make the README something like `This denotes with the column has data or not` ?

gabrielliwerant:
Updated with the API change.

gregnb:
Ok great! I'll review this if I can get some free time tonight. Thanks again

gregnb:
@gabrielliwerant Approved. I will merge this when I do the next release in the next couple of days

PR #474, opened 2019-03-08
Added padding, cursorOnHover, customCSVBody, maxHeight control

chrisswordfish:
Found this component really useful in my application but needed a bit more control over a few components. Specifically I used this table to render oData endpoints so everything was done server-side. Added some style control to table options and a customCSVBody property which allows you to configure a seperate call to retrieve bulk data for CSV rather than simply exporting whatever is on the current page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,17 +150,20 @@ The component accepts the following props:
 |**`elevation`**|number|4|Shadow depth applied to Paper component
 |**`caseSensitive `**|boolean|false|Enable/disable case sensitivity for search
 |**`responsive`**|string|'stacked'|Enable/disable responsive table views. Options: 'stacked', 'scroll'
+|**`maxHeight`**|string|'499px'|Control height of the table when in scroll mode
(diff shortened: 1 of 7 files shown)


gregnb:
hey chris thanks for the PR. I like the custom function for csvbody but the styling options I feel like those can be done via a mui theme provided. Did you try that route?

chrisswordfish:
Hey, thanks for the email. Yeah you're probably  right, I did use muitheme overrides to do them in the first place but the code didn't look as neat haha. Not fussed if you don't want to keep them. The padding option however I think is useful to keep because it is a custom property on the original MuiTable API, it's not simply CSS like the others.

Cheers,
Chris.

On 8 Mar. 2019 8:13 pm, gregn <notifications@github.com> wrote:

hey chris thanks for the PR. I like the custom function for csvbody but the styling options I feel like those can be done via a mui theme provided. Did you try that route?

gregnb:
Hey @chrisswordfish I'm going to decline the PR as I don't want to bloat the API surface. Thank you for sending.

Issue #472, opened 2019-03-07
Body scrolling active when viewColumns toggle is open


william-chu:
## Expected Behavior
Similar to the rowsPerPage option toggle, the body tag should be given a style of `overflow: hidden` to prevent the user from scrolling the background while selecting the columns to be viewed

## Current Behavior
When you open the viewColumns toggle menu, the body scrolls while the toggle is absolutely positioned in the fixed overlay. When there are multiple tables on a page, the placement of this menu's position should not move since it loses its position in context to the table it belongs to.

## Your Environment

| Tech         | Version |
|--------------|---------|
| @material-ui/core |    ^3.8.2    |
| MUI-datatables  |     2.0.0-beta-50    |
| React        |    ^16.3.1    |
| browser      |   Chrome  72.0.3626.119   |
| etc          |         |

gregnb:
Could you post a video/gif of what you are seeing?

william-chu:
@gregnb Here is a video of it, the rowsPerPage select menu seems to add the overflow: hidden style to the body tag which locks the background scrolling, but the viewColumns does not. https://drive.google.com/file/d/13-PyvFloiuDZ_KJrhiDgNWOs5FA9OQXL/view?usp=sharing

gregnb:
@william-chu thanks for the video. Would you like to look into this issue? Maybe if we set a container

https://material-ui.com/api/popper/

we can inject it in the right place without it being fixed. 

also:
https://material-ui.com/utils/popper/

```
The children is Portal to the body of the document to avoid rendering problems. You can disable this behavior with disablePortal.
```

william-chu:
@gregnb Thanks for the prompt response! I submitted a PR that should disable the body scroll when the filter/viewColumns buttons are selected by removing the container props on these popovers. https://github.com/gregnb/mui-datatables/pull/479

PR #479, opened 2019-03-09
Removed container prop on toolbar popovers to prevent body scroll

william-chu:
For issue https://github.com/gregnb/mui-datatables/issues/472

Removed the container prop for the filter and viewColumns buttons in the `TableToolbar.js` file. Without this, the container defaults to body and applies the `overflow: hidden` style to prevent the body from scrolling.

This matches the behavior of the rowsPerPage select menu.

### DIFF
--- a/src/components/TableToolbar.js
+++ b/src/components/TableToolbar.js
@@ -188,7 +188,6 @@ class TableToolbar extends React.Component {
           {options.viewColumns && (
             <Popover
               refExit={this.setActiveIcon.bind(null)}
-              container={tableRef}
               trigger={
                 <IconButton
                   aria-label={viewColumns}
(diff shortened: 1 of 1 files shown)


gregnb:
Thanks @william-chu I will merge this fix. Ideally I think the Popover should be migrated to use Popper which mui has written a quick abstraction over. Popper will create popovers that are positioned absolutely I believe so if a user scrolls up/down it'll still be in the correct spot. Would you be interested in taking that up?

william-chu:
@gregnb Yep, that seemed to be the difference from what I was reading, I'll try to get it switched over to a Popper once I find some time to look at it more closely. Should the Popover component remain in place if/when I try to make those changes? Not sure if it was being used anywhere else besides the default toolbar...

PR #484, opened 2019-03-11
Add mouse down event in cells

Jopenergy:
Sorry about that but I did not notice that Chrome does not trigger Event on middle click anymore using the onClick property. But it's still triggering using the MouseDown property.

Thank you !


patorjk:
Just going through old PRs. In the current version of the table this can be accomplished via the setCellProps method.

Issue #427, opened 2019-02-19
className prop not being used


ashams1026:
https://github.com/gregnb/mui-datatables/blob/caf05577c00193c03b2217d562b6604612ddc11f/src/MUIDataTable.js#L116-L117

It looks like somewhere down the line the use of className passed into MIUDataTable stopped being used.

gregnb:
I believe because you can do most of the table styling through the MUI provider. Do you want to send in a PR to clean this up?

ashams1026:
True! New to MUI, so I'm just getting the hang of styling. I was just confused at first cuz I couldn't get it to work, took me a bit to realize it wasn't going anywhere after passing it through props.

Can't quite make any contributions at the moment, unfortunately. Great work btw!

seansean11:
I would actually be interested in getting this to work, so I submitted a PR #486 .  In our project, we would prefer to toggle different styles based on class overrides.

PR #486, opened 2019-03-12
feature(MUIDataTable): pass className prop to underlying table

seansean11:
PR references #427 
- wire up className prop to component for users that are not using CSS in JS
---
In our case, we are using Material-UI with SCSS and `.scss` files for each component.  We'd like to be able to add classes to this component in order to override styles, rather than set up an MUI provider for the different instances of the component.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -9,7 +9,7 @@ import TableBody from './components/TableBody';
 import TableResize from './components/TableResize';
 import TableHead from './components/TableHead';
 import TableFooter from './components/TableFooter';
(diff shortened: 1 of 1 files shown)


src/MUIDataTable.js
```diff
@@ -9,7 +9,7 @@ import TableBody from './components/TableBody';
 import TableResize from './components/TableResize';
 import TableHead from './components/TableHead';
 import TableFooter from './components/TableFooter';
-import TablePagination from './components/TablePagination';
```
seansean11: - This import was unused, so I ✂️ removed it.

Issue #491, opened 2019-03-13
customBodyRender fails on null data


gabrielliwerant:
~~**Incoming PR**~~
**PR up**

When data is `null`, `customBodyRender` encounters an error and the table does not display.

## Expected Behavior
When `null` data is encountered by the `customBodyRender` function, it should properly parse the value as `null` and the table should display with the appropriate rendering of the value.

## Current Behavior
Results in console errors which prevent continued execution of the script, and the table fails to render.

## Steps to Reproduce (for bugs)
Change one of the "Location" values to `null` in `examples/component/index.js` and attempt to view table in a browser.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  | ^3.2.0 |
| MUI-datatables  | 2.0.0-beta.58 |
| React  | ^16.2.0 |
| browser  | Chrome 72.0.3626.121 |

PR #492, opened 2019-03-13
Issue 491

gabrielliwerant:
Addresses https://github.com/gregnb/mui-datatables/issues/491.


ninkambazi:
hallow there is any thing like incremental rows adding rows and Colspan or rowspan?

PR #501, opened 2019-03-19
Added options for data manipulation onDownload of CSV file

mellson:
I've added the possibility to add custom header and footer names to the csv output. I needed this for a project and hopefully it's useful to someone else.

If this gets merged I'll add the types to the DefinitelyTyped repo.


mellson:
Hi @gregnb 👋Thanks for the great work on this library! If you get an available moment I would love some feedback on this PR, hopefully it's useable for others than myself.

gregnb:
@mellson Thanks for submitting the PR. Just got back from a long vacation. Allow me some time to review this and see if I have any changes.

mellson:
Hi @gregnb hope you've landed on your feet again. Have you had some time to look through this one yet? (sorry if I'm too impatient)

gabrielliwerant:
@mellson Hello, I took a look at your PR, thanks for your work!

However, I just want to make sure I understand the use case here. Excel-style files have a concept of headers/footers that are part of meta-data, but as I understand it, csv files do not. So what is created here is another means of modifying the column labels for export only, and an additional row of different column names as the final data row:

image

So, I think it's possible that calling them `header` and `footer` could lead to confusion as they don't appear to serve the same role that they would in a typical spreadsheet, so maybe different terminology is in order.

Also, can you explain the use case for having a bottom row of different column names?

My concern would be that we might be adding specific functionality here that has a very narrow use case. Another option would be instead to provide an `onDownload` callback which would allow customization of output before downloading. This would have the advantage of not specifying very specific sets of features which are the only avenues for customization, but rather allow the full range of use cases for anyone. What do you think @gregnb?

mellson:
@gabrielliwerant thanks for the great feedback 👏 

I think you are on point with all your comments.
I've added this because we have an integration with another system which requires us to emulate headers and footers in this way - e.g. a very specific and narrow use case as you mention.

So maybe the `onDownload` callback option would indeed be a better approach. Do you have any further thoughts around that or should I just try to implement my own thoughts of how that would work?

gabrielliwerant:
@mellson Thank you so much for graciously accepting the feedback!

Greg will have final say, so I'm 100% sure what he would accept, but if it were up to me, I would accept something like `onDownload(buildHead, buildBody)` where `buildHead` and `buildBody` are the respective functions for creating those parts of the csv and that accept `columns` and `data` as their arguments. `onDownload` would probably need to be called just before the final csv data is converted to a Blob, and should return a csv of the same format if `onDownload` is truthy.

Might need to also pass `columns` and `data` along from the util function into the callback. I like parsimonious functions wherever possible, but including `columns` and `data` as part of the callback arguments ensures that any modifications made to them elsewhere can be utilized by the invoker of the function in cases where they were not managed directly via state when passed into the table.

Also, just a reminder that the PR has fallen behind, so will need latest from master merged in. Sorry for the inconvenience.

mellson:
@gabrielliwerant thank you for the input, I will have a look at it in the coming days and update this PR when I have an initial implementation for `onDownload`

mellson:
Hi @gabrielliwerant I did a quick first attempt here, if you like this direction I will tidy it up and update the README accordingly.

gabrielliwerant:
@mellson Thanks for taking another crack at this!

I think we can mostly leave `MUIDataTable.js` alone for now, and confine most of the changes to `utils.js`. What I was thinking was exposing a callback in the main options rather than in `downloadOptions`, much the same way we already have a bunch of accessible callbacks related to toolbar functionality, such as `onSearchOpen` and `onFilterChange`.

My thoughts on the utils code was something like the following:

```
function createCSVDownload(columns, data, options) {
  ...

  const buildHead = columns => {
    return columns
      .reduce(
        (soFar, column) =>
          column.download
            ? soFar + '"' + replaceDoubleQuoteInString(column.name) + '"' + options.downloadOptions.separator
            : soFar,
        '',
      )
      .slice(0, -1) + '\r\n';
  };
  const CSVHead = buildHead(columns);

  const buildBody = data => {
    return data
      .reduce(
        (soFar, row) =>
          soFar +
          '"' +
          row.data
            .filter((field, index) => columns[index].download)
            .map(columnData => replaceDoubleQuoteInString(columnData))
            .join('"' + options.downloadOptions.separator + '"') +
          '"\r\n',
        [],
      )
      .trim();
  };
  const CSVBody = buildBody(data);

  const csv = options.onDownload ? options.onDownload(buildHead, buildBody) : `${CSVHead}${CSVBody}`.trim();

  ...
```

Then you would set it table options like so:

```
  const options = {
    filter: true,
    filterType: 'dropdown',
    responsive: 'scroll',
    onDownload: (buildHead, buildBody) => buildHead(this.state.columns) + buildBody(this.state.data)
  };
```

The idea here is to keep the footprint as small as possible, so the only code we'd really need to add is just to convert the code already being used to functions we can pass into our callback and add `onDownload` to props. This should enable all the customization with a minimal amount of code change, so as to reduce the chance of breaking existing things.

mellson:
Hi @gabrielliwerant 👋 thanks again for talking another look at this.
I've tried to do a version based on your input!

I have however added the column and body data to the callback to give that flexibility. That way we ensure that the user can have total freedom in choosing which data and which state of the data gets written to the file.

I have removed the export of the column builder as well.

How does this version look to you?

README.md
```diff
@@ -163,6 +163,7 @@ The component accepts the following props:
 |**`print`**|boolean|true|Show/hide print	 icon from toolbar
 |**`download`**|boolean|true|Show/hide download icon from toolbar
 |**`downloadOptions`**|object||Options to change the output of the CSV file. Default options: `{filename: 'tableDownload.csv', separator: ','}`
+|**`onDownload`**|function||Callback function that triggers when the user downloads the CSV file. `function(buildHead: (columns) => string, buildBody: (data) => string, columns, data) => string`
```
gabrielliwerant: Let's add a note here about being able to use this function also to customize the csv output.

mellson:
@gabrielliwerant good point on keeping the folder names consistent, I have updated the name. I've also updated the README a bit and tried to add a test (keep in mind this is my first ever mocha test, so I hope it works 🤔 )

README.md
mellson: Done, added in the last commit

gabrielliwerant:
This looks great to me! Test works wonderfully. Marking this for final approval by @gregnb.

gregnb:
This is great I like what this PR turned into. Merged!

AndrePoletto:
Guys, I was trying to use this function on my project and wasn't able to. I even tried with the onDownload example code, but it seems not to work (The headerNames doesn't change).
Can you help me?

mellson:
If you can post a question on stack overflow I'll be happy to take a look 👍

AndrePoletto:
I already solved it @mellson, thank you anyway for the attention. Just one more thing, you could update the Readme file, adding a link to the code example.

Issue #442, opened 2019-02-26
Columns resize on updated props


klappy:
When using column resizing, the first column grows and the others shrink when data is manipulated.

I'm passing the data in as props so that may be causing the table to re-render and recalculate the widths.

## Expected Behavior

When new props are passed and data is updated, the columns should not recalculate their widths.

## Current Behavior

Every time an action is performed the first column appears to grow in width. This causes all others to shrink in width and is impractical to resize each column to get them to return to normal.

This happens during adding/removing rows, or editing cells. You can also import the same file over and over and watch the behavior as well.

## Steps to Reproduce (for bugs)

![column-resizing](https://user-images.githubusercontent.com/118073/53378696-94db6680-3934-11e9-9030-baeecb2c823c.gif)

1. visit: https://klappy.github.io/tsv-editor/
2. Use this file to import after removing the .txt as it only supports .tsv files.
[example.tsv.txt](https://github.com/gregnb/mui-datatables/files/2903198/example.tsv.txt)
3. Click in a field to "edit".
4. "Tab" to cycle through next and previous cells to see resizing.

You can also clone the repo and after `npm i` and `npm start` you can run locally.
https://github.com/klappy/tsv-editor

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    ^3.9.2     |
| MUI-datatables  |    ^2.0.0-beta-54     |
| React        |    ^16.8.2     |
| browser      |   latest Chrome on Mac     |
| etc          |         |

klappy:
@gregnb My apologies that my first feedback is a bug. This is an excellent project and I am appreciative of the amazing pace at which feature and bugs are addressed! Excellent job delivering and maintaining such a versatile component!

gregnb:
No worries @klappy thanks for the feedback. Honestly, I have noticed this issue for a while now and I don't have a great approach on fixing it yet. I would love for someone to pitch in on this one since I'm strapped for time

gabrielliwerant:
I have a partial solution for this, fyi. PR incoming.

gabrielliwerant:
Just a reminder, should this ticket be closed? @gregnb

gregnb:
Closed! Thanks again for solving this @gabrielliwerant

klappy:
Thanks guys!! This is a great addition to a great project.

PR #502, opened 2019-03-19
Issue 442 - Column Resizing on Data Updates

gabrielliwerant:
Addresses https://github.com/gregnb/mui-datatables/issues/442.

- Adds demonstration code to the resize example
- Fixes bug in resize by preventing resize calculations when there is no actual resizing (hence, vanilla state updates will not recalculate width)
- Add some minor code quality improvements (orphaned vars, misspelled function)

There is still a slight stutter sometimes when state is changed before or after a column resizing event. An alternative fix which addresses everything including the stutter would be to remove the `componentDidUpdate` here: https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L183. It's a much simpler fix, but I assumed that this might have unintended consequences, so I went for the slightly more complex fix.

### DIFF
--- a/examples/resizable-columns/index.js
+++ b/examples/resizable-columns/index.js
@@ -4,9 +4,35 @@ import MUIDataTable from "../../src/";
 
 class Example extends React.Component {
 
-  render() {
+  constructor(props) {
+    super(props);
 
-    const columns = ["Name", "Title", "Location"];
+    this.state = { counter: 1 };
+  }
+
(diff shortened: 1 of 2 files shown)


gabrielliwerant:
All tests are passing. There currently isn't any testing for TableResize at all. Please advise.

gregnb:
@gabrielliwerant That's fine. This is awesome. Thanks for nailing down this issue

PR #533, opened 2019-04-03
Prevent closure of search bar  when any other toolbar buttons are clicked

karthick-vinod:
Before this change, if user enters text in search bar and then clicks on filter icon (or any other icon on toolbar), the search text filed disappears. After this, the search bar does not disappear if it contains any search text.


gregnb:
Thanks for the PR. I'm back from a long vacation. I will review this shortly.

gregnb:
Approved & Merged. Thank you!

PR #535, opened 2019-04-03
Reference MUI v3 instead of v1

shawndibble:
Currently you are referencing v1 in your doc for material ui, however it is using the current version. Updating the document's reference.


gregnb:
Thank you for the PR!

Issue #407, opened 2019-02-07
Rendering of filters


arturoribes:
## Expected Behavior

When filters are rendered as chips, they should show the column name and not only the value. This is particularly noticeable when you have multiple yes/no columns, so you don't know which filter corresponds to which column (e.g. in case you want to remove it)

## Current Behavior

Chips show only the value of the filtered column

## Steps to Reproduce (for bugs)

Filter a column and watch the generated chip

gabrielliwerant:
`customFilterListRender` is now available, as is custom rendering of filters in general in `2.6.0`. See documentation and examples for more information.

PR #552, opened 2019-04-09
Rendering of filters (#407) 

giupea:
Added customFilterListRender to column options (identity function as default)


gregnb:
hey @giupea can you explain the use case for these changes? also provide an example. Someone has attempted this once in the past and I'm open to the change but this PR is light on details/README as well

giupea:
Hi @gregnb, 
these changes provide the ability to define a custom label for the Chip component in TableFilterList.
As pointed out in #407, showing the filter value only can lead to ambiguous chips.
With these changes, one can add a 'customFilterListRender' property to column options to define what should be shown in column filterList chips.
I've just added to the PR changes to the README file and to column-filters example.

gregnb:
Approved. Will merge into next release!

bzemms:
Great but, why only render the chip? what about to render it under the filters popup ?

PR #556, opened 2019-04-11
Fix the colspan value when 'expandableRows' is set and data is empty

niuware:
When the option **expandableRows** is set to true and **selectableRows** is set to false, the column span for the cell with the "noMatch" label is not correct. The correct value should be the *column count* plus 1 when setting to true either *expandableRows* or *selectableRows* options.


gregnb:
Thanks for submitting this PR! Merged

Issue #483, opened 2019-03-11
changedColumn gets the value instead of column name on onFilterChange event


blopez2010:
When the filter is changed and according to the parameter name, `changedColumn` should be the name of the column to be filtered, instead, I'm getting the value of the column to be filtered.

## Expected Behavior
`changedColumn` should get the column name to be filtered instead.

## Current Behavior
`changedColumn` is getting the value to be filtered.

## Steps to Reproduce (for bugs)
I didn't do too much to repro it, just add the event `onFilterChange` and a `console.log(changedColumn, filterList)` and that's it.

```
      onFilterChange: (changedColumn, filterList) => {
        console.log(changedColumn, filterList)
      }
```

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  ^3.9.2       |
| MUI-datatables  |  ^2.0.0-beta.56       |
| React        |   ^16.8.1      |
| browser      |   Chrome Version 72.0.3626.121      |

gregnb:
@blopez2010 Thanks for finding the issue. Can you send in a PR?

gabrielliwerant:
PR up for this issue.

gregnb:
Should be resolved upgrade to the latest version 2.0.0

cherukumilli:
@gregnb @gabrielliwerant @blopez2010 
I still see the issue in the latest version 2.0.0.

gabrielliwerant:
Yes, that makes sense, the PR has not been merged yet.

cherukumilli:
@gabrielliwerant - thx for reopening the issue. Is there a temporary workaround that you can suggest till the PR gets merged?

gabrielliwerant:
@cherukumilli Unfortunately, it's not a one-liner. The main fix is here https://github.com/gregnb/mui-datatables/pull/560/commits/8760dfc0793f4b21e96c766a0df4e6a5ac4bad50. If you had to move forward before this gets merged, you could fork and apply that fix, at your own risk, of course!

gabrielliwerant:
Closing this issue as it should have been resolved by the PR referenced here, and is a part of the release made shortly afterwards.

PR #560, opened 2019-04-12
Fix for Issue 483

gabrielliwerant:
Fixes issue https://github.com/gregnb/mui-datatables/issues/483 (`changedColumn` was showing filtered value instead of column name).

When removing filters (`TableFilterList.js`), there is no information about column names, so `null` is supplied instead. We could default to the filter value removed in this case (as it was previously), but not sure what is best.

UPDATE:

The proper column name should now always be passed along with the filter information. This was done by passing the column names along to the filter list component, and correlating them with filter position (which matches column position).


gregnb:
Alright let me think on this one maybe we can come up with a better solution to solve that null issue

gabrielliwerant:
I added a state value to pass into the `TableFilterList` in order to prevent the `null` after removing filters. Also added testing around the filtered column name in the callback in order to help prevent regressions. @gregnb

gabrielliwerant:
@gregnb Added a more elegant fix here which should solve the outstanding issues.

gregnb:
Looks great thanks @gabrielliwerant

Issue #274, opened 2018-11-15
Extends selectableRows to have 'single', 'multiple' or 'none'  options


rnugraha:
Extends selectableRows options to have 'single', 'multiple' or 'none' so that there are more options to select a row.

## Expected Behavior
In the component's options, one should be able to set

|Name|Type|Default|Description
|:--:|:-----|:--|:-----|
|**`selectableRows`**|string|''multiple'|Configure row selection. Options: 'single', 'multiple' or 'none'.

When "single" is selected: 
- It will not render checkboxes
- Rows are selectable but only single selection
- Selected row is highlighted

When "multiple" is selected: 
- It will render checkboxes
- Rows are selectable
- Selected row is highlighted

When "none" is selected: 
- It will not render checkboxes
- Rows are not selectable

## Current Behavior
There are only two options if rows are selectable or not. If it is selectable it will display checkboxes and it implies multiple selections only.

marco270:
I also agree with this behavior

Psykar:
I'd like to take this one step further, and have it accept a function.
`(rowIndexes, data) => []bool`
Where each row in the return value specifies if that row is selectable or not.

Psykar:
Actually, probably a duplicate of https://github.com/gregnb/mui-datatables/issues/274

gregnb:
> When "single" is selected:

It will not render checkboxes
Rows are selectable but only single selection
Selected row is highlighted

I'm not sure I follow? what would we show instead of a checkbox here? radio button?

simonkaspersen:
Second this. 
When "single" is selected, it would be nice with no checkboxes, only highlight selected row.

gabrielliwerant:
This issue has been addressed, as `selectableRows` now has the options listed here, except that for `'single'`, checkboxes are rendered on the rows (but only one is selectable at a time). It is important for UX reasons to present an indication that a user can select a row, hence the checkboxes showing by default.

However, selection on row click will become available via this PR, if you want to follow it: https://github.com/gregnb/mui-datatables/pull/607. When it does, you can override styles to hide the checkboxes, hence achieving the functionality for `'single'` described here.

PR #563, opened 2019-04-14
Fix for issue 274

janfranco27:
Fixes issue #274 extending the default behaviour of `selectableRows` so it supports one of `single`, `multiple` or `none`.

The checkbox in the header section will be only displayed when `selectableRows == 'multiple'`.
In the case of `single` selection, the checkboxes will be kept next to the rows, but no in the header. Also clicking any other row different than the selected one will change the selection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ The component accepts the following props:
 |**`filterType `**|string|'dropdown'|Choice of filtering view. Options are "checkbox", "dropdown", "multiselect" or "textField"
 |**`textLabels `**|object||User provided labels to localize text
 |**`pagination`**|boolean|true|Enable/disable pagination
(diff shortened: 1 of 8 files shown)


gregnb:
@janfranco27 can we add in some backward capability for the boolean? it is a breaking change. if a boolean is provided and the value is 'true' then make the value 'multiple'. If the value is 'false' then make the value 'none'

gregnb:
@janfranco27 never mind. I see that code now. my bad!

vnellore:
Hello, thank you for this timely enhancement. I am looking forward to this feature to use it in my current project as soon as possible.

gregnb:
Thanks for the PR! Will be merged into the next release

PR #564, opened 2019-04-16
Fix for Issue 258 (Do not render toolbar when all options are missing)

gabrielliwerant:
Prevents the rendering of the toolbar when the all of the toolbar display items are missing, including:

- filter
- search
- print
- download
- viewColumns
- customToolbar
- title

Also adds a test case for this scenario, as well as an example on the example page.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -13,6 +13,8 @@ import classnames from 'classnames';
 import cloneDeep from 'lodash.clonedeep';
 import merge from 'lodash.merge';
 import isEqual from 'lodash.isequal';
+import find from 'lodash.find';
+import isUndefined from 'lodash.isUndefined';
 import textLabels from './textLabels';
 import { withStyles } from '@material-ui/core/styles';
 import { buildMap, getCollatorComparator, sortCompare } from './utils';
@@ -55,6 +57,23 @@ const defaultTableStyles = {
 const TABLE_LOAD = {
   INITIAL: 1,
   UPDATE: 2,
+};
+
+// Populate this list with anything that might render in the toolbar to determine if we hide the toolbar
+const TOOLBAR_ITEMS = [
+  'title',
+  'filter',
+  'search',
+  'print',
+  'download',
+  'viewColumns',
+  'customToolbar'
+];
+
+const hasToolbarItem = (options, title) => {
+  options.title = title;
+
+  return !isUndefined(find(TOOLBAR_ITEMS, i => options[i]));
 };
 
 class MUIDataTable extends React.Component {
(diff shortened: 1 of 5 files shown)


gabrielliwerant:
The CI check needs to be run again, the failure is due to missing dependencies, which is a just a race condition issue since the CI tested the code before the dependencies were finished installing.

gregnb:
Error: Cannot find module 'lodash.isUndefined'

gabrielliwerant:
Right, they're there, the CI build just tested before it finished installing the dependencies, so it needs to be run again.

https://github.com/gregnb/mui-datatables/pull/564/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R100

gabrielliwerant:
Right, they're there, the CI build just tested before it finished installing the dependencies, so it needs to be run again.

https://github.com/gregnb/mui-datatables/pull/564/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R100

src/MUIDataTable.js
```diff
@@ -13,6 +13,8 @@ import classnames from 'classnames';
 import cloneDeep from 'lodash.clonedeep';
 import merge from 'lodash.merge';
 import isEqual from 'lodash.isequal';
+import find from 'lodash.find';
+import isUndefined from 'lodash.isUndefined';
```
gregnb: try changing this to:

import isUndefined from 'lodash.isundefined';

gabrielliwerant: Haha, that's crazy, no errors for me when running this locally, but nice catch, will fix.

gregnb:
Awesome! Merged

gregnb:
hey @gabrielliwerant would you be interested in co-maintaining this repo with me? I'm so swamped at this point the package is really starting to fall apart sadly.  There are some terrible defects that I do need to look out for a hand.

For example: https://github.com/gregnb/mui-datatables/issues/565

gabrielliwerant:
Oh, sorry, just saw this. I would be honored! I may also have a good block of time soon, so just let me know what issues are highest priority. For now, I'll start taking the issues marked as "bug" or "help wanted".

gregnb:
Awesome @gabrielliwerant :) Thanks so much

belkinev:
Thanks for fix. There is similar issue with other toolbar.
@gabrielliwerant how can we hide MUIDataTableToolbarSelect that appear when some rows are selected?
I have tried custom rendering with **customToolbarSelect** prop but it change only part with delete button.

gabrielliwerant:
@belkinev You should be able to hide it with custom theme overrides, like so:

```
const myTheme = createMuiTheme({
  overrides: {
    MUIDataTableToolbarSelect: {
      root: {
        display: 'none'
      }
    }
});

<MuiThemeProvider theme={myTheme}>
  <MUIDataTable ... />
</MuiThemeProvider>
```

PR #570, opened 2019-04-20
Issue 565 Sort Order Swapped

gabrielliwerant:
Fixes https://github.com/gregnb/mui-datatables/pull/570.

After the sort order was changed here https://github.com/gregnb/mui-datatables/pull/508/files, the ordering was swapped. We restore it by switching the values in the comparator.

Also updates the sort test and adds another one for the opposing sort direction.


gregnb:
Great! Nice Catch

PR #574, opened 2019-04-24
fix issue 514 index for rowsSelected option

### LINKED ISSUE #514: weird `allRowsSelected` state after setting `rowsSelected` and clicking checkbox in context of filter
My app is setting `options.rowsSelected` to allow user to select checkboxes `onRowClick`. However, this is leading to weird behavior in `onRowsSelect` where `allRowsSelected` will have the right `dataIndex` but the wrong [row]`index` in the context of a filter.

In the console below, you can see the first `allRowsSelected` is fine after clicking the first checkbox the first time (`index: 0`), but then after setting `rowsSelected: [32]`, the next time I click the same checkbox to disable it, you can see that the same (`dataIndex: 32`) now has two indexes (`index: 32` and `index:0`).

image

## Expected Behavior

Expected that after setting `rowsSelected` to `[32]`, clicking the checkbox again would result in `allRowsSelected` to be empty `[]` rather than having two duplicate `dataIndexes`.

1. Apply a filter (via UI)
2. Click checkbox (where `dataIndex > 0`) and in `onRowsSelect` change app state (e.g. via redux) so that a `options.rowsSelected` eventually receives all the dataIndexes reported in allRowsSelected.
3. Click same checkbox and notice duplicates in allRowsSelected, rather than empty array.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  3.9.2       |
| MUI-datatables  |   2.0.0-beta.58      |
| React        |  16.3.0       |
| browser      | Chrome    |

kliu4:
This PR fix index for rowsSelected option. When rowsSelected are added in option, { index: rowPos, dataIndex: row } are added into selectedRowsData. However, the index could not be same as dataIndex, e.g, users filter data first and set rowsSelected. The PR fix #514

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -437,7 +437,16 @@ class MUIDataTable extends React.Component {
     if (TABLE_LOAD.INITIAL) {
       if (options.rowsSelected && options.rowsSelected.length) {
         options.rowsSelected.forEach(row => {
-          selectedRowsData.data.push({ index: row, dataIndex: row });
+          let rowPos = row;
+
+          for (let cIndex = 0; cIndex < this.state.displayData.length; cIndex++) {
+            if (this.state.displayData[cIndex].dataIndex === row) {
+              rowPos = cIndex;
+              break;
+            }
+          }
+
+          selectedRowsData.data.push({ index: rowPos, dataIndex: row });
           selectedRowsData.lookup[row] = true;
         });
       }


kliu4:
@gabrielliwerant @gregnb the gif shows the `allRowsSelected` doesn't have duplicate row. I guess you used `console.log(currentRowsSelected, allRowsSelected);` currentRowsSelected is the current row selected in your case is [{index: 7, dataIndex: 7}] in your case; allRowsSelected is the all rows selected [{index: 7, dataIndex: 7}]. This is correct.

The PR fixed the issue as in `https://codesandbox.io/s/w705n98ko5`, and following gif shows the issue: 1) filter the records by select `b` in my case; 2)click `test` to set rowsSelected to [1]; 3)then click the checkbox to unselect the row and now it has the error to show to rows in allRowsSelected
issue514

The code says it is `TABLE_LOAD.INITIAL`, but in fact it checks the rowPos whenever users set the `rowsSelected`. So the PR solved the issue

gabrielliwerant:
@kliu4 You are correct, I'm sorry, I did not understand the issue properly. Thank you for clarifying and explaining. I can verify that the issue is fixed here, so this looks good to me!

kliu4:
@gabrielliwerant Thanks for your review

Issue #525, opened 2019-03-27
Expandable Rows Not Available Under 960px width


ShaggyKris:
When the window width is <=960px, the icons for expanding rows disappears.

## Expected Behavior

Rows to be expandable at all screen sizes
## Current Behavior

Expand icon does not show up when window.innerWidth <= 960px.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    ^3.9.2     |
| MUI-datatables  |     2.0.0-beta.58    |
| React        |    16.7.0     |
| browser      |    Chrome, Safari     |
| etc          |         |

PR #581, opened 2019-04-27
Allow select boxes and expandable row icons in responsive view

gabrielliwerant:
Fixes https://github.com/gregnb/mui-datatables/issues/525.

The issue looked like a bug to me, but I thought it also might have been your intent to hide expandable icons and select boxes on responsive breakpoints, so I'm curious what you think. It does take up valuable space, but it also seems like it could be confusing UX to hide the options without any way to reinstate/toggle them back on (other than a wider viewport).

548px:

image

image


gregnb:
Definitely a bug good catch. Merged!

PR #585, opened 2019-04-30
Bugfix/issue 305 data update inconsistencies

gabrielliwerant:
Fixes the remaining pagination issues in https://github.com/gregnb/mui-datatables/issues/305.

- Enhance table example to demonstrate issues with state resets
- Add testing for state reset issues related to pagination (and add ids for this purpose)
- Remove deprecated lifecycle hook
- Update only the necessary state rather than also resetting table to initial values

I know that testing via the DOM can be brittle, but I tried to make it as robust as possible by using ids. I worked for a long time trying to test it in another manner, but the bug only occurs on state change via the lifecycle hooks, so it looked like I needed to actually trigger those to test properly.


gregnb:
Totally fine man I'm not too hung up on adding id's. I'm going to merge this in

PR #586, opened 2019-04-30
Fix toolbar trash icon button height

gabrielliwerant:
Fixes improper height for icon as demonstrated when hovering (notice the icon all the way on the right).

Broken:
Capture_broken

Fixed:
Capture_fix


gregnb:
Awesome nice catch :)

Issue #557, opened 2019-04-11
Detect before row deletion to prevent rows delete on some server condition.


mageshkumar1598:
I am trying to prevent row deletion if the condition doesn't not meet. Now I can able to do that only after row deletion in onRowsDelete() method.Kindly help me with this.

gabrielliwerant:
I've got a PR up that might help. The idea is that returning false from onRowsDelete will prevent a delete action from taking place. There will be an example in the code to demonstrate this as well.

Issue #424, opened 2019-02-18
onRowsDelete 


RutulDarji:
onRowsDelete: (data) => { console.log( data )  },
i have done this.
i want to customize that and don't allow to delete rows of that table 
How can i do that??

or 
if i select any rows i want to disable the delete button ??
i tried with customize toolbar still its not working

sbtgE:
It's too late to make any changes in onRowsDelete method. You can solve that using custom ToolbarSelect like this:

```
options = {
   customToolbarSelect: (selectedRows, displayData, setSelectedRows) => (
      <React.Fragment/>
   ),
}
```

React.Fragment will not add unnecessary elements to your DOM.

RutulDarji:
why it is too late ?

sbtgE:
@RutulDarji  because your rows already disappeared from the table didn't they? onRowsDelete is callback after delete actually. Isn't solution with customToolbarSelect suitable for your case?

RutulDarji:
I am working on local machine so delete rows not affect that much

RutulDarji:
customToolbarSelect works for me but what happen after i completed my task i have clear all my states but the previously selected times like .. i have selected 2 rows and it shows on header that 2 rows selected after the completion of the task i have cleared all my states so in react it will re render but the 2 rows selected still stay at table header

RutulDarji:
customToolbarSelect works for me but what happen after i completed my task i have clear all my states but the previously selected times like .. i have selected 2 rows and it shows on header that 2 rows selected after the completion of the task i have cleared all my states so in react it will re render but the 2 rows selected still stay at table header..

sbtgE:
There is another option `rowsSelected` - you can map it to your state and according to your words clear that as well

matteo1976:
Sorry, I don't understand.
Example:
columns[id,value] 
data are from external API and it will be something like this:
`{id:123, value:value1}{id:678, value:value2}` ecc... 
Imagine you have many rows ...
With onRowsDelete I have back only the index of the table and the array of selection....
`onRowsDelete:(rows)=>{console.log(rows)}`
I have in console 
    data: Array[1]
     0: Object
     index: 1
     dataIndex: 1

but i need to have my id not the index of the row of the table.... 
suppose I selected my first row
I'd expect something: 
    data: Array[1]
     0: Object
     index: 1
     dataIndex: 1
     id:123
     value:value1
so that I can remotely delete the record with the selected id

or alternatively
[...]

Issue #193, opened 2018-09-15
How to control deletion of rows in table?


bonjefir:
I need to cancel deletion of some of row(s) which user deleted but its operation failed in back-end side for some reason.

```
const options = {
	filter: true,
	filterType: 'dropdown',
	responsive: 'scroll',
	onRowsDelete: (rowsDeleted) => {
		for (var key in rowsDeleted.data) {
			this.removeItem(this.state.item_id[rowsDeleted.data[key].dataIndex])
				.then(res => {
					if (res != true) {
						// ???
					}
			})
		}
		console.log(rowsDeleted, "were deleted!");
	}
};
```
Here `removeItem` returns a promise which shows the status of deletion. I am wondering deletion in UI applied after calling `onRowsDelete` or before it and how can recover deleted rows OR preventing MUI to delete items in its `rowsDeleted` which their `res` is `false`.

gregnb:
Hey @bonjefir thanks for filing the issue. Let me think about this some more. 

One change that will soon happen is the rowDelete action will be added to the onServerRequest callback so you can look for that event there and react to it. If you do receive the event there, maybe if there is a total failure you can just return the data for that page back to what it was from your API?

majose1500:
Sorry but the onRowsDelete options not work to me. I'm trying to do a basic example like show message when a row was delete but do not see nothing. Exist any functional example of onRowsDelete options?

`const options = {
	filterType: 'dropdown',
	responsive: 'stacked',
	onRowsDelete: (rowsDeleted) => {
		console.log(rowsDeleted, "were deleted!");
	}
};`

toolaugh:
Normaly, If we want to delete a row of table, We have to pass a ID of the data Row and send it to the server. But, The data that we use to build the table does not have the ID of the row data. So we have to map the selected row with the data from server that we want to delete.

d-gubert:
> Normaly, If we want to delete a row of table, We have to pass a ID of the data Row and send it to the server. But, The data that we use to build the table does not have the ID of the row data. So we have to map the selected row with the data from server that we want to delete.

To solve this you could add the ID as the first column, and not actually show the column, like this:
```javascript
columns = [
    {
        name: "ID",
        options: {
            display: "excluded",
            filter: false,
            sort: false,
            download: false,
        },
    },
    "Pacote",
    "Validade",
    "Data de Criação",
    "Ativo",
];
```
[...]

PR #587, opened 2019-05-01
Enhancement/issue 193 424 skip row delete

gabrielliwerant:
A lot of people have been asking for the ability to prevent deletion of certain rows via the selection toolbar, so I thought I'd give it a shot. Should close https://github.com/gregnb/mui-datatables/issues/424 and https://github.com/gregnb/mui-datatables/issues/193 but possibly others as well.

The idea is to allow the `onRowsDelete` function to return `false`, which will then skip the deletion. I added some tests and sample functionality around this, as well as updating the docs accordingly.

Let me know what you think!

Also closes https://github.com/gregnb/mui-datatables/issues/557

UPDATE:

Also added documentation to the README around `onRowsDelete`, as it was out of date, but I'm open to suggestions on ways to make it more clear, since the object entering the function is complex.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -831,6 +831,7 @@ class MUIDataTable extends React.Component {
     const cleanRows = data.filter(({ index }) => !selectedMap[index]);
 
     if (this.options.onRowsDelete) {
+      if (this.options.onRowsDelete(selectedRows) === false) return;
       this.options.onRowsDelete(selectedRows);
     }
 
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ The component accepts the following props:
 |**`downloadOptions`**|object||Options to change the output of the CSV file. Default options: `{filename: 'tableDownload.csv', separator: ','}`
 |**`viewColumns`**|boolean|true|Show/hide viewColumns icon from toolbar
 |**`onRowsSelect`**|function||Callback function that triggers when row(s) are selected. `function(currentRowsSelected: array, allRowsSelected: array) => void`
-|**`onRowsDelete`**|function||Callback function that triggers when row(s) are deleted. `function(rowsDeleted: array) => void OR false` (Returning `false` prevents row deletion.)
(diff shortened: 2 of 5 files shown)


matteo1976:
I think it would be a good feature if the onRowDelete function will return an array of objects that where the objects will contain all the data of the rows...(one object for each row deleted) with a simple check on the length of this array would be exactly like having the false ... and at the same time it could be useful to intercept the deleted data and use them for example to delete them in a remote db...
If there are the request to have a control for prevent deletion , I think, it would be better have a function like isRorwsDeleted but onRowDeleted is more like an event to menage.
this is my modest advice

gabrielliwerant:
@matteo1976 My idea is to keep the functions as simple as they need to be for people to use them for their own purposes. In the case of the `onRowsDelete` callback, you actually should have all of the data you need to do whatever you want. The function allows you to look at the dataIndex of any selected row about to be deleted, which is enough to find those rows in your state and then do whatever you want to them.

gregnb:
@gabrielliwerant Will review when I get some time. At quick glance we're changing the returning data structure right? if we merge should probably go with a minor bump for release

gabrielliwerant:
The data structure is the same, but the docs were out of date, and now the user can optionally return `false`. I think a bump to the minor version makes sense.

gregnb:
@gabrielliwerant Going to tag this one LGTM. When we do another release this week we'll bump the minor version.

src/MUIDataTable.js
```diff
@@ -831,6 +831,7 @@ class MUIDataTable extends React.Component {
     const cleanRows = data.filter(({ index }) => !selectedMap[index]);
 
     if (this.options.onRowsDelete) {
+      if (this.options.onRowsDelete(selectedRows) === false) return;
       this.options.onRowsDelete(selectedRows);
```
mschwebl: This line should have been removed. Now when onRowsDelete doesn't return false, it is called again immediately.

gabrielliwerant: You are correct, nice catch! I will push a fix.

gabrielliwerant: https://github.com/gregnb/mui-datatables/pull/631

PR #589, opened 2019-05-02
customRowRender

Skn0tt:
I have a use-case in my project where I need a non-column based layout for smaller screens. This is to prevent horizontal scrolling, which is not a very good UX on mobile.
I still wanted to make use of the (great) filtering and sorting system of mui-datatables.

This PR adds a new `customRowRender` option to mui-datatables, that can be used like this:

```
customRowRender: ({ data }, rowIndex) => (
  <div>
    My Important data field: {data[0]}
  </div>
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The component accepts the following props:
 |**`customToolbar`**|function||Render a custom toolbar
 |**`customToolbarSelect`**|function||Render a custom selected rows toolbar. `function(selectedRows, displayData, setSelectedRows) => void`
 |**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage) => string`&#124;` React Component`
-|**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender({ data, dataIndex }, rowIndex) => React Component`
+|**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender(data, dataIndex, rowIndex) => React Component`
 |**`customSort`**|function||Override default sorting with custom function. `function(data: array, colIndex: number, order: string) => array`
 |**`customSearch `**|function||Override default search with custom function. `customSearch(searchQuery: string, currentRow: array, columns: array) => boolean`
 |**`elevation`**|number|4|Shadow depth applied to Paper component
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ The component accepts the following props:
 |**`customToolbar`**|function||Render a custom toolbar
 |**`customToolbarSelect`**|function||Render a custom selected rows toolbar. `function(selectedRows, displayData, setSelectedRows) => void`
 |**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage) => string`&#124;` React Component`
+|**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender({ data, dataIndex }, rowIndex) => React Component`
 |**`customSort`**|function||Override default sorting with custom function. `function(data: array, colIndex: number, order: string) => array`
 |**`customSearch `**|function||Override default search with custom function. `customSearch(searchQuery: string, currentRow: array, columns: array) => boolean`
 |**`elevation`**|number|4|Shadow depth applied to Paper component
--- a/examples/customize-rows/index.js
+++ b/examples/customize-rows/index.js
@@ -6,7 +6,7 @@
 
 import React from 'react';
 import ReactDOM from "react-dom";
-import MuiDatatable from "../../src/";
+import MuiDataTable from "../../src/";
 
 function YourCustomRowComponent(props) {
   const { name, cardNumber, cvc, expiry } = props;
@@ -48,7 +48,7 @@ const creditCards = [
 
 function Example() {
   return (
-    <MuiDatatable
+    <MuiDataTable
       title="Cards"
       data={creditCards}
       columns={[
--- a/examples/customize-rows/index.js
+++ b/examples/customize-rows/index.js
@@ -5,7 +5,8 @@
 */
 
 import React from 'react';
-import MuiDatatable from "mui-datatables";
+import ReactDOM from "react-dom";
+import MuiDatatable from "../../src/";
 
 function YourCustomRowComponent(props) {
   const { name, cardNumber, cvc, expiry } = props;
--- a/examples/customize-rows/index.js
+++ b/examples/customize-rows/index.js
@@ -1,0 +1,95 @@
+/*
+  See another example of how to use `customRowRender` at
+  https://github.com/Skn0tt/mui-datatables-responsive-demo
+  https://mui-datatables-responsive-demo.skn0tt.now.sh
+*/
+
+import React from 'react';
+import MuiDatatable from "mui-datatables";
+
+function YourCustomRowComponent(props) {
+  const { name, cardNumber, cvc, expiry } = props;
+
+  return (
+    <div>
+      <h1>
+        {name}
+      </h1>
+      <p>
+        Number: {cardNumber} <br/>
+        CVC: {cvc} <br/>
+        expiry: {expiry}
+      </p>
+    </div>
+  );
+}
+
+const creditCards = [
+  {
+    name: "Tom Tallis",
+    cardNumber: "5500005555555559",
+    cvc: "582",
+    expiry: "02/24"
+  },
+  {
+    name: "Rich Harris",
+    cardNumber: "4444444444444448",
+    cvc: "172",
+    expiry: "03/22"
+  },
+  {
+    name: "Moby Dixon",
+    cardNumber: "3566003566003566",
+    cvc: "230",
+    expiry: "12/25"
+  }
+];
+
+function Example() {
+  return (
+    <MuiDatatable
+      title="Cards"
+      data={creditCards}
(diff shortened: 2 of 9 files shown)


gabrielliwerant:
@Skn0tt Have you had a look at the `responsize: 'stacked'` option that provides a non-column based layout already?

Skn0tt:
I did, but it didn't match the asthethics I was looking for.

gregnb:
@gabrielliwerant so I don't think I'm fully against adding this maybe it could be useful to someone? What are your thoughts

gabrielliwerant:
It's interesting. I'm on the fence.

I played around with it a little bit, and it seems difficult to maintain a consistent experience with the other table options like column hiding, selecting rows, filtering, etc. It makes it easy to miss implementing data that belongs in the table, which will impact what happens when users use other toolbar options.

If it's used to completely change the nature of a data table, then it seems like it's out of the scope of a data table library. Should it be possible to completely change the user experience? Perhaps. Should it be easy? I'm less sure about that one.

On the other hand, there are already a lot custom rendering options, so what's the harm in one more? I guess there's something that seems more fundamental about rows to me.

What would be convincing to me is an example that demonstrates usage of this feature to do something table-like that isn't possible yet in some other way. Perhaps the responsive issue alluded to earlier would make a good example.

Another way to go about solving the problem might be by providing additional CSS classes so the theme can be overridden even more easily to provide different experiences/skins.

Skn0tt:
> What would be convincing to me is an example that demonstrates usage of this feature to do something table-like that isn't possible yet in some other way. Perhaps the responsive issue alluded to earlier would make a good example. @Skn0tt 

I implemented a small example so you can see what I mean.
You can find the code and an explanation why I think this is useful at [skn0tt/mui-datatables-responsive-example](https://github.com/Skn0tt/mui-datatables-responsive-demo).
You can see the example online at <https://mui-datatables-responsive-demo.skn0tt.now.sh>.
I hope you like it! 😄

gabrielliwerant:
@Skn0tt Thank you so much for creating the example! That is pretty cool, I can see how that would be useful. Would you be able to add an example to the PR?

rollup.config.js
```diff
@@ -35,7 +35,7 @@ export default {
     }),
     uglify({
       compress: {
-        warnings: false,
+        // warnings: false,
```
gabrielliwerant: Is there a reason this line needs to commented out, or was this from debugging?

Skn0tt: I needed to disable it in order to have my local tests running, but I just reverted that change in https://github.com/gregnb/mui-datatables/pull/589/commits/09917f9b98ef47da6f3a5aea513cafe5194e1c9c.

Skn0tt:
I just added the example. I will make sure to extend the typescript definitions at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mui-datatables/index.d.ts) once this is merged.

README.md
```diff
@@ -147,6 +147,7 @@ The component accepts the following props:
 |**`customToolbar`**|function||Render a custom toolbar
 |**`customToolbarSelect`**|function||Render a custom selected rows toolbar. `function(selectedRows, displayData, setSelectedRows) => void`
 |**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage) => string`&#124;` React Component`
+|**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender({ data, dataIndex }, rowIndex) => React Component`
```
gabrielliwerant: To keep the argument structure similar to `customBodyRender`, how about using the arguments all separately, so `data, dataIndex, rowIndex`, rather than `{ data, dataIndex }, rowIndex`?

examples/customize-rows/index.js
```diff
@@ -0,0 +1,95 @@
+/*
+  See another example of how to use `customRowRender` at
+  https://github.com/Skn0tt/mui-datatables-responsive-demo
+  https://mui-datatables-responsive-demo.skn0tt.now.sh
+*/
+
+import React from 'react';
+import MuiDatatable from "mui-datatables";
```
gabrielliwerant: Missing a few things here. This line should be `import MUIDataTable from "../../src/";` and `MUIDataTable` should be used below instead of `MuiDatatable`. Also, we'll need the `import ReactDOM from "react-dom";` dependency as well.

test/MUIDataTableBody.test.js
```diff
@@ -263,4 +263,31 @@ describe('<TableBody />', function() {
     assert.isAtLeast(options.setRowProps.callCount, 1);
     assert(options.setRowProps.calledWith(data[1]));
   });
+
+  it("should use 'customRowRender' when provided", () => {
+    const options = { customRowRender: () => <div>Test_Text</div> };
+    const selectRowUpdate = stub();
+    const toggleExpandRow = () => {};
+
+    const t = mount(
```
gabrielliwerant: I believe elsewhere, we mostly use the name `fullWrapper` for a `mount` wrapped component, so it would be good to keep with that convention for clarity.

README.md
Skn0tt: Changed in 5744a08740562cf5dc4d0a7de75b2cbf22c0c864.

examples/customize-rows/index.js
Skn0tt: Well, I should have seen that myself 😅
I guess I'm used to Typescript too much...

Fixed in 925366855e065ce211d24f0d9f11b1842b46a622.

test/MUIDataTableBody.test.js
Skn0tt: I looked around in the repo and found that you use `fullWrapper` when you are mounting the "full" `<MuiDataTable/>`, but `mountWrapper` when you are wrapping only some sub-component. So I renamed `t` to `mountWrapper`, also in some other tests I found using `t` as a variable name: 26fdf818b249dc66623d8439548dd9483d4de1b1.

examples/customize-rows/index.js
```diff
@@ -0,0 +1,96 @@
...
+        },
+        {
+          name: "expiry",
+          label: "Expiry"
+        },
+      ]}
+      options={{
+        selectableRows: "none",
+        responsive: "scroll",
+        customRowRender: ({ data }) => {
```
gabrielliwerant: With the changes to the arguments, this has fallen out of date, sorry!

examples/customize-rows/index.js
```diff
@@ -0,0 +1,96 @@
+/*
+  See another example of how to use `customRowRender` at
+  https://github.com/Skn0tt/mui-datatables-responsive-demo
+  https://mui-datatables-responsive-demo.skn0tt.now.sh
+*/
+
+import React from 'react';
+import ReactDOM from "react-dom";
+import MuiDatatable from "../../src/";
```
gabrielliwerant: For consistency, we should keep the case the same as it is elsewhere, so it would `MuiDataTable` instead of `MuiDatatable`.

examples/customize-rows/index.js
Skn0tt: Fixed in https://github.com/gregnb/mui-datatables/pull/589/commits/8e1724c7bc3e0eafb92792e6ed1230cdaec899c8

examples/customize-rows/index.js
Skn0tt: Fixed in https://github.com/gregnb/mui-datatables/pull/589/commits/c57a81c6cc47652231b1ef20294147c5c221e460.

gabrielliwerant:
This looks good to me, thanks for your hard work on this! Pending final approval from @gregnb.

shawnmitchell:
This is very cool, but I'm having some updating issues

first screen shot below you can see a `customRowRender` is defined and the toggle switch to expand descriptions is checked
image

but when I toggle the switch to off, dev-tools tells me there is no `customRowRender` defined for the table, but it does not rerender with the default row layout. 
image

I've checked the example (the credit card example is super cool) and I think I'm handling the switch correctly. 

Any thoughts on what might be going on?

Skn0tt:
If you look closely at the example, you'll see that the key of the DataTable is changed depending on wether `customRowRender` is used or not. This forces MuiDataTable to remount everytime `customRowRender` is changed. That's the mechanism I use myself for ensuring the updates are coming through, but this has a downside: On resizing, the table forgets all its state. For me, this works, but I can see how this is suboptimal (at best) for other usecases.

shawnmitchell:
Thank you, @Skn0tt  for the *immediate* response! And thanks for the contribution. It's really cool!

Skn0tt:
You're welcome. Is better updating something you'd be interested in? I can have a look at it tomorrow and see what I can do.

shawnmitchell:
I already keep most of the table state in the parent component / localStorage so that settings are sticky.

shawnmitchell:
@Skn0tt one thing that would be cool however would be to inject a flag into tableState so if you're using `customBodyRender` for a column in default row rendering mode, you can return the data in a different format for `customRowRender`. 

In my case, I'm using `customRowRender` to show a card style layout if the user wants to expand the description field in the table. If they don't want it expanded, I use `customBodyRender` to render a clickable more info icon that displays the description in a modal. But if I'm using `customRowRender` I want to just return the raw description. 

`customBodyRender` gets passed the tableState so if I knew from there that I was customRowRender'ing, I'd just return the raw text. 

Does it make sense?

PR #594, opened 2019-05-05
[customFilterListRender] custom render also filter options in toolbar filter lists

bzemms:
#407

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ const columns = [
 |**`empty`**|boolean|false|This denotes whether the column has data or not (for use with intentionally empty columns)
 |**`filterList`**|array||Filter value list [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
 |**`filterOptions`**|array||Filter options [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
-|**`customFilterListRender`**|function||Function that returns a string used as the chip label. `function(value) => string` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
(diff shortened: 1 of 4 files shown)


bzemms:
@gregnb please review

gabrielliwerant:
Hello, I'm trying to understand this use case a bit better. Can you provide a screenshot or example that demonstrates what you're going for in the UI?

bzemms:
> Hello, I'm trying to understand this use case a bit better. Can you provide a screenshot or example that demonstrates what you're going for in the UI?

Sure, in case that you need to customize also the options in the filter popover... and not only the filter chip

image

gabrielliwerant:
Thank you, will review when I get a chance.

gabrielliwerant:
So, I think what you were looking for here is a way to have a custom rendering of the filters, is that correct?

If so, there is another implementation here which uses the new `filterOptions` API: https://github.com/gregnb/mui-datatables/pull/708. If that is what you need, then we should probably close this one. I'm sorry about that. Had we continued from here, I would have asked for an example and similar API additions as in the new PR, so it just happened to come up before I got around to reviewing this one.

gabrielliwerant:
@bzemms Custom filter rendering is now available in `2.6.0` via `filterOptions: { display: Component }` and `filterType: 'custom'` in column options. I believe this satisfies the use case here, correct?

bzemms:
> @bzemms Custom filter rendering is now available in `2.6.0` via `filterOptions: { display: Component }` and `filterType: 'custom'` in column options. I believe this satisfies the use case here, correct?

In fact its not the same thing, in my case I want to render each text of the checkbox filter , in #708 there is no filterData, and the custom render is only on static fields no matter what are the data in the table

gabrielliwerant:
@bzemms You can do whatever you like inside the display option, including looping over all your data for a given column, and displaying checkboxes for each.

gabrielliwerant:
Closing, as I believe everything here is now possible via filterOptions.

PR #597, opened 2019-05-07
Suggestion to fix critical issue, table is not updated.

bzemms:
seem you set table with old (`prevProps`) data instead of `this.props` (related to #585)
TL;DR;
@gabrielliwerant @gregnb I'm not sure I agree with this approach to use `componentDidUpdate` instead of `componentWillReceiveProps` but I got issue that state is not updating.

work fine now..

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -190,7 +190,7 @@ class MUIDataTable extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
-      this.setTableData(prevProps, TABLE_LOAD.INITIAL);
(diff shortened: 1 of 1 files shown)


gabrielliwerant:
@bzemms `componentWillReceiveProps` is being deprecated by React, so it is best to move away from it if we want to have an upgrade path going forward.

However, I think you are right that it should be `this.props`. Would you mind presenting a failing example so that I can determine why the tests passed and possibly write a new test?

gabrielliwerant:
Ah, it breaks `serverSide: true`. I can work on a test, but in the mean time, I'm merging this, as I agree it's a critical patch, and I created the regression.

gabrielliwerant:
@bzemms P.S. Thank you so much for catching that!

PR #602, opened 2019-05-09
Allow the toolbar to be rendered when a row is selected

InnuceEAN:
When selecting a column the toolbar is set too hidden. This adds a new option "alwaysRenderToolbar" to allow the toolbar to be visible even after the selection of a column.

Default value is false so the original behaviour is not changed.


InnuceEAN:
Hi,

any feedback on this?

gabrielliwerant:
Version `2.13.0` has `disableToolbarSelect` which allows you to remove the select toolbar, which reveals the regular toolbar.

InnuceEAN:
This is not quite the same if disableToolbarSelect looses the select toolbar..

gabrielliwerant:
Agreed. I will review this if you can resolve the conflicts, and also create a more robust API around the available options. I'm thinking something like

```
selectToolbarVisibilityOptions: {
  disabled: true | false (acts like the current option, and will not render the toolbar at all)
  enabledButHidesMainToolbar: true | false (acts like the option you want, where it doesn't replace the main toolbar, but default will act as it does in the table now, where it hides the main toolbar)
}
```

What do you think? I'm open to pithier naming, but I'm trying to err on the side of clarity.

patorjk:
This doesn't seem like a good idea - why would you want a table with 2 toolbars visible? However, if this were to be done I would recommend an enum option. Options parameters lend themselves to being confusing and also have weird edge cases. For example, what happens if both selectToolbarVisibilityOptions.disabled and selectToolbarVisibilityOptions.enabledButHidesMainToolbar are true?

Closest thing to this type of API I can think of is the "placement" paramater that some Material UI components have. If you were to take a cue from that, something like this may make sense:

    selectToolbarPlacement: ["on-toolbar", "under-toolbar", "none"]

InnuceEAN:
I like the enum option, but both would break the existing option disableToolbarSelect in the API, I hesitate a bit to do that. Also I would rename "on-toolbar" to "replace".

patorjk:
disableToolbarSelect would have to be deprecated and selectToolbarPlacement="none" would replace it.

InnuceEAN:
ok, I have pushed the new option selectToolbarPlacement with options ['replace', 'above-toolbar', 'none'] and deprecated disableToolbarSelect.

InnuceEAN:
Any feedback to the new option?

README.md
```diff
@@ -198,6 +198,7 @@ The component accepts the following props:
 |**`onTableInit`**|function||Callback function that triggers when table state has been initialized. `function(action: string, tableState: object) => void`
 |**`setRowProps`**|function||Is called for each row and allows you to return custom props for this row based on its data. `function(row: array, dataIndex: number) => object` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-styling/index.js)
 |**`setTableProps`**|function||Is called for the table and allows you to return custom props for the table based on its data. `function() => object` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-styling/index.js)
+|**`selectToolbarPlacement`**|string|'replace'|Controls the visibility of the Select Toolbar, options are 'replace', 'above-toolbar' and 'none'
```
gabrielliwerant: Can we add a short description here for each of the options?

src/MUIDataTable.js
```diff
@@ -200,6 +200,7 @@ class MUIDataTable extends React.Component {
       onDownload: PropTypes.func,
       setTableProps: PropTypes.func,
       setRowProps: PropTypes.func,
+      selectToolbarPlacement: PropTypes.string,
```
gabrielliwerant: Let's use the `oneOf` for the prop types definition, as we do for other "enums."

src/MUIDataTable.js
```diff
@@ -328,6 +335,7 @@ class MUIDataTable extends React.Component {
       separator: ',',
     },
     setTableProps: () => ({}),
+    selectToolbarPlacement: 'replace',
```
gabrielliwerant: I think we should make the options a bit more consistent in their naming, so either `replace` and `above` or `replace-toolbar` and `above-toolbar`.

src/MUIDataTable.js
```diff
@@ -358,6 +366,12 @@ class MUIDataTable extends React.Component {
         );
       }
     });
+
+    if (this.options.disableToolbarSelect === true) {
```
gabrielliwerant: We should also add a deprecation notice for disableToolbarSelect === false.

src/MUIDataTable.js
```diff
@@ -1360,15 +1374,16 @@ class MUIDataTable extends React.Component {
...
           <TableToolbarSelect
             options={this.options}
             selectedRows={selectedRows}
             onRowsDelete={this.selectRowDelete}
             displayData={displayData}
             selectRowUpdate={this.selectRowUpdate}
           />
-        ) : (
+        )}
+        {(selectedRows.data.length === 0 || ['above-toolbar', 'none'].includes(this.options.selectToolbarPlacement)) &&
```
gabrielliwerant: It would be better to create a constant with this array of options, and place it toward the top of the file. That way, it's easier to alter the names in a single location without forgetting anything, and it will document better what these options are.

src/MUIDataTable.js
InnuceEAN: getDefaultOptions() will set disableToolbarSelect before handleOptionDeprecation() is called. This means that the deprecation warning would always be printed, even if the option is not set by the user.

I would like to avoid the rewrite necessary for that.

InnuceEAN:
I have done all but the deprecation change (see comment). But to be honest I'm not a fan of the array with options.

src/MUIDataTable.js
gabrielliwerant: That's a good point, but it should be done regardless. However, I can do that separately, so no worries on your end.

gabrielliwerant:
Hey @InnuceEAN, thanks for working on the changes. Can you elaborate on your statement, "I'm not a fan of the array with options." Which aspect are you referring to, and do you have an alternative in mind?

InnuceEAN:
I was referring to SELECT_TOOLBAR_PLACEMENT_OPTIONS , I think it bloats the code a bit, but thats not an issue for me.

InnuceEAN:
Conflicts resolved again. Any other issues I have not handled?

InnuceEAN:
Any other issues?

patorjk:
@InnuceEAN I'm currently working on [v3](https://github.com/gregnb/mui-datatables/pull/1300). Are you still interested in this? 

I tried it out [here](https://codesandbox.io/s/admiring-faraday-qmi6d?file=/examples/customize-toolbar/index.js) and it appears to work fine. I have my reservations, since I don't like how the table shifts when an item is selected. However, since that behavior isn't the default and this feature request seems to come up every so often, I'm fine putting it in if you're interested in it.

InnuceEAN:
Sure, I'm still interested.

PR #603, opened 2019-05-09
Add searchText option to initialise datatable with search text

karthick-vinod:


karthick-vinod:
@gabrielliwerant added `searchText` to `customize-search` example as requested.

gabrielliwerant:
@karthick-vinod Excellent, looks good to go into next release!

gabrielliwerant:
Will close https://github.com/gregnb/mui-datatables/pull/441.

ericpyle:
@gabrielliwerant how do I _clear_ the searchText _AFTER_ it's been set? Setting it to an empty string, null, or undefined does not seem to restore back the full (unsearched) list.

ericpyle:
@gabrielliwerant also if I initialize `searchText` and then click the search icon in mui-datatables, the searchText string does not show up in the search text box, even though the table results do reflect `searchText` results

gabrielliwerant:
@ericpyle There were a couple of issues with this feature, which I have addressed here https://github.com/gregnb/mui-datatables/pull/647. Thank you for pointing some of them out.

After those fixes make their way to the next release, you should be able to update `searchText` dynamically and pass it an empty string to reset it.

As far as clicking the search icon, I think you mean closing the search bar, then opening again, as this is the only way you end up in the state where clicking again leaves nothing in the search box. I think that should be the default functionality, as the `searchText` option is mostly meant for initialization. If you want to add it back after opening up the search bar again, you can do so with the `onSearchOpen` option once the fixes are merged in.

ericpyle:
@gabrielliwerant thanks for fixing those! I believe the searchText behavior I need is that if searchText it's used to initialize the table, then the table will initially show the search text box as being open with the searchText as its text value. Otherwise the table would not look like it's in the right state.

gabrielliwerant:
@ericpyle That is how `searchText` is supposed to work, but it was not doing this in the case where there was also a custom `page` option.

Issue #516, opened 2019-03-24
Expand row onRowClick


aliaadil:
## Expected Behavior

Would like to expand the row when anywhere on the row is clicked not just the arrow on the left that appears when using renderExpandableRow

## Current Behavior

When using the renderExpandableRow option only the arrow appears in the left and have to click on it to expand the row

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.9.2     |
| MUI-datatables  |   2.0.0-beta.59     |
| React        |   16.8.5     |
| browser      |   Chrome     |

georgezee:
I needed this as well. For now, I've added this feature by forking the repo and adding this commit: 
https://github.com/georgezee/mui-datatables/commit/7938d386e9a7bdb46fc759d1d91078b4be6ca503

This would be considered a breaking change if it became the default for everyone, so we could add a new boolean option called "expandRowOnClick", or text option "rowClickBehaviour".

@gregnb - if you think either of the above options would be a good idea, I could raise a PR with the change. You can see how we use this at codeschooldirectory.co.za.

iamjoeker:
I think an option to enable this behavior would be great!

gabrielliwerant:
I like the idea of an option, since it doesn't prescribe specific onClick behavior. Maybe boolean for now, pivoting to an enum if other options are added (I'm sensing another use case here for row selection).

If you submit a PR, I'll be happy to review it!

georgezee:
PR: https://github.com/gregnb/mui-datatables/pull/607

You're welcome to change the option name if you'd prefer something else, there are several possibilities.

rodriigovieira:
Is there a way to already implement this with some workaround? Or somehow a manual way to make the onRowClick trigger the toggleExpandRow?

I'd adore implementing this as quick as possible.

georgezee:
@rodriigovieira - I couldn't see a way to trigger it just using the existing onRowClick. If you can't wait until the PR is reviewed and merged, you could point to the branch on my fork in the meantime, i.e. update the mui-datatables reference in your package.json to:
`"mui-datatables": "git+https://www.github.com:georgezee/mui-datatables.git#1621d52",`
though keep in mind that this doesn't have all the latest commits from the main repo.

If you do use this, you'll need to add the `fullRowTrigger` property to the table options, set to True.

rodriigovieira:
I've put the line as you wrote it in my dependencies and got this error when executing `yarn install`:

```
[...]

PR #607, opened 2019-05-10
 Add fullRowTrigger option to allow action anywhere on row click

georgezee:
The select/expand actions will be triggered by a click anywhere on the row, instead of just the regular checkbox (select)  or arrow icons (expand).

Expand example:
expand-row-click-anywhere

If both expand and select features are enabled, then both actions are triggered, though I feel this scenario is uncommon.


georgezee:
From issue #516 . 

I didn't see any current tests for expandable rows, so didn't add tests for this, but I'm open to suggestions.

README.md
```diff
@@ -141,9 +141,10 @@ The component accepts the following props:
 |**`pagination`**|boolean|true|Enable/disable pagination
 |**`selectableRows`**|string|'multiple'|Numbers of rows that can be selected. Options are "multiple", "single", "none".
 |**`isRowSelectable`**|function||Enable/disable selection on certain rows with custom function. Returns true if not provided.  `function(dataIndex) => bool`
-|**`resizableColumns`**|boolean|false|Enable/disable resizable columns
 |**`expandableRows`**|boolean|false|Enable/disable expandable rows
-|**`renderExpandableRow `**|function||Render expandable row. `function(rowData, rowMeta) => React Component`
+|**`renderExpandableRow`**|function||Render expandable row. `function(rowData, rowMeta) => React Component`
+|**`fullRowTrigger`**|boolean|false|Enable/disable trigger of select/expand when row is clicked. When False, only specific elements will trigger these actions.
```
gabrielliwerant: To make it clear that this option works in tandem with `expandableRow`, I think we call this `expandableRowOnClick` instead, as I think you suggested originally.

src/components/TableBody.js
```diff
@@ -105,10 +105,23 @@ class TableBody extends React.Component {
   };
 
   handleRowClick = (row, data, event) => {
-    // don't trigger onRowClick if the event was actually a row selection
+    // Don't trigger onRowClick if the event was actually a row selection
     if (event.target.id && event.target.id.startsWith('MUIDataTableSelectCell')) {
       return;
     }
+
+    // Check if we should trigger row expand/select when row is clicked anywhere
```
gabrielliwerant: Triggering both select and expand at the same time is UX I believe we should avoid. What I recommend is adding a second option for `selectableRows` called `selectableRowsOnClick` that will provide this, similar to `expandableRows`, so that the options are:

- `expandableRowsOnClick: true, selectableRowsOnClick: false`: Only expandable rows are opened on click
- `expandableRowsOnClick: false, selectableRowsOnClick: true`: Only selectable rows are selected on click
- `expandableRowsOnClick: false, selectableRowsOnClick: false`: Default behavior
- `expandableRowsOnClick: true, selectableRowsOnClick: true`: Both expandable rows and selectable rows are activated on click, provided that those options are enabled.

georgezee:
@gabrielliwerant - the changes you suggest make sense. I will rename the variables and add the cursor style on Friday, when I am next able to work on this. I don't mind if anyone else wants to proceed with this before then.

If you could add the testing component for this separately, that would be appreciated.

georgezee:
@gabrielliwerant - I've split the previous option into separate options for `expandableRowsOnClick` and `selectableRowsOnClick` that work as their name implies.

I added the `cursor: pointer` style to a new `hoverCursor` property which is only set when either of the previous two options are true. I'm not sure if it would be preferable to merge the style into the existing hover theme property, but I thought it cleaner the way it is now presented.

src/components/TableBody.js
```diff
@@ -105,10 +105,22 @@ class TableBody extends React.Component {
   };
 
   handleRowClick = (row, data, event) => {
-    // don't trigger onRowClick if the event was actually a row selection
```
gabrielliwerant: We can fix the edge case by adding

```
    let isExpandableIconClick = false;

    // We need to know if our row click was actually on our expandable icon
    if (event.target.nodeName === 'path' && event.target.parentNode.id === 'expandable-button') isExpandableIconClick = true;
    else if (event.target.id === 'expandable-button') isExpandableIconClick = true;
```

to the top of the `handleRowClick` function and `&& !isExpandableIconClick` to the `selectableRowsOnClick` conditional (line 114).

gabrielliwerant: Then in `TableSelectCell`, we need to add `id="expandable-button"` to the `KeyboardArrowRight` icon.

georgezee: @gabrielliwerant - good catch on the edge case.
I implemented the change as you specified in commit 511dc7c.

However looking at the rest of the `handleRowClick` function, we specifically skip calling `onRowClick` if the selectable icon is clicked, so I think it would be consistent to do the same here with the expandable icon. It also feels cleaner, and so I've added a second commit with this change (65dbadd).
This may be considered a breaking change though (expandable icon no longer triggering onRowClick), so I'll leave the decision to you, and would be happy to revert commit 65dbadd if you prefer your original suggestion.

gabrielliwerant: I like your change here, but I agree it could break code that relied on the click event previously. How about we keep your code, but instead of skipping the functionality, we add a deprecation warning via console and add back the `isExpandableIconClick ` bits? Then in a future release, we can break the backwards compatibility.

georgezee: @gabrielliwerant - instead of adding the `isExpandableIconClick` variable, and associated pieces that would need to be removed later, I added a single conditional to check for the deprecated condition that issues the warning and calls onRowClick explicitly. This should make it simpler to remove in a future release.

I copied the deprecated format from the only other instance of the word 'deprecated' in MUIDataTable.js, though downgraded to a warning for this case.

gabrielliwerant: Yes, good call, that does make it easier to remove, thank you!

Haha, looks like your search didn't include uppercase! I was thinking we would take after the convention here https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L266. I like using `error` because I think we want it to stand out as something devs should change in advance of removing the functionality.

Also, while you're at it, could you add the new options to the `expandable-rows` example, so that we have something to point people to?

georgezee: The line you linked to (L266) is the same one I copied from.
I don't feel that `error` would be the right level here, as developers would not have any way of removing it. i.e. if you have expandableRows turned on, and onRowClick defined, then the warning/error will appear every time someone expands a row. There is no way to avoid it, even if you don't depend on the `onRowClick` execution on the expand action.
An `error` would indicate that something needs to be changed, but that may not be true for most use cases, and in this case you wouldn't have any way to clear the error.

I've added `expandableRowsOnClick: true` to the `expandableRows` example and considered doing the same for `selectableRows` but decided that this wouldn't be the default preference for most users. Perhaps I should add `selectableRowsOnClick: false` to that example, so that the property is more visible?

gabrielliwerant: Your point is well taken. My thinking was that because the error only shows on the occasion of both `expandableRpws` and `onRowClick`, it wouldn't be a disruptive message to too many people, and I was thinking of erring on the side of a strong notice for any functionality people might be relying upon (since it's easier to ignore warnings).

I suppose the more robust way to deprecate would be to wrap the deprecated functionality with a new option, which people could then switch on to avoid the deprecated behavior, as well as the error. However, I don't think we need to be that robust here, and I don't have a strong enough feeling on the matter to hold the PR for any more changes here.

Also, my thoughts on adding the options to the examples was just for the sake of demonstration rather than something representing real-world use cases, but I'm fine with the single addition for now.

georgezee:
Great - thanks for all your work and reviews for improvement.
This will allow our implementation to switch away from the fork I've been running to implement what is a key feature for us.

Issue #43, opened 2018-05-05
Hide a column completely 


RaySheikh:
Hi,

first of all I want to thank you for putting this project together.  
 
Is there a way to hide a column completely from the view columns filter?
Currently If i use a column to store a key, the filter false option will hide it but it will still show up under view columns filter as unchecked.  
Or if i use blank column to add a button, it still shows up under the view column as a checked column. Is there a way to completely hide the column in the view column filter?

Thanks, 

Ray

gregnb:
hey @RaySheikh thanks for the feedback. Yes there is a way to hide it using the `display` property. An example of that usage:

```
const columns = [
      {
        name: "Name",
        options: {
          display: false,
        }
      },      
];
```

RaySheikh:
Hey @gregnb  thanks for replying. I am using the `display: false` property but when filtering the columns using the column filter button they still show up.  Please see the screenshot below.

![image](https://user-images.githubusercontent.com/35114618/39719677-c6d07d48-51ee-11e8-82f5-2abc0c5c4cdf.png)

Thanks,

Ray

gregnb:
Ok, I'll take a look into this bug tonight. That is in the 'view columns' but how about the filtering? is it hiding there properly?

RaySheikh:
Yes, the Filter Table works perfectly fine.

gregnb:
Ok thanks @RaySheikh  I'll reopen this ticket and fix the view columns issue

tgBryanBailes:
I'm not sure I understand the issue here. The way to hide a column completely is to not include it in the data you give to `mui-datatables`. Using the `display` option hides it on load, but allows the user to choose to see it if they would like. This seems to work exactly as I would expect.

gregnb:
@RaySheikh After looking into this issue @tgBryanBailes is right. If I were to change ViewColumns to show a checkbox based off display being true/false then users will not be able to toggle on/off after load. I don't want to add another property because simply not providing the data is how you should solve this

klequis:
I'll also start by saying thank you for developing this grid. It is really nice!

A way to hide a column is almost the first thing I looked for. In my case it will be a mongoDB object id which the user should not be able to see but will be used for all click events as a unique identifier for the row. Is there a way to not include it in the data give to the grid and still determine what data in the database the row is associated with. I could see keeping a matching data structure with ID and sync via index, but that isn't a clean approach.

Thanks,
Karl

georgezee:
[...]

PR #609, opened 2019-05-10
Document option for hiding column from 'View Column' list

georgezee:
I was looking to implement the feature requested in #43 , when I discovered the option already exists, it just is not documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ const columns = [
 |:--:|:-----|:--|:-----|
 |**`display`**|string|'true'|Display column in table. `enum('true', 'false', 'excluded')`
 |**`empty`**|boolean|false|This denotes whether the column has data or not (for use with intentionally empty columns)
+|**`viewColumns`**|boolean|true|Allow user to toggle column visibility through 'View Column' list
 |**`filterList`**|array||Filter value list [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
 |**`filterOptions`**|array||Filter options [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -208,6 +208,7 @@ const columns = [
 |:--:|:-----|:--|:-----|
 |**`display`**|string|'true'|Display column in table. `enum('true', 'false', 'excluded')`
 |**`empty`**|boolean|false|This denotes whether the column has data or not (for use with intentionally empty columns)
+|**`viewColumns`**|boolean|true|Allow user to toggle column visibility through 'View Column' list
```
gabrielliwerant: I like the added explanation, but to keep this language parallel with the other similar options, how about `Show/hide viewColumns icon from toolbar, which allows the user to...`

georgezee: @gabrielliwerant - this option doesn't affect the icon display. It can be confusing because the identically named option exists for the whole table (as well as individual columns). The former sets the icon, the latter is for the list that opens when the icon is clicked.

To add to the confusion, this option sets whether the column name should be shown/hidden from the list that shows/hides the column itself. 
Having said that, how about:
"Allow user to show/hide the column name in the 'View Column' list."

gabrielliwerant: Oh my, you are right, thanks for clarifying that. Given that, I have no issues with your description.

Seems like this option could use a better name. Something like `hideable`? Would you want to make those changes here? If not, that's fine, and we can move this one forward as is.

georgezee: It would be considered a breaking change if a site is using the existing option.
As such, it would need to also be documented as part of future release notes.

Personally, I'm ok to continue using the option as provided, and was just happy to discover it is already possible to do this.

gabrielliwerant: Agreed, it should be noted and bumped in version accordingly. I was planning on leaving that note in the PR if you wanted to move forward with it, but it's not critical.

georgezee: Sure, I can make the change to the column level option name, we should just agree on what it should be.
`hideable` is not appropriate because it will determine both if a column can be hidden (if visible) or visible (if hidden).

How about `allowToggle`?

PR #612, opened 2019-05-10
Fix import paths to support better tree shaking

nathancharles:
Importing directly from `@material-ui/core` is pulling in the entire bundle instead of the portions used.
This just updates the import paths across the app to use more direct imports which should allow for better tree shaking.

### DIFF
--- a/docs/utils/Menu.js
+++ b/docs/utils/Menu.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { withStyles } from '@material-ui/core';
+import { withStyles } from '@material-ui/core/styles';
 import Drawer from '@material-ui/core/Drawer';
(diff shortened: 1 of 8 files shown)


nathancharles:
@gregnb any chance to get this merged?

gabrielliwerant:
Thank you for taking an interest in the project!

My understanding of imports is that the destructuring form of `import { dep1, dep2, ... } from 'library'` already supports tree shaking by pulling out only the dependencies that are being used, so I don't think this PR is necessary, as all the lines you changed already use the correct form as far as I can tell.

gabrielliwerant:
Based on this resource https://developers.google.com/web/fundamentals/performance/optimizing-javascript/tree-shaking/, and my own understanding, I believe this library already supports treeshaking in production builds. Without any further input on this in the next few weeks, I'll close the PR.

Issue #531, opened 2019-04-02
How to access nested object properties in columns


raikusy:
Hello, I am passing an array of objects in **data** prop. Those objects in the array have some nested objects. I need to access the nested object properties. How can I do that?
Demo data:
```javascript
const data = [
{
    id: 1,
    person: {
      name: "Bla"
    }
  },
  {
    id: 2,
    person: {
      name: "Lol"
    }
  }
]
```
Expected column:
```javascript
<MUIDataTable
title="Person List"
columns={
  [{
      name: 'id',
      label: 'ID',
      options: {
        filter: true,
        sort: true,
      }
    },
    {
      name: 'person.name',
      label:  'Person Name',
      options: {
        filter: true,
        sort: false,
      },
    },
  ]
}
data={history}
/>
```
So all I want is to show the person name in column. How can I do that? The `person.name` in above code doesn't work.

MBCSDave:
I think the only way you have is to flatten your data structure.

wdh2100:
var flattenObject = function(ob) {
  
  return Object.keys(ob).reduce(function(toReturn, k) {

    if (Object.prototype.toString.call(ob[k]) === '[object Date]') {
      toReturn[k] = ob[k].toString();
    }
    else if ((typeof ob[k]) === 'object' && ob[k]) {
      var flatObject = flattenObject(ob[k]);
      Object.keys(flatObject).forEach(function(k2) {
        toReturn[k + '.' + k2] = flatObject[k2];
      });
    }
    else {
      toReturn[k] = ob[k];
    }

    return Object.keys(toReturn).map(function(_) { return toReturn[_]; });
  }, {});
};

var data  = [
{
    id: 1,
    person: {
      name: "Bla"
    }
  },
  {
    id: 2,
    person: {
      name: "Lol"
    }
  }
];

var arrayValues = data.map( (_data) =>{return flattenObject(_data)});

console.log(arrayValues);

raikusy:
@wdh2100 Hey! Thanks for your reply. I was able to flatten my object using your code. But problem is, in my real data there are many unused properties, which I don't want to show in columns. I want to show specific items in columns. How can I do that?

raikusy:
Hey thanks for all the response. My problem is solved.
```javascript
function flattenObject(ob) {
  const toReturn = {};

  Object.keys(ob).map(i => {
    if (typeof ob[i] === 'object' && ob[i] !== null) {
      const flatObject = flattenObject(ob[i]);
[...]

Issue #475, opened 2019-03-08
Using mui-datatables in a type-safe way by having access to the original object


dandrei:
(mostly based on my reply on issue #109)

Hi @gregnb, thank you for your great work on this.

I wonder if it's possible use `mui-datatables` in a more "type-safe" way. Maybe it's already available or maybe it's something that can be easily implemented.

Currently we specify the columns as such:

    const columns = [
      {
        name: 'object_field',
        label: 'Object field'
      }
    }

This will take the `object_field` field from each object in the array sent to `data`. However, specifying the field name as a string isn't type-safe. TypeScript has no way of knowing whether objects in `data` contain or don't contain the field `object_field`.

### Question:

Is it possible perhaps, instead of having `name` as a required field, another option could be added? (let's call it `value`, required when `name` isn't specified), that would be used as such:

    const columns = [
      {
        value: (obj) => obj.object_field,
        label: 'Object field'
      }
    }

This way, we could use it with TypeScript:

    value: (obj: KnownType) => obj.object_field,

...and TypeScript would check that `object_field` does indeed belong to `KnownType`. Since it's an extra option, it wouldn't interfere with the current way of doing things or JS-only users.

Would also be nice to get the original data object in `customBodyRender`. Use case for this is: if we want to render one column based on a value from another.

Right now we *do* have access to the values array in `tableMeta.rowData`,  The caveat is: the order matters. Changing the column order requires a change in how we access `tableMeta.rowData`. But would be nice to be able to keep the type information if using TypeScript.

So maybe we could have the original object available in the parameters sent to `customBodyRender`?

What are your thoughts? Thanks!

gregnb:
hey @dandrei thanks for taking the time to write and explain your issue. This would be a big breaking change and I'm not comfortable at the moment doing this. With that said, if a enough of the user community comes back saying this is affecting them I would consider making this change. I will make this issue for community feedback

dandrei:
@gregnb thanks for your prompt reply. In the meanwhile, I've written an API that sits in front of the `mui-datatables` API and does exactly what I wanted :).

It takes:
- A list of objects.
- Type-aware column definitions (a definition object consists of the column's `label`, and two functions: one for formatting the `value` (used in sorting), the other used to `render` what gets displayed).

It then produces `mui-datatables`-compatible `data` and `columns` that can be plugged directly into `<MUIDataTable />`.

[...]

Issue #147, opened 2018-08-21
Allow passing data as an array of objects


rooby:
It would be very useful to be able to pass in an array of object for data and then in the column definition you would specify the name of the property on the object.

It would be even better if you could specify nested properties with dot notation, similarly to how [react-table](https://react-table.js.org) works.

jkeruzec:
For feedback,

With my spring rest controller, I have developped utils to take my Rest JSON data and transform it to a raw data array mapped to column to be able to display the data.

It will be very usefull to directly be able to read the JSON and avoid this specific mapping technique.

So for short, yes interested in this functionnality ! ;)

benjaminbours:
Same as @jkeruzec , i'm interested too :)

x0fma:
Does someone started to work on this ?

stipe09:
I'm interested too, it will be very usefull.

VIJAYMYSORE:
+1

pavlovicIvan:
Hi, this feature would be very usefull for our project.
Do you know when will be implemented?

stipe09:
Are you considering to do this feature?

sidharthdigiworks:
It's easier if we have array of objects since data from server usually is formatted that way and we can opt to choose which properties need to be shown in the table.

jpmillan:
+1

revskill10:
You can use this helper function to transform:
```
const convertDataToArray = (data, keys) =>
  data.map(obj => Array.from(keys, k=>obj[k]))

```

Diaver:
I've made in my fork: https://www.npmjs.com/package/material-datatable

rooby:
@Diaver are you going to submitting a pull request to address this issue in this project?

Diaver:
@rooby I am not it is possible -> In my form at least one big change: my datatable accept object array

revskill10:
@Diaver Why not adding one prop like `itemDataType`, which could be `array` or `object` ?

Diaver:
@revskill10 because it is not just one required change

revskill10:
@Diaver I mean let's just reuse the logic, and what we can do is transforming/processing props before passing down the render logic. Because we want to keep maintainance on one project only (?)

Diaver:
@revskill10 
There are some reasons:
1. It is already a lot of changes in my fork.
2. Keep both logic in one project doesn't seems to be good idea for me, at least until current code will be completely refactored.
[...]

PR #619, opened 2019-05-13
Allow for accessing nested object data.

GrandVizierOlaf:
This PR allows for the use of dot notation in the column names to access nested object data.

For instance, the modified example adds a phone object to each existing user and each phone has a "home" and "cell" attribute. These nested attributes can be accessed with a column name of "phone.home" and "phone.cell", respectively.

Any amount of nested will work with this.

This addresses #147, #531, and (I think) #475.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -111,12 +111,14 @@ class MUIDataTable extends React.Component {
       customToolbar: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
       customToolbarSelect: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
       customFooter: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
+      customRowRender: PropTypes.func,
       onRowClick: PropTypes.func,
       resizableColumns: PropTypes.bool,
       selectableRows: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['none', 'single', 'multiple'])]),
       isRowSelectable: PropTypes.func,
       serverSide: PropTypes.bool,
       onTableChange: PropTypes.func,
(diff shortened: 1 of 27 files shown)


skyleronken:
Can we merge this? the conflict is in the package-lock.json file and those changes are only on optional flags.

gabrielliwerant:
We can't merge this until the conflicts are resolved, but it also needs review, which I will provide when I have the time.

src/MUIDataTable.js
```diff
@@ -374,6 +374,8 @@ class MUIDataTable extends React.Component {
     return { columns: columnData, filterData, filterList };
   };
 
+  leaf = (obj, path) => (path.split('.').reduce((value,el) => value[el], obj));
```
gabrielliwerant: A very elegant solution, but we need to be forgiving of missing data, as we are with other cells at the moment. If we can't find the object property specified in the path, we should make sure to set `undefined` instead of cause a breaking error.

gabrielliwerant:
Also would be great to have a test for this, if you've got the time.

src/MUIDataTable.js
GrandVizierOlaf: I have made the requested change, but I can't make heads or tails of how to test it properly. If you can offer some guidance on where the test(s) would live I can give it a shot, but you can see from the modified example that I spot checked with invalid attributes and invalid sub-attributes.

epavan:
Hey guys, is this feature working? I've the following issue:

const applicant = { name: 'test', surname: 'test' };

Screenshot 2019-08-01 at 12 59 21

gabrielliwerant:
@epavan I need more information to debug your code (in particular, the actual code). Make sure you are using this feature as indicated in the example `data-as-object`.

Issue #617, opened 2019-05-13
Add onTableChange callback on table initialization


shawnmitchell:
## Expected Behavior
I'm using a custom toolbar and using `onTableChange` to track the table state.

I have that state as soon as the user performs a sort or filter operation, but not before then. Adding a call to `onTableChange` on initialization will provide the client with the internal state to handle any custom actions on an unmodified table.

It is reasonable imo to fire the `onTableChange` callback on initialization as the table state has in fact changed from uninitialized to initialized. 

I created a pull request with my fix but some tests are not passing. If this is something the owners believe fits with the library I will be more than happy to work with them to fix the tests. 

My change is a one-liner to add a callback to the call to `setTableData` to call `setTableAction` after `setState` completes. `setTableAction` results in `onTableChange` being fired and the client receiving the initialized table state. 

```
initializeTable(props) {
    this.getDefaultOptions(props);
    this.setTableOptions(props);
    this.setTableData(props, TABLE_LOAD.INITIAL, () => {
      this.setTableAction('tableLoadInitial');
    });
  }
```

shawnmitchell:
If there is interest, I'll also add the callback to 

```
componentDidUpdate(prevProps) {
    if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
      this.setTableData(this.props, TABLE_LOAD.INITIAL, () => {
        this.setTableAction('tableLoadInitial');
      });
    }

    if (this.options.resizableColumns) {
      this.setHeadResizeable(this.headCellRefs, this.tableRef);
      this.updateDividers();
    }
  }
```

gabrielliwerant:
Hm, how about a separate callback for initialization like `onTableInit` or something? I would accept something like that. I worry about combining with a change callback, as I might not expect that to fire on initial hydration.

shawnmitchell:
@gabrielliwerant, that would also work. Would you be okay sending it on construction and componentDidUpdate where prevProps !== this.props? For me, sending new props is essentially reinitializing the table, so `onTableInit` would still be appropriate. 

Add setTableInitialized in the form of setTableAction...
```
  setTableAction = action => {
    if (typeof this.options.onTableChange === 'function') {
      this.options.onTableChange(action, this.state);
    }
  };

  setTableInit = action => {
    if (typeof this.options.onTableInit === 'function') {
      this.options.onTableInit(action, this.state);
    }
  };
```

add `setTableInit` callback in `initializeTable` 
```
[...]

PR #622, opened 2019-05-15
add onTableInit callback

shawnmitchell:
add onTableUpdate on props change
update README.md
update test coverage


shawnmitchell:
@gabrielliwerant https://github.com/gregnb/mui-datatables/issues/617

shawnmitchell:
If accepted, I'll also update https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/mui-datatables

shawnmitchell:
that rollup.config.js change is just because you can't `yarn upgrade...` link to a branch with that deprecated uglify warnings option.

shawnmitchell:
@gregnb @gabrielliwerant I'd love to get this reviewed. I think it opens up a lot of good solutions for custom toolbars. I am cognizant of the fact that you just got 2.2.0 out the door, but this is a nice, simple improvement with no changes to internals.

gabrielliwerant:
@shawnmitchell It's on my radar, I've just been busy this past week and weekend!

rollup.config.js
```diff
@@ -35,7 +35,7 @@ export default {
     }),
     uglify({
       compress: {
-        warnings: false,
+        // warnings: false,
```
gabrielliwerant: Can you provide the exact yarn version used and the command run so I can test this? I did not have a problem with this locally.

test/MUIDataTable.test.js
```diff
@@ -800,6 +810,14 @@ describe('<MUIDataTable />', function() {
     assert.deepEqual(state.data[0].data[2], 'Las Vegas');
   });
 
+  it('should call onTableInit when MUIDataTable is initialized', () => {
```
gabrielliwerant: This is great! But I think it would be nice to also trigger an update and ensure that the call count for `onTableInit` remains `1`. Something from one of the tests to table update could serve well here, such as:

```
    instance.selectRowUpdate('head', 0);
    shallowWrapper.update();
```

rollup.config.js
shawnmitchell: Hey @gabrielliwerant, I'm using yarn 1.15.2. The command I'm using is `yarn upgrade mui-datatables@https://github.com/shawnmitchell/mui-datatables.git#enhancement-on-table-init` so that grabs the fork from my repo and rebuilds locally. This is not part of the pull request per se, I just modified it so I could link to my stuff in my project.

test/MUIDataTable.test.js
shawnmitchell: This is a great suggestion. Added this: 

```
  it('should call onTableInit only 1 time when creating table and calling selectRowUpdate method with type=head', () => {
    const options = { selectableRows: true, onTableInit: spy() };
    const shallowWrapper = shallow(<MUIDataTable columns={columns} data={data} options={options} />).dive();
    const instance = shallowWrapper.instance();

    instance.selectRowUpdate('head', 0);
    shallowWrapper.update();

    assert.strictEqual(options.onTableInit.callCount, 1);
  });
```

shawnmitchell:
> This looks really good, just a few minor things!
> 
> Also, can you provide an example on this function being used by a table? One of the generic examples might be best, but it's up to you.

Added an example based on my use case of needing tableState before any table manipulations that trigger `onTableChange` to support Excel download

examples/customize-toolbar-excel-download/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+      const data = tableState.data.filter(row => filteredRowIndexArray.includes(row.index));
+      const visibleColumns = columns.filter(col => col.display === 'true' && col.download === true);
+
+      // boolean array of visible columns
+      const colMap = columns.map(col => col.display === 'true');
+      const filteredData = data.map(({ data }) => data.filter((cell, i) => colMap[i] === true));
+
+      const sheetHeader = [visibleColumns.map(col => col.label)];
+      const data = sheetHeader.concat(filteredData);
```
gabrielliwerant: Beware of duplicate `const` declaration here, as this will throw an error.

examples/customize-toolbar-excel-download/index.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import MUIDataTable from '../../src';
+import CustomToolbar from './CustomToolbar';
+// import XLSX from 'xlsx';
+
+const Example = () => {
+  const [tableState, setTableState] = React.useState();
```
gabrielliwerant: Can we avoid using hooks, as this is a newer React feature that won't work on lower versions, such as those supported by this library (which will also make this example more inclusive, since more people will be able to run it).

rollup.config.js
gabrielliwerant: Thanks for that. Yeah, it looks like the warning option is deprecated in the `compress` config. I'll fix in a separate PR, so as not to tangle it with other issues.

examples/customize-toolbar-excel-download/index.js
shawnmitchell: Sure thing. I'm very deep into TS and Hooks these days but I'll fix it up today.

examples/customize-toolbar-excel-download/index.js
shawnmitchell: Yep, definitely a bug. Will fix asap. Thanks!

shawnmitchell:
Thanks @gabrielliwerant for the review. That example code I pushed was a bit of an embarrassment. I've fixed it up, removing hooks, and also just got it to actually run properly. Took me a minute to figure out how to setup to run and test `examples` :-|

examples/customize-toolbar-excel-download/index.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import MUIDataTable from '../../src';
+import CustomToolbar from './CustomToolbar';
+import XLSX from 'xlsx';
```
gabrielliwerant: This import presents a problem for those looking to run this locally, as we won't have access to this package without adding it manually. Ideally, the examples would run without having to download external dependencies. Is there a way to perhaps mock this package locally so that it will run and demonstrate the code works without necessarily needing all the excel transformations? Maybe this example is too ambitious.

shawnmitchell: It’s not really doing much with xlsx except converting a 2d array to a sheet and downloading it. Maybe just comment out that stuff and the import? That way if someone wants to use it as a template they have the code.

gabrielliwerant: The problem is that the maintainers of this library will be responsible for anything put into it, so if it's unclear how a particular example works, it could lead to people opening issues with questions about it, that then require attention. Alternatively, if the XLSX code is included in the example, it's a lot of extra code in this project for a single example.

I think it would be best at this point to just remove the XLSX-dependent code so that it doesn't lead to confusion or additional questions about how to make it work. Excel downloads in general were something that this library has steered away from providing because it's a lot of additional code, and there's a reasonable desire to keep things a bit trim, instead of becoming a humongous, everything-to-everyone style of framework.

At some point, it may be good to add the ability to handle 3rd-party plugins to extend the functionality of the library without bloating it, and at that point, we can reconsider more complex examples and use cases.

shawnmitchell: Sounds good to me. I'll pull it out.

gabrielliwerant: I added this PR to demonstrate what I meant: https://github.com/gregnb/mui-datatables/pull/651. No need to pull everything, just the parts specific to the XLSX package. You can add the alterations here or we can move forward with the new PR, it's up to you.

shawnmitchell: I just renamed the example folder name since we're not doing an excel download anymore.

shawnmitchell:
Hey, that's great! It was nice working with you, @gabrielliwerant on this.

PR #647, opened 2019-05-24
Bugfix/search text issues

gabrielliwerant:
Fixes a couple of issues introduced by the `searchText` option.

### DIFF
--- a/examples/customize-search/index.js
+++ b/examples/customize-search/index.js
@@ -1,8 +1,12 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import ReactDOM from 'react-dom';
 import MUIDataTable from '../../src/';
 
 class Example extends React.Component {
+  state = {
+    searchText: 'Computer'
+  };
+
   render() {
     const columns = [
       {
@@ -77,7 +81,8 @@ class Example extends React.Component {
       filter: true,
       filterType: 'dropdown',
       responsive: 'stacked',
-      searchText: 'Computer',
+      page: 1,
+      searchText: this.state.searchText,
       customSearch: (searchQuery, currentRow, columns) => {
         let isFound = false;
         currentRow.forEach(col => {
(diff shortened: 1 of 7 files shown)


ericpyle:
@gabrielliwerant thanks for fixing those! 

I believe the other searchText behavior I need is that if searchText is used to initialize the table, then the table will initially show the search text box as being open with the searchText as its text value. Otherwise the table would not look like it's in the right state. Is that how you expect it to work?

gabrielliwerant:
@ericpyle That should be working in this branch. Feel free to pull it down and test it out!

ericpyle:
@gabrielliwerant I was just trying to test this locally, and ran into build errors (e.g. `'xargs' is not recognized as an internal or external command`). But it looks like it's in master now so I'll test it there :)

ericpyle:
@gabrielliwerant I see that you still need to bump the version before I can test via npm.  Btw, after making "prettier" a no-op in package.json, I then got this error:

```
C:\Users\PyleE\Documents\repos\github\mui-datatables>yarn build
yarn run v1.13.0
$ cross-env NODE_ENV=production npm run prettier && rollup -c

src/index.js → dist/index.js...
[!] (uglify plugin) Error: Error transforming bundle with 'uglify' plugin: `warnings` is not a supported option
Error: Error transforming bundle with 'uglify' plugin: `warnings` is not a supported option
    at error (C:\Users\PyleE\Documents\repos\github\mui-datatables\node_modules\rollup\dist\rollup.js:189:14)
    at Promise.resolve.then.then.catch.err (C:\Users\PyleE\Documents\repos\github\mui-datatables\node_modules\rollup\dist\rollup.js:17630:5)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:757:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

ericpyle:
@gabrielliwerant `yarn dev` does run the `mui-datatables` repo, but I'm not quite sure how to test that in my own project (pardon if I'm missing something obvious!)

gabrielliwerant:
You're right, this isn't super easy to do. One way is to fork the project, add a new build off of master, and pull that into your project via npm. Another way is to manually clone it into your node_modules and run the build on the master branch there.

You could also try running the project locally, and then adding in your functionality to an existing example.

If you can afford to wait for the next release, then you can test it then and open a new issue if there are still bugs.

ericpyle:
@gabrielliwerant Thanks. I was able to do a build with `yarn prepare` off `master`. After not having much luck linking to the `dist` directory, I was able to do a `yarn add .../mui-datatables/dist` into my project. `searchText` did seem to bring up the search bar, that's great!

Issue #610, opened 2019-05-10
Count option is not refreshed Server side rendering V2.1.0


jkeruzec:
I've got some issues with count options with server side rendering.
Count is not updated when I update my options parameter.

1. Count is initialized with 0 in first render
2. Then count is fetch with async method 
3. Options is updated with right count
4. MuiDatatable is not refresh (as options is setup in componentWillMount method) It uses only number of element of my data array

``` javascript
async componentDidMount() {
        // Start fetching data
        manufacturersCollection.fetch();
        const value = await phonesCollection.fetchCount(this.componentName)
        runInAction(() => this.count = value);
        phonesCollection.fetch({data: {category: this.componentName, page: 0}});
        const navigationData = this.props.location.navigationData;
        NavigationUtils.showNavigationMessage(navigationData, this.snackBar);
    }

// Prepare data from collection Transform list of object to simple array with order
    @computed
    get preparedData() {
        return DataTableUtils.createDataArray(phonesCollection, this.columns);
    }
render() {
        
        const {classes, breadcrumbs} = this.props;
        
     // Options for basic grid
        // TODO: Options in custom component
        const options = { selectableRows: false,
                        serverSide: true,
                        count: this.count,
                        download: true,
                        print: false,
                        viewColumns: false,
                        responsive: "scroll",
                        filterType: 'dropdown',
                        rowsPerPage: 15,
                        onTableChange: (action, tableState) => {
                            switch (action) {
                            case 'changeRowsPerPage':
                            case 'changePage':
                                phonesCollection.fetch({data: {category: this.componentName, page: tableState.page, size: tableState.rowsPerPage}});
                              break;
                            case 'search':
                                if(tableState.searchText && tableState.searchText.length > 1) {
                                    phonesCollection.fetch({data: {category: this.componentName, page: tableState.page, size: tableState.rowsPerPage, searchText: tableState.searchText}});
                                } else if(tableState.searchText === null) {
                                    phonesCollection.fetch({data: {category: this.componentName, page: tableState.page, size: tableState.rowsPerPage}});
                                }
                                break;
                            default:
                                break;
                          }
                        },
                        downloadOptions: {separator: ';'}
                    };
        
        // Display normal data on page
        return (
                <React.Fragment>
                    {breadcrumbs}
                    <div style={{ margin: 20 }}>
                    {/* Page header + toolbar  */}
                    <Typography classes={{root: classes.headLineDataTable}} variant="h4" color="initial" noWrap>
[...]

PR #648, opened 2019-05-25
Fix issue #610 about count from server does not refresh

bndynet:
Fix the `count` does not refresh from server

### DIFF
--- a/examples/serverside-pagination/index.js
+++ b/examples/serverside-pagination/index.js
@@ -7,6 +7,7 @@ class Example extends React.Component {
 
   state = {
     page: 0,
+    count: 1,
     data: [["Loading Data..."]],
     isLoading: false
   };
--- a/examples/serverside-pagination/index.js
+++ b/examples/serverside-pagination/index.js
@@ -1,13 +1,13 @@
 import React from "react";
 import ReactDOM from "react-dom";
-import { CircularProgress } from '@material-ui/core';
+import { CircularProgress, Typography } from '@material-ui/core';
 import MUIDataTable from "../../src/";
 
 class Example extends React.Component {
 
   state = {
     page: 0,
-    count: 100,
+    count: 0,
     data: [["Loading Data..."]],
     isLoading: false
   };
@@ -20,7 +20,7 @@ class Example extends React.Component {
   getData = () => {
     this.setState({ isLoading: true });
     this.xhrRequest().then(data => {
-      this.setState({ data, isLoading: false });
+      this.setState({ data, isLoading: false, count: 121 });
     });
   }
 
@@ -49,9 +49,14 @@ class Example extends React.Component {
   }
 
   changePage = (page) => {
+    this.setState({
+      isLoading: true,
+    });
     this.xhrRequest(`/myApiServer?page=${page}`).then(data => {
       this.setState({
+        isLoading: false,
         page: page,
+        count: 450,
         data
       });
     });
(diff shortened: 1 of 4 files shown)


isaacperaza:
+1, I tested and this fix works.

bndynet:
@gabrielliwerant Do you have time to review my code? If something wrong, please give me feedback, I will fix it ASAP. Thanks in advance.

src/MUIDataTable.js
```diff
@@ -1064,7 +1064,7 @@ class MUIDataTable extends React.Component {
       searchText,
     } = this.state;
 
-    const rowCount = this.options.count || displayData.length;
+    const rowCount = this.props.options.count || this.options.count || displayData.length;
```
gabrielliwerant: This is a great catch, and certainly seems to work on my end as well!

However, there might be an opportunity here for an extra fix. There are a lot of table values that currently rely on props and watching for props to change, to refresh the state, however, I think this is a react anti-pattern. I believe it would be better to rely more on state instead, and I think the library should move in this direction.

What that would mean is that we would need to initialize a `count` in the initial state, add `count: options.count,` to this https://github.com/gregnb/mui-datatables/blob/c1cc06fc96c11a1af7ac23b9b856a2c31dca604e/src/MUIDataTable.js#L482 and then change your line to look for `this.state.count` as well.

What do you think? I'm not married to this approach, and I don't want to hold up the fix too long, so we can always go with this and make additional changes later.

examples/serverside-pagination/index.js
```diff
@@ -1,13 +1,13 @@
...
-import { CircularProgress } from '@material-ui/core';
+import { CircularProgress, Typography } from '@material-ui/core';
 import MUIDataTable from "../../src/";
 
 class Example extends React.Component {
 
   state = {
     page: 0,
-    count: 100,
+    count: 0,
```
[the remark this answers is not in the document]
bndynet: 💯 Yes, removing the `count` will work fine. That's my mistake. It's just for testing. Thanks for pointing it out.

examples/serverside-pagination/index.js
```diff
@@ -20,7 +20,7 @@ class Example extends React.Component {
   getData = () => {
     this.setState({ isLoading: true });
     this.xhrRequest().then(data => {
-      this.setState({ data, isLoading: false });
+      this.setState({ data, isLoading: false, count: 121 });
```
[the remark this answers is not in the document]
bndynet: The response `data`  is one-page data, so `data.length` is the page size.  I want to test the pagination from the server side.  The `data` from server-side is one-page data, not total. And the `count: 121` is records total.

src/MUIDataTable.js
bndynet: In my case, the record count and page total will be changed after page changed. So I need a way to refresh it. Currently, I am using solution https://github.com/gregnb/mui-datatables/issues/610#issuecomment-491197237.  Is it the right way to fix this? Or any best ways to meet my case?

bndynet:
**Loading**

image

**Data initialized**

image

**Page changed**

image

examples/serverside-pagination/index.js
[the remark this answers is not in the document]
gabrielliwerant: This example is not doing what you think. The entire state of the data is being overridden here. You can see all of the state data by adding a callback to this line `() => console.log(this.state.data)`. After this section fires, the count is `data.length`, but is being set to `121` anyway.

src/MUIDataTable.js
gabrielliwerant: The solution I suggested above also works, but it gets us away from props, and uses state internally, instead.

bndynet: Good solution, I had used internal `state.count` to instead of `props.count` and it worked.

gabrielliwerant:
Still not sure where we're getting the values of `121` and `450` from. They appear random to me and don't match the actual number of results. I think what we want to test here is dynamically changing the count value, but can you explain why we're using the values above instead of something else?

isaacperaza:
@gabrielliwerant I'm using the fix suggested by @bndynet and it works perfect, check production output:

Screen Shot 2019-06-03 at 9 20 37 AM

The image display 1 to 25 records (Page 1) of a total of 109 records (Total of server side data).

Note: I just share a bottom table screenshot and I modified the user names for confidentiality.

bndynet:
@gabrielliwerant Sorry about lated reply. You point out my mind. The method to test this case may be confused, so I changed the test code to clarify it.

Issue #643, opened 2019-05-23
Changing columns prop no longer updates state after 2.0.0


muvari:
## Expected Behavior

Changing the columns prop should update the internal column state and render the new columns. This was working in 2.0.0.

## Current Behavior

The columns stay the same after updating the prop.

## Steps to Reproduce (for bugs)

1. Create an initial table
2. Add a new column to the columns prop
3. Table does not render the new column

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.9.3     |
| MUI-datatables  |   2.2.0      |
| React        |   16.8.6      |

IDrissAitHafid:
Same problem, i have the same environment!!

karthick-vinod:
Sandbox Link: https://codesandbox.io/embed/cranky-goldberg-koszc

**Probable cause:**  (Might be wrong)

The condition at https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L316 will always be true because `this.props.columns` already contains `newColumns` as it is being called from `componentDidUpdate`

![image](https://user-images.githubusercontent.com/45751032/58467390-2edfc480-8159-11e9-8c07-8670a360f46b.png)



This could have been the case since certain logic was moved out of `componentWillReceiveProps` into `componentDidUpdate`.

![image](https://user-images.githubusercontent.com/45751032/58466561-7f562280-8157-11e9-8aae-fade0837becb.png)

SatorSquare:
I am also seeing this issue and am interested in a fix.

karthick-vinod:
PR: https://github.com/gregnb/mui-datatables/pull/650

tiagofernandez:
Updating the `options` prop also stopped working, after 2.0.0.

karthick-vinod:
> Updating the `options` prop also stopped working, after 2.0.0.

Handled `options` and `columns` now.

patorjk:
Just ran into this myself. I tried dynamically updating the rowsPerPage option and it didn't work:

https://codesandbox.io/s/0lmlx

If I use the 2.0.0 version of mui-datatables (as the original sandbox does), it works fine.

VagnerNico:
Hello again, I guess this PR [#687](https://github.com/gregnb/mui-datatables/pull/687) fix the column options changed outside the table.

jkeruzec:
Very interested in this one. 
I can't migrate to V2.x.x because of this issue for the moment..

[...]

PR #650, opened 2019-05-28
Fix for issues/610 and issues/643

karthick-vinod:
Fix for https://github.com/gregnb/mui-datatables/issues/610 and https://github.com/gregnb/mui-datatables/issues/643

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -196,7 +196,15 @@ class MUIDataTable extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
+    if (
+      this.props.data !== prevProps.data ||
+      this.props.columns !== prevProps.columns ||
+      this.props.options !== prevProps.options
+    ) {
+      if (this.props.options !== prevProps.options) {
+        this.getDefaultOptions(this.props);
+        this.setTableOptions(this.props);
+      }
       this.setTableData(prevProps, TABLE_LOAD.INITIAL, () => {
         this.setTableAction('propsUpdate');
       });
@@ -388,15 +396,13 @@ class MUIDataTable extends React.Component {
   };
 
   setTableData(props, status, callback = () => {}) {
-    const { options } = props;
-
     let tableData = [];
     let { columns, filterData, filterList } = this.buildColumns(props.columns);
     let sortIndex = null;
     let sortDirection = null;
 
     const data = status === TABLE_LOAD.INITIAL ? this.transformData(columns, this.props.data) : props.data;
-    const searchText = status === TABLE_LOAD.INITIAL ? options.searchText : null;
+    const searchText = status === TABLE_LOAD.INITIAL ? this.options.searchText : null;
 
     columns.forEach((column, colIndex) => {
       for (let rowIndex = 0; rowIndex < data.length; rowIndex++) {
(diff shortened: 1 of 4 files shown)


gabrielliwerant:
Just a heads up, I'm not ignoring this PR. I have in fact begun to review it, and will hopefully have feedback for you within the next few days.

gabrielliwerant:
@karthick-vinod Firstly, thank you for being proactive and attempting to fix this issue. It is an important issue, however, I've been working on it a bit and thinking about it a lot, and I think there are fundamental problems causing it in the library at the moment.

We are extending the use of the `componentDidMount` lifecycle here as a way to refresh derived state. Attempting to use lifecycle hooks in this way is recommended against by react as it causes numerous issues (and also turns out to be very difficult to test): https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html.

This is not your fault, as you're just following the existing conventions. However, we do indeed have edge cases with these kinds of fixes, and in this case, it does revert prior fixes/issues (https://github.com/gregnb/mui-datatables/issues/305, https://github.com/gregnb/mui-datatables/pull/585, https://github.com/gregnb/mui-datatables/pull/597).

Fundamentally, the library is attempting to allow users to have it both ways when it comes to state. It is attempting to allow changes via props to be reflected in the table AND it is attempting to manage an internal derived state which should also update on user actions and prop updates. This entails two separate sources of truth: the props sent to the table, and the internal state of the table as derived from those props. I think this project is too complex to allow both approaches without becoming a maintenance nightmare.

I think the library should shift to forcing most changes to occur via props instead of derived internal state, but this will be a major shift in the API and cause a lot of breaking changes. What I'd like to do at the moment is use this as an opportunity to shift the library in that direction, without finishing it, hopefully fixing the bugs here without breaking anything else and paving the way forward. I will invite you to take a look when I'm done.

If it seems like it is taking too long for the changes, I will return to this PR and see if there is a way it can be cleaned up to prevent the regressions so that we are not without a fix for too long.

gabrielliwerant:
@karthick-vinod I'll go back to reviewing this PR and see if we can get a fix in here that doesn't include regressions. Work on the refactor will take place separately, as it is too large to handle in response to critical bugs: https://github.com/gregnb/mui-datatables/issues/679.

tiagofernandez:
It'd be nice to get this fixed in v2, as I think many of us are stuck with v2.0.0 – thanks @gabrielliwerant for looking into this!

src/MUIDataTable.js
```diff
@@ -196,8 +196,16 @@ class MUIDataTable extends React.Component {
...
   componentDidUpdate(prevProps) {
-    if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
-      this.setTableData(this.props, TABLE_LOAD.INITIAL, () => {
+    if (
+      this.props.data !== prevProps.data ||
+      this.props.columns !== prevProps.columns ||
+      this.props.options !== prevProps.options
+    ) {
+      if (this.props.options !== prevProps.options) {
+        this.getDefaultOptions(this.props);
```
gabrielliwerant: Reverting back to the default props here is problematic as it conflicts with changes in state. For example, this opens a regression where setting a page number in props and then changing the page number in the table causes any change in options to revert back to the original page.

src/MUIDataTable.js
```diff
@@ -196,8 +196,16 @@ class MUIDataTable extends React.Component {
...
+        this.setTableOptions(this.props);
+      }
+      this.setTableData(prevProps, TABLE_LOAD.INITIAL, () => {
```
gabrielliwerant: Using `prevProps` here opens this regression: https://github.com/gregnb/mui-datatables/pull/597.

test/MUIDataTable.test.js
```diff
@@ -211,6 +220,67 @@ describe('<MUIDataTable />', function() {
     assert.deepEqual(state.data, expectedResult);
   });
 
+  it('should correctly re-build internal table data and displayData structure with columns prop change', () => {
```
gabrielliwerant: Much appreciate the attempt to test the library and the changes, however, both of these tests pass in `master` or with all your other changes here removed, meaning that they don't actually demonstrate that your code fixes the problem. This is part of the difficulty with logic in the lifecycle methods, in that it is very difficult to mimic their behavior in the same way as in actual usage of the library.

patorjk:
Should be fixed as of version 2.15.0

PR #651, opened 2019-05-29
Add onTableInit

gabrielliwerant:
Based on https://github.com/gregnb/mui-datatables/pull/622


shawnmitchell:
I just thought it would make sense to change the name of the example since we’re no longer downloading excel

Issue #605, opened 2019-05-10
Clean up UI around multiple default `sortDirection` options


gabrielliwerant:
The `sortDirection` behavior only affects the final column sorted, as multiple columns cannot be sorted simultaneously, since this would require reordering individual cell data among rows, thus scrambling the table data. However, it is unclear what is happening from a UI perspective, as it shows both arrows.

What should probably occur is that only the final arrow should show, and there should be a console warning that multiple columns have been set to be sorted.

patorjk:
The jQuery Datatables handles this by having an order property in the table that's an array:

https://datatables.net/reference/option/order

The table allows you to select multiple columns by holding Shift and clicking the column. It's pretty useful. Example use case: Sort a table by the Last Name column, and then secondary sort on the First Name column.

gabrielliwerant:
Hm, not sure I understand how a secondary sort would work since it must necessarily change the sort order from the first time around. The only time I can think of where it would matter is if there are multiple entries that are the same on the initial sort. But I'm not sure how to make it clear which sort direction should receive priority. If you can think of a decent API and UX for that feel free to open a new issue with a detailed explanation, and it can be treated as a new feature. This issue is specifically to solve some confusing UI around the current functionality.

gabrielliwerant:
Should be resolved in `2.8.0`.

PR #659, opened 2019-06-03
Fix issues/605: Handle multiple columns with sortDirection

karthick-vinod:
Fix for https://github.com/gregnb/mui-datatables/issues/605

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -332,6 +332,8 @@ class MUIDataTable extends React.Component {
     let filterData = [];
     let filterList = [];
 
+    let sortEnabled = false;
+
     if (this.state.columns.length && isEqual(this.rawColumns(newColumns), this.rawColumns(this.props.columns))) {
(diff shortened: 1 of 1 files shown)


src/MUIDataTable.js
```diff
@@ -332,6 +332,8 @@ class MUIDataTable extends React.Component {
     let filterData = [];
     let filterList = [];
 
+    let sortEnabled = false;
```
gabrielliwerant: Since we have a `sort` option as well as a `sortDirection` option, the var name might be a little more clear if it were something like `sortDirectionSet` instead.

gabrielliwerant:
I'll give this another week or so and then open up a new PR based on it with the changes I'm after and close this one, just fyi.

gabrielliwerant:
Continued here https://github.com/gregnb/mui-datatables/pull/814.

gabrielliwerant:
Closed via https://github.com/gregnb/mui-datatables/pull/814.

PR #664, opened 2019-06-04
Updates 'options' in componentDidUpdate

noobtiger:
In previous version, it looks like options were being updated everytime in componentWillRecieveProps. But it was removed and options were not being applied if they were updated.
I can elaborate the use case if you want.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -211,6 +211,11 @@ class MUIDataTable extends React.Component {
       this.setHeadResizeable(this.headCellRefs, this.tableRef);
       this.updateDividers();
     }
+    this.updateOptions(this.props);
+  }
+
+  updateOptions(props) {
+    this.options = merge(this.options, props.options);
   }
 
   initializeTable(props) {
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -200,6 +200,7 @@ class MUIDataTable extends React.Component {
       this.setTableData(this.props, TABLE_LOAD.INITIAL, () => {
         this.setTableAction('propsUpdate');
       });
+      this.updateOptions(this.props);
     }
 
     if (this.props.options.searchText !== prevProps.options.searchText) {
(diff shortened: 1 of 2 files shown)


gabrielliwerant:
Hello. Can you describe what was broken before this PR and what this PR fixes with a specific example?

noobtiger:
So we have a custom toolbar setup which returns a react component and the react component props are passed from the parent component.
```
function ReactComponent(props: Props) {
 const [count, setCount] = React.useState(0);
  const options = {
    customToolbarSelect: () => {
      return (
        <div>
          <CustomComponent count={count} />
        </div>
      );
    },
  };
 return (
      <MuiDataTable
        {...props}
        options={options}
      />
 )
}
```

In the above example, initial load works fine, but if the count prop is updated, the value doesn't get reflected in the custom toolbar. 
I looked at the code, it looks like the customToolbarSelect function gets executed and it loses the closure chain, so the value change doesn't get reflected. In the previous commits, the options gets applied every time because initializeOptions is called on every update. This was removed in the current commit and initializeOptions is only called on initial load. 
The change here will merge the options on update.

gabrielliwerant:
@noobtiger Thank you, but there's still not enough for me to reproduce on my end.

I added 

```
  updateCount = () => {
    let { count } = this.state;
    count++;
    this.setState({ count });
  };

  ...

  <Tooltip title={"Count"}>
    <IconButton className={classes.iconButton} onClick={this.updateCount}>
      <CompareArrowsIcon className={classes.icon} />
    </IconButton>
  </Tooltip>
```

and can confirm that `count` is updated. Maybe a codesandbox example would help?

Beamanator:
I'm working on a project where we try to update the `MUIDataTable` options during a state change. The changed options do not get reflected in the table. Here is a codesandox my co-worker created to demonstrate: https://codesandbox.io/embed/eloquent-chaplygin-fb0y3

As you can see, clicking the button at the top toggles `isDownloadable` in state, and `getTableOptions` is being called on component re-render, but the download icon in the table is not being removed & added upon `options` update.

gabrielliwerant:
@Beamanator Thank for the example and the explanation. Will review next time I get a chance.

PR #666, opened 2019-06-05
Fix table toolbar select height

rcfish:
The table toolbar select when displayed expands to 82px in height (vs the 64px for the standard toolbar). This has the effect that the overall table height is changed and users see the table data shift up or down. Now the table toolbar select remains at 64px and there is no data shift up or down.


src/components/TableToolbarSelect.js
```diff
@@ -12,26 +12,18 @@ const defaultToolbarSelectStyles = theme => ({
...
     position: 'relative',
     zIndex: 120,
     justifyContent: 'space-between',
+    alignItems: 'center',
     paddingTop: theme.spacing.unit,
     paddingBottom: theme.spacing.unit,
   },
   title: {
     paddingLeft: '26px',
-    top: '50%',
```
gabrielliwerant: I think we might need these styles under `title`. If you take a look at this with the `customize-toolbarselect` example, where there are multiple icons in a custom toolbar, you'll see that they don't line up without them.

rcfish: Thanks. Will check it out and update.

rcfish: @gabrielliwerant I don't think we have an issue here. The `customize-toolbarselect` example has the same `iconButton` styling as the default component which is the styling I removed. Removing it from the example and the icons all line up with the title.

rcfish: image

gabrielliwerant: Lets remove that styling from the example then, so that it is not broken.

rcfish: Example updated

gabrielliwerant: Thank you!

rcfish:
I'm happy to revert back the webpack file, wasn't sure what the protocol was for this.

PR #670, opened 2019-06-06
Add table properties option

nathancharles:
With the release of `material-ui@4.0.0` the [table component](https://material-ui.com/api/table/#props) has new properties. Currently there is no way to pass any properties to the material-ui table component, as is supported for rows and cells.

This PR introduces a new table options property `tableProps`. `tableProps` is an object and is applied directly to the material-ui table component


gabrielliwerant:
Hm, wondering if there's a better API to include these options. We need to be careful about adding an option that allows for so many different things to be added as it could be a pandora's box, and it will be difficult to reign it in later as people may begin to rely on it.

I'm guessing that part of the issue is allowing options that are in material v4 without creating options that will break in material v3, right?

gabrielliwerant:
@nathancharles I'm open to adding this, but a couple of things:

1. A better name for the option, for now, would be `setTableProps`, as this keeps us consistent with similar features.
1. We should allow `classNames` to be sent in with this option, which we would need to check for and then add to the default `className`s.

nathancharles:
Good call on the option name, all for keeping consistency! I included the default className and added a test

Issue #542, opened 2019-04-05
Download CSV downloads all data instead of just current page data in server side mode


ssword:
In server-side mode, download CSV will only download data that is available to render the current page. It is possible to use a different source for CSV creation so that we can provide full data in the CSV file. 

## Expected Behavior

Given an option to provide custom data while Creating CSV. 
## Current Behavior

Only use data that is queried for current page rendering. 

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    "@material-ui/core": "^3.9.0"     |
| MUI-datatables  |     "mui-datatables": "^2.0.0-beta-1"    |
| React        |   "react": "^16.6.0"      |
| browser      |    Chrome     |
| etc          |         |

gregnb:
hey @ssword what would you propose in terms of changes? Let's see some examples in terms of API changes

ssword:
@gregnb,  thanks for answering my issue. I have checked the code, perhaps we can add another option in `downloadOptions` called `customCSVdata`. 
Then when `serverside` option is true, and `customCSVdata` is not null, we will replace `data` with `customCSVdata`. 

````javascript
    const options = {
      filter: false,
      search: false,
      count: rowCount,
      serverSide: true,
      downloadOptions: { 
          filename: 'eventTable.csv',
          // new API change added here
          customCSVdata: fullServersideData,
      },
      rowsPerPage: rowsPerPage,
      rowsPerPageOptions: [10, 20, 50],
      selectableRows: false,
      elevation: 0,
      },
    };
````

**Comment by Avd6977** [comment]
I think instead you could just add an option that skips the code that runs after options.onDownload. If the API (server-side) downloads a file, no need for the rest of that code to run and download another file. Otherwise, if you just want to pass the data through to the table, you could use onDownload.

**Comment by ronaiza-cardoso** [comment]
How can I download external CVS files?

**Comment by thbl** [comment]
+1
Does any one have an example on how to download the csv server side, instead of just the first page in table ?
@ssword @gabrielliwerant @Avd6977 @gregnb

**Comment by patorjk** [comment]
@thbl - @Avd6977  is right, this is how I do it:

```
onDownload: (buildHead, buildBody, columns, data) => {
    props.download_from_server(); // this makes a REST call to the server and downloads the data
    return false;
}.
```

I used a library called [downloadjs](https://www.npmjs.com/package/downloadjs) to download the results from my call. That library is a little old though, not sure if there are any newer solutions.

[...]

PR #674, opened 2019-06-06
Fix for Issue #542

Avd6977:
Skip downloading the file if it is downloaded server side.

### DIFF
--- a/examples/csv-export/index.js
+++ b/examples/csv-export/index.js
@@ -1,55 +1,58 @@
-import React from "react";
-import ReactDOM from "react-dom";
-import MUIDataTable from "../../src/";
+import Button from '@material-ui/core/Button';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import MUIDataTable from '../../src/';
 
 class Example extends React.Component {
+  state = {
+    downloadFile: true,
+  };
 
   render() {
-
     const columns = [
-      "Name",
-      "Title",
-      "Location",
-      "Age",
+      'Name',
+      'Title',
+      'Location',
+      'Age',
       {
-          name: "Salary",
-          options: {
-              download: false,
-          }
+        name: 'Salary',
+        options: {
+          download: false,
+        },
       },
     ];
 
     const data = [
-      ["Gabby George", "Business Analyst", "Minneapolis", 30, 100000],
-      ["Aiden Lloyd", "Business Consultant", "Dallas",  55, 200000],
-      ["Jaden Collins", "Attorney", "Santa Ana", 27, 500000],
-      ["Franky Rees", "Business Analyst", "St. Petersburg", 22, 50000],
-      ["Aaren Rose", "Business Consultant", "Toledo", 28, 75000],
-      ["Blake Duncan", "Business Management Analyst", "San Diego", 65, 94000],
-      ["Frankie Parry", "Agency Legal Counsel", "Jacksonville", 71, 210000],
-      ["Lane Wilson", "Commercial Specialist", "Omaha", 19, 65000],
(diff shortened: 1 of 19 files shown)


Avd6977:
@gabrielliwerant Any chance this PR could make the version bump with the custom filters?

gabrielliwerant:
@Avd6977 I don't think so, but I can do a smaller release after this next one for some quick fixes.

gabrielliwerant:
Does this PR actually answer the #542 issue? Looks like that user was referring to an issue where only the current page is downloaded with serverside, not extra downloads.

Avd6977:
That'd be awesome, thank you!

It provides one possible answer to it, which is to just download their own file. The other answer is to get all of the data serverside using "onDownload", which is existing functionality.

Avd6977:
@gabrielliwerant Just curious when you think that smaller release may be.

gabrielliwerant:
This feature probably won't make next release, in case you were wondering. I need some time to think about it, and how it may interact with an earlier PR that is also vying to change download options.  There are also some more critical things that should take priority.

Avd6977:
Ok, would you prefer it be in a different manner? Our use case is that the entire CSV is downloaded server-side but we'd like to still use the download button.

gabrielliwerant:
I'm reluctant to extend the API for specialized use cases unless there's a strong case for it and it can't be done in another manner. What if instead the `onDownload` callback were enabled to skip downloading if it returned `false`? This way, we don't need a new option, and it can be utilized in a wider variety of contexts.

Avd6977:
I'm good with that. Committed the code change. Let me know if that looks like what you were talking about.

test/MUIDataTableToolbar.test.js
```diff
@@ -241,4 +240,30 @@ describe('<TableToolbar />', function() {
 
     assert.strictEqual(onDownload.callCount, 1);
   });
+
+  it('should trigger onDownload prop callback when calling method handleCSVDownload and skip download if return from onDownload is false', () => {
```
gabrielliwerant: This test doesn't prove that the download is skipped, so I'm not sure it's doing anything for us that the other test for `onDownload` isn't already doing. I think that the way the CSV creation works at the moment, it's difficult to determine in a test whether or not the download was successful (just whether or not the callback fired). To do so, we'd probably need to expose another callback that fires after successful download and returns the data, but this is outside the scope of this PR, and I'm not sure we need it at this point.

Long story short, I appreciate the attempt to add to our coverage, but I think it would better to remove this test for now.

Instead, if you want, we could add to the `csv-export` example a conditional that would trigger this option and prevent download, so at least we'd have a built in way to verify in the codebase. If you don't want to do that, that's fine, I can add it after this gets merged in, before the next release.

Avd6977: Removed the test and added `onDownload: () => false` to the example!

examples/csv-export/index.js
```diff
@@ -62,6 +62,7 @@ class Example extends React.Component {
           filename: 'excel-format.csv',
           separator: ';',
       },
+      onDownload: () => false,
```
gabrielliwerant: So, the only issue with this is that it defaults to off for people testing csv export functionality in the example. I was thinking something like this might be more instructive:

```
state = { isOnDownload: true };

...

const options = { onDownload: () => this.state.isOnDownload };

...

<button onClick={ () => this.setState({ isOnDownload: false }) }>Shut off download functionality</button>
```

Does that make sense?

Avd6977: Yep, makes sense. Updated and pushed!

gabrielliwerant: Thank you!

Issue #657, opened 2019-05-31
trouble maintaining columns filterList state after re-rendering MuiDatables


ericpyle:
As far as I can tell, since version 2.1.0, MuiDatatables is having trouble maintaining `filterList` state during re-rendering when `columns[].options` maintain their own `filterList`. [Version 2.0.0 seems to work okay](https://codesandbox.io/embed/muidatatables-change-columns-onfilterchange-200-working-fzb36).

## Expected Behavior

When configuring columns with filterList, re-rendering the MuiDatables component should respect columns' filterList, if columns filterList has changed since table initialization.

## Current Behavior

After initializing a column with a filterList, then using the UI to remove the filter (thus emitting `filterChanged`) results in posting the filterList to higher order component's state, which then propagates back down to rebuild the initial column's filterList. This triggers the MuiDatatable to re-render, but it reverts back to the original filterList rather than 1) the new filterList being passed in through columns or 2) its most recent state as reflected in the `filterChanged` event. (Effectively undoing the filter change event).

## Steps to Reproduce (for bugs)

1. configure columns with non-empty filterList
2. handle onFilterChange to change the state of the current columns filterList

I've simulated this unexpected behavior here (try clicking on the `x` for `Franky Miles` to remove that filter. Notice the messages in the console):
https://codesandbox.io/embed/muidatatables-change-columns-onfilterchange-ybhyv

![changeColumns-onFilterChange](https://user-images.githubusercontent.com/1125565/58734762-f7be1d00-83bd-11e9-8cbd-63f9677a2be8.gif)

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  3.9.2    |
| MUI-datatables  |  2.3.0   |
| React        |         |
| browser      |         |
| etc          |         |

ericpyle:
By the way, I had thought that one way to avoid this was to not change the initial columns options filterList, and so let MuiDatatables manage its own state. However, whenever I changed the data (e.g. added a new row to the table), the filterList would also revert back to the initial columns options filterList.

ericpyle:
@gabrielliwerant any idea how hard this is to fix? I'm kind of blocked since this would be a regression from the current behavior of my software. I tried going back to version 2.0.0 which doesn't have this issue, but that version doesn't have `searchText` which would be another regression if I took it out.

gabrielliwerant:
[...]

PR #678, opened 2019-06-06
fix for issue 657

ericpyle:
fix for issue #657 "trouble maintaining columns filterList state after re-rendering MuiDatables"

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -332,7 +332,7 @@ class MUIDataTable extends React.Component {
     let filterData = [];
     let filterList = [];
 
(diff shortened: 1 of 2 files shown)


ericpyle:
@gabrielliwerant  okay, I can see that my fix does not respect the model differences between props.columns and state.columns. I'm beginning to understand the code better. so I'll go back to the drawing board

gabrielliwerant:
@ericpyle Sounds like you might be about to fall down the same rabbit hole I did a few weeks ago when I discovered the core issue! I think a major refactor is needed to solve many of the outstanding issues, which I'm working on, but I'm open to hacks that will keep v2 moving along in the mean time.

PR #682, opened 2019-06-11
Added example of server side simulation, changing the count

VagnerNico:
Used the xhrRequest derived from the serverside-pagination example it works like a fake request to the server.
I pass the tableState simulating query_params sent to the server like page, length and even filter values. So the server returns an array of data with length equals to rowsPerPage from tableState and the count value for the current query.
The filters work in the same way returning the array of filtered data and the new count.

### DIFF
--- a/examples/serverside-simulation/index.js
+++ b/examples/serverside-simulation/index.js
@@ -1,0 +1,327 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import { CircularProgress } from '@material-ui/core';
+import MUIDataTable from "../../src/";
+
+class Example extends React.Component {
+
+  state = {
+    page: 0,
+    count: 0,
+    data: [["Loading Data..."]],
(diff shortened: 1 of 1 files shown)


gabrielliwerant:
So, as you noted, we do already have the `serverside-pagination` example which contains a sample async function that changes the `count` property, among other things. Is there something happening in this example that is fundamentally different, or that we can't just add to the existing example?

VagnerNico:
I guess they are similar, the main difference is that I'm maintaining the data fixed (original example randomly changes data), besides the filters are sent to the server changing data and the count. So count will change when the user filters the data and not when he changes the page.

gabrielliwerant:
Could we add to the existing `serverside-pagination` example the additional updates on filter changes? Would be nice to keep similar use cases in the same place for ease of experimentation and demonstration.

Issue #370, opened 2019-01-16
Uncaught Error: Provided options.page of `6` is greater than the total available page length of `1`


amrithds:
Fatal error is generated and page stops rendering when the page number is greater than 1 and the filter is applied which results in less number of pages than the current options.page value. 

mui-datatables Version : 2.0.0-beta-52

## Expected Behavior

The expected behavior is the page should reset to 1 or to the last page of total available page.

## Current Behavior
Fatal error
Uncaught Error: Provided options.page of `6` is greater than the total available page length of `1`

## Steps to Reproduce (for bugs)
1. Unset all filters
2. Go to page 5 (Greater than 1)
3. Apply select filters on the column so that result is having 1 page
4. Check the console

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.0.0     |
| MUI-datatables  |    2.0.0-beta-52     |
| React        |    16.3.0     |
| browser      |    chrome     |

dhanalakshmimuthusamy:
HI
I also got the same error while applying the filtering for the MUI datatables .Is there there any fix for that?

mui-datatables Version : 2.0.0-beta-54

## Environment
Tech	Version
Material-UI	3.9.0
MUI-datatables	2.0.0-beta-54
React	16.7
browser	chrome

vmf91:
I also got the same error with the current environment:

| Tech         | Version |
|--------------|---------|
| Material-UI  |     3.1.0    |
| MUI-datatables  |   2.0.0-beta.56      |
| React        |    16.5.2     |
| browser      |    Chrome     |

ericpyle:
I got this as well :( with 

material-ui|3.9.2
mui-datatables | 2.0.0-beta.58
react | 16.3.0
browser | chrome

ericpyle:
by the way, I haven't actually been able to reproduce this 100% of the time. Sometimes I can seemingly follow the same steps and it doesn't crash. @vmf91 @dhanalakshmimuthusamy @amrithds do you know the exact minimal steps needed to reproduce this crash reliably? Or might this be a race condition?

I would like to test this workaround but until I can find a relative path, I am not sure it works.

>   componentDidCatch(error, info) {
>     // Display fallback UI
>     this.setState({ page: 0 });
>   }
> 

in render()

> const options = {
>      ...
>       page: this.state.page
>     };
>     return (
>       <Paper className={classes.root}>
>         <MUIDataTable
>           ...
>           options={options}
>         />
[...]

PR #685, opened 2019-06-11
Fix issue #370

QuentinChevance:


src/MUIDataTable.js
```diff
@@ -852,6 +854,7 @@ class MUIDataTable extends React.Component {
         }
 
         return {
+          page: 0,
```
QuentinChevance: I added this change to set the current page to 0 when a filter occurs.

gabrielliwerant:
Could you pull in latest `master` here? It has the prettifier changes, so I think that will clean up the diffs a bit.

QuentinChevance:
> Could you pull in latest `master` here? It has the prettifier changes, so I think that will clean up the diffs a bit.
Done :+1:

PR #687, opened 2019-06-12
Column options changing

VagnerNico:
Now you can control options outside the table, in the example added the filterOptions and filterList are changed outside the table.

### DIFF
--- a/examples/column-options-update/index.js
+++ b/examples/column-options-update/index.js
@@ -1,0 +1,139 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import MUIDataTable from "../../src/";
+
+class Example extends React.Component {
+
+  state = {
+    filterList: [
+      ['Franky Miles'],
+      ['Business Analyst'],
+      [],
+      [],
+      []
+    ],
+    filterOptions: ['this', 'test', 'is', 'working'],
+  }
+
+  handleFilterOptionsChange = (event) => {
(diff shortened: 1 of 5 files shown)


VagnerNico:
Coverage decreased because I inserted a new boolean condition here:
```javascript
if (this.state.columns.length && isEqual(this.rawColumns(newColumns), this.rawColumns(this.props.columns))  && !Object.keys(newColumns).some(val => this.state.columns[val] !== newColumns[val])) {
    const { columns, filterList, filterData } = this.state;
    return { columns, filterList, filterData };
}
```
I will try to write some test to raise up the coverage.

VagnerNico:
tableChanging
Created a giant gif to show the example.

VagnerNico:
By the urgency, I built locally and changed in my application, until now I didn't find any breaking with the other tables (not server-side).

gabrielliwerant:
I believe this was fixed in `2.5.1`. Can you verify that is the case? If so, it still might good to have your example and test if you wouldn't mind paring this down to those changes only.

VagnerNico:
Fixed in 2.5.1, works like a charm in my application.
I will update my local branch maintaining only the example and commit again.

gabrielliwerant:
Continued here https://github.com/gregnb/mui-datatables/pull/848.

gabrielliwerant:
Closed via https://github.com/gregnb/mui-datatables/pull/848.

Issue #689, opened 2019-06-12
preserve filterList state across data prop changes


ericpyle:
## Expected Behavior

After having initialized the columns filterList, data props changes should not undo any subsequent filterList changes via UI.

## Current Behavior

In the context of having initialized mui-datatables with a columns filterList, data props changes undo any subsequent changes to filterList via UI

## Steps to Reproduce (for bugs)

1. You can see an example of how changing the data after changing the filter will just revert back to the earlier filter. Try clicking on the x to delete either of the filters:
2. https://codesandbox.io/s/muidatatables-change-data-after-onfilterchange-x33ok

![changeData-onFilterChange](https://user-images.githubusercontent.com/1125565/59371363-72663100-8d0a-11e9-8018-f339368b0fd6.gif)

new failing test\MUIDataTable.test.js
```
  it('should not update internal column state on data props change after initial columns options filterList', () => {
    const initColumns = cloneDeep(columns);
    initColumns[0].options.filterList = ['Joe James'];
    const mountShallowWrapper = mount(shallow(<MUIDataTable columns={initColumns} data={data} />).get(0));
    const instance = mountShallowWrapper.instance();
    // now disable Joe James filter
    instance.filterUpdate(0, 'Joe James', 'Name', 'dropdown');
    mountShallowWrapper.update();
    mountShallowWrapper.setProps({ data: [] });
    mountShallowWrapper.update();
    // mimic componentDidUpdate(prevProps) {
    // if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
    // instance.setTableData(mountShallowWrapper.props(), 1 /* instance.TABLE_LOAD.INITIAL */);

    const updatedState = mountShallowWrapper.state();
    const { columns: updatedColumns } = updatedState;
    // make sure we still no longer have an active filter
    assert.deepEqual(updatedState.filterList, [[], [], [], [], []]);
  });
```

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  | 2.3.0        |
| React        |         |
| browser      |         |
| etc          |         |

VagnerNico:
The problem resides on the App rerender when u change the data, the filter value that was set in the const columns will be constant, so u need to control the filters from the table wrapper component, that is, receive the filterList and setState them spreading for your columns.
Today even if you make this changes the filters will not work, maybe the PR #687 could solve the problem.
The code look like this:
```JavaScript
import React from "react";
[...]

PR #690, opened 2019-06-12
Debug data props change

ericpyle:
For issue #689


src/MUIDataTable.js
```diff
@@ -332,7 +332,7 @@ class MUIDataTable extends React.Component {
     let filterData = [];
     let filterList = [];
 
-    if (this.state.columns.length && isEqual(this.rawColumns(newColumns), this.rawColumns(this.props.columns))) {
+    if (this.state.columns.length) {
```
ericpyle: As far as I can tell the isEqual code is not needed. All the current calling contexts pass in `this.props.columns`

src/MUIDataTable.js
```diff
@@ -368,7 +368,7 @@ class MUIDataTable extends React.Component {
       columnData.push(columnOptions);
 
       filterData[colIndex] = [];
-      filterList[colIndex] = [];
+      filterList[colIndex] = columnOptions.filterList || [];
```
ericpyle: buildColumns should return filterList based on the initial columnOptions filterList

src/MUIDataTable.js
```diff
@@ -435,10 +435,6 @@ class MUIDataTable extends React.Component {
         filterData[colIndex] = cloneDeep(column.filterOptions);
       }
 
-      if (column.filterList) {
```
ericpyle: this code is what was overwriting any changes to state.filterList

ericpyle:
@gabrielliwerant see any issues with this?

ericpyle:
@gabrielliwerant superseded by fix for https://github.com/gregnb/mui-datatables/pull/711

The only caveat is that the client must keep track of filterList changes and propagate them back down through column props in order to not lose UI based filter state changes.

PR #692, opened 2019-06-12
fixes "Uncaught TypeError: prevState.selectedRows is not iterable"

ericpyle:
```
MUIDataTable.js:952 Uncaught TypeError: prevState.selectedRows is not iterable
    at MUIDataTable.setState.prevState (MUIDataTable.js:952)
```

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -949,7 +949,7 @@ class MUIDataTable extends React.Component {
             return arr;
           }, []);
 
-          let newRows = [...prevState.selectedRows, ...selectedRows];
+          let newRows = [...prevState.selectedRows.data, ...selectedRows];
           let selectedMap = buildMap(newRows);
 
           if (isDeselect) {


gabrielliwerant:
@ericpyle Can you provide me with an example or steps to cause the breakage so that I can verify the fix on my end?

patorjk:
I can verify this, though the fix should be:

`let newRows = [...selectedRows];`

If you do:

`let newRows = [...prevState.selectedRows.data, ...selectedRows];`

The "selectedRows" state object will have duplicate information.

You see this error if you upgrade babel and run the project. "prevState.selectedRows" is an object and is not iterable (it's equal to {data: stuff, lookup: [stuff]}). In the current setup, it looks like babel just ignores non-iterable objects so this line has no ill effect. Most likely this is a copy-and-paste bug (since a similar line is found a little further down). Since it has no ill effect currently, it has gone unnoticed.

Issue #438, opened 2019-02-22
Accessing event property inside onRowClick


iKrushYou:
Is it possible to provide access to the event property while inside the `onRowClick` function?

I am trying to access the event's `ctrlKey` and `metaKey` properties so that I can enable opening links in a new tab just as normal hyperlinks would function. I can do so using the regular onClick function seen [here](https://codesandbox.io/s/xozjww4xvq)

gabrielliwerant:
I believe it is already accessible if you add it as the 3rd argument:

```
onRowClick = (rowData, rowMeta, e) => {
    const { ctrlKey, metaKey } = e;
    console.log(ctrlKey, metaKey);
}
```

iKrushYou:
@gabrielliwerant honestly i hate myself for not trying this before posting... but I had read the source [here](https://github.com/gregnb/mui-datatables/blob/671461ca92d45ebe41f33a077cf43a9225fb1b70/src/components/TableBody.js) and didn't see event `onClick={options.onRowClick ? options.onRowClick.bind(null, row, { rowIndex, dataIndex }) : null}`

but I guess I don't quite 100% understand how functions work in ES6 yet

gabrielliwerant:
Haha, no worries, it can be confusing to know which arguments are getting passed along!

jdfx:
This doesn't work with typescript, the type doesnt include the event.

Issue #409, opened 2019-02-08
Table Height overflowing when 10 rows are showing


iKrushYou:
Is there a way to get the height of the table to match up with the height of all the rows? It seems like when displaying 10 rows, the last one gets cut off and there's a slight bit of scrolling

https://codesandbox.io/s/k22kzo8zzv

gabrielliwerant:
If you use responsive scroll, you can use the following to set a custom height:

```
const myTheme = createMuiTheme({
  overrides: {
    MUIDataTable: {
      responsiveScroll: {
        maxHeight: '580px'
      }
    }
});

<MuiThemeProvider theme={myTheme}>
      <MUIDataTable ... />
</MuiThemeProvider>
```

Bear in mind that if you have a horizontal scroll bar, it will take up some space at the bottom of the table, so you may want to test that condition to determine what your optimal table height should be.

iKrushYou:
@gabrielliwerant Thanks for the response. I can confirm that this works [here](https://codesandbox.io/s/6yjmqv2xk3)

Will this be the accepted solution or is this just currently a workaround?

iKrushYou:
Also, @gabrielliwerant i know this is out of place, but do you know off-hand how to override hovering over the row to show the cursor hand and /or overriding the active state to highlight darker when being pressed? (I can open a separate issue if need be)

gabrielliwerant:
@iKrushYou I'm unsure of the exact intent, so I don't know whether or not my answer is the accepted solution. But as to your other question, here you go!

```
MuiTableRow: {
      root: {
        cursor: 'pointer'
      },
      hover: {
        '&$root': {
          '&:hover': {
            background: 'gray'
          }
        }
      }
    }
```

iKrushYou:
thanks @gabrielliwerant 
that worked. Of course, it had to be on the `MuiTableRow` and not on the `MUIDataTableRow` as I was trying... 

My intention was to highlight on hover and highlight darker when pressing. Luckily, i was able to change the style to `&:hover:active` and it worked as exptected

JReinhold:
As of [V2.10.0](https://github.com/gregnb/mui-datatables/tree/2.10.0) it is now possible to use the prop `responsive="scrollFullHeight"` to make the table full height and a scroll behavior at the same time. That should fix the issue.

See https://github.com/gregnb/mui-datatables/pull/867

PR #703, opened 2019-06-18
Feature/row select

iKrushYou:
Implemented the option `rowSelect` to handle case in [comment](https://github.com/gregnb/mui-datatables/issues/409#issuecomment-467617050) of #409 

`rowSelect` default is `false`. When `true`, the row will cause the mouse to change to `pointer` to indicate that the row is selectable (an `onClick` will fire)

Screen Shot 2019-06-17 at 10 02 50 PM

And, when the user clicks on the row, it will highlight a different color to indicate that the row is being pressed like a button.

Screen Shot 2019-06-17 at 10 03 19 PM

The reason for this solution instead of just overriding the theme as suggested [here](https://github.com/gregnb/mui-datatables/issues/409#issuecomment-469428528) is my usecase involves many different tables across the application -- some of which have clickable rows and some which do not. So this seemed like the best solution overall.

This also includes an update to the documentation to include `event` as a parameter in the `onRowClick` function as mentioned in issue #438 

Will close PR #693


iKrushYou:
Closed because I didn't realize i was on a suuuuper old build

PR #704, opened 2019-06-18
Feature/row select

iKrushYou:
Implemented the option `rowSelect` to handle case in [comment](https://github.com/gregnb/mui-datatables/issues/409#issuecomment-467617050) of #409 

`rowSelect` default is `false`. When `true`, the row will cause the mouse to change to `pointer` to indicate that the row is selectable (an `onClick` will fire)

Screen Shot 2019-06-17 at 10 02 50 PM

And, when the user clicks on the row, it will highlight a different color to indicate that the row is being pressed like a button.

Screen Shot 2019-06-17 at 10 03 19 PM

The reason for this solution instead of just overriding the theme as suggested [here](https://github.com/gregnb/mui-datatables/issues/409#issuecomment-469428528) is my usecase involves many different tables across the application -- some of which have clickable rows and some which do not. So this seemed like the best solution overall.


iKrushYou:
Hey @gabrielliwerant how do I assign to someone? I think that's the procedure right?

gabrielliwerant:
At the moment, I'm the only one that can really review these, and I tend to assign myself with whatever is next of my list.

To be honest, this PR would be low priority for me and I'm actually not sure I agree a change to the library is needed here. As you've already mentioned, this behavior is already customizable with the options available to you in the library. It's not really a good practice to add something to library being used by many people just because one individual has a very specific use case. I have to consider the long term implications on maintaining the library and the growing complexity of the API, and measure that against the utility being provided. For that reason, I'm reluctant to add features that are already possible through individual customization.

Another thing you can do, if it's really important to you to maintain this feature for your own code, is to use your own fork in your projects. You will need to keep it up to date yourself if you want the latest and greatest from this library as it moves forward, but otherwise, there's nothing stopping you from implementing what you need without it being included in the API of the main project.

gabrielliwerant:
The option to select on the row instead of the checkbox (or expand icon for expandable rows) was added to this release: https://github.com/gregnb/mui-datatables/releases/tag/2.4.0. When these options are `true`, the cursor will change to indicate it can be clicked. Does this help with your use case?

patorjk:
What described here is mostly available in the table right now with the selectableRowsOnClick option, and will fully be available in the v3 release. What's currently missing: I noticed the pointer icon still appears on rows that aren't selectable (fixed in v3) and there isn't an option to hide the checkboxes (this will be added in v3 with a selectableRowsHideCheckboxes option). Closing for now.

Issue #688, opened 2019-06-12
Custom filter ranges for numbers and dates


Avd6977:
## Expected Behavior
1. Filter Type can have new types called numberRange and dateRange
1. Number and Date Range can accept custom component tags to allow a custom date-picker
1. Two inputs appear for the column which are Start Date and End Date

## Current Behavior
Columns which have numbers or dates cannot be filtered to a user-input range with start and end. Possibly the new custom filter logic would allow a number range to work if the user types 1-10 and we parse the "-" out, but that's a lot of reliance on the user to do it properly. For date-ranges, a date picker would be preferable with two inputs.

Avd6977:
@gabrielliwerant I should've looked more into that branch, but I have some work on this if you'd like to provide some input.
https://github.com/Avd6977/mui-datatables/tree/DateRangeFilter

Additionally, it might be a better route to just do a custom filter which allows JSX to be passed in.

Avd6977:
@gabrielliwerant Any thoughts on this? Would definitely appreciate your input on how we would do something like this.

gabrielliwerant:
So, I agree that what we need is a generic custom filter render which will allow users to specify whatever components they wish instead of the default options. If you'd like to take a swing at that and open up a PR, I will gladly take a look when I get a chance.

gabrielliwerant:
Closing, as this feature should now be available in `2.6.0` via `filterOptions: { display: Component }` and `fitlerType: 'custom'`, thanks to @Avd6977!

PR #708, opened 2019-06-19
Custom filter - #688

Avd6977:


Avd6977:
@gabrielliwerant Here's a PR for the ability to add a custom filter. I made a number range component as the test

Avd6977:
Adding #688 link

examples/range-filters/index.js
```diff
@@ -0,0 +1,155 @@
...
+                return age < filters['min'] || age > filters['max'];
+              } else if (filters['min']) {
+                return age < filters['min'];
+              } else if (filters['max']) {
+                return age > filters['max'];
+              }
+              return false;
+            },
+            display: (filterList, onChange, index, column) => (
+              <div>
```
gabrielliwerant: To help keep the proliferation of new examples to the minimum needed, I'd like to see this added instead to the existing `customize-filters` example.

Also, if we could add

```
<div>
  <FormLabel>Age</FormLabel>
  <FormGroup row>
    // extra fields here
  </FormGroup>
</div>
```

the form group and label will help make it more clear what is happening in the example.

And lastly, we'll need to remove the `cellHeight` property from here https://github.com/gregnb/mui-datatables/blob/bbb1a27e27c1072e0897ada3fa4433a544f83e43/src/components/TableFilter.js#L311 to prevent the display from being truncated as will be possible now with some custom fields.

src/components/TableFilter.js
```diff
@@ -264,6 +268,19 @@ class TableFilter extends React.Component {
     );
   }
 
+  renderCustomField(column, index) {
+    const { classes, filterList, options } = this.props;
+    const display = (column.filterOptions && column.filterOptions.display) || options.filterOptions.display;
```
gabrielliwerant: This will result in an error when the type is set to `'custom'` but there is no `display` property defined. I think we should be kind here and instead display an empty field with a console error that explains that the `display` property must be set if using a `'cusomt'` filter type.

src/components/TableFilterList.js
```diff
@@ -36,16 +39,27 @@ class TableFilterList extends React.Component {
...
     return (
       <div className={classes.root}>
-        {filterList.map((item, index) =>
-          item.map((data, colIndex) => (
+        {filterList.map((item, index) => {
+          if (columnNames[index].filterType === 'custom' && filterListRenderers[index](item)) {
+            return (
+              <Chip
+                label={filterListRenderers[index](item)}
+                key={0}
```
gabrielliwerant: Couldn't we still use the `colIndex` for the `key` here?

test/MUIDataTableFilter.test.js
```diff
@@ -18,7 +18,30 @@ describe('<TableFilter />', function() {
...
+        options: {
+          filter: true,
+          filterType: 'custom',
+          customFilterListRender: v => `City: ${v}`,
+          filterOptions: {
+            names: [],
+            logic(city, filters) {
+              return false;
+            },
+            display: (filterList, onChange, index, column) => (
```
gabrielliwerant: This is a good start, but was there a plan to provide an actual test case for this option? If not, I can work on one once this PR is merged.

examples/range-filters/index.js
Avd6977: Done

src/components/TableFilter.js
Avd6977: Added a message, feel free to tweak the wording if you want!

src/components/TableFilterList.js
Avd6977: Changed to just use index

test/MUIDataTableFilter.test.js
Avd6977: Would you terribly mind writing the test? Or giving some feedback on what you may want in the test?

gabrielliwerant: No problem, I tend to add what I can after merges but before releases. They're not required for the PRs, but are nice to have. Once the next release is up, you should be able to find the commit for it in the project if you want to look at it. Not sure myself how I'll tackle it yet!

src/components/TableFilter.js
```diff
@@ -264,6 +268,24 @@ class TableFilter extends React.Component {
...
+    const display =
+      (column.filterOptions && column.filterOptions.display) ||
+      (options.filterOptions && options.filterOptions.display);
+
+    return (
+      <GridListTile key={index} cols={1}>
+        <div className={classes.textFieldRoot}>
+          <FormControl key={index}>
+            {(display && display(filterList, this.handleCustomChange, index, column)) ||
+              "Property 'display' is required when using custom filter type."}
```
gabrielliwerant: Oops, sorry, one more thing. What I was actually thinking here was to not have anything display, and instead show a warning to this effect in the console.

My reasoning is that I don't want the UI to be given a default means of displaying errors/warning information, as I think that should be up to the implementation.

Avd6977: Oh, my fault for misunderstanding that. Pushed changes to just be console log and show nothing when display isn't passed in

src/components/TableFilter.js
```diff
@@ -274,13 +274,15 @@ class TableFilter extends React.Component {
       (column.filterOptions && column.filterOptions.display) ||
       (options.filterOptions && options.filterOptions.display);
 
+    if (!display) {
+      console.log('Property "display" is required when using custom filter type.');
```
gabrielliwerant: Can we use `console.error` here so that it will be more clear to the devs that they forgot something?

Avd6977: Updated!

test/MUIDataTableFilter.test.js
gabrielliwerant: FYI, here is the PR with the test, in case you were still interested: https://github.com/gregnb/mui-datatables/pull/736.

Avd6977: Oh, quite simple really! Thanks for adding that :)

gabrielliwerant: Yep, this one wasn't too bad, just needed to test that the display actually worked.

PR #711, opened 2019-06-21
Allow derived column state to be rebuilt on prop change

gabrielliwerant:
Attempt at a quick fix for https://github.com/gregnb/mui-datatables/issues/643.

There is at least one open PR that also incorporates something similar, but there's more going on, and I've been suspicious that this condition is needed for a while, so I want to try this as a hotfix and get community feedback incorporating other changes.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -351,11 +351,6 @@ class MUIDataTable extends React.Component {
     let filterData = [];
     let filterList = [];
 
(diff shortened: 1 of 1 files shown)


ericpyle:
@gabrielliwerant fwiw (as you may have already guessed) this approach will exhibit the issue #689 "preserve filterList state across data prop changes" (see failing test there) and conflicts with pull request #690.

I will try to workaround this by pushing the filterList changes to app state and updating the columns props more in line with my first approach in #657 and get back with you

PR #715, opened 2019-06-24
add test for #711

ericpyle:
passing test for #711 (originally from my issue #657)

### DIFF
--- a/test/MUIDataTable.test.js
+++ b/test/MUIDataTable.test.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { spy } from 'sinon';
 import { mount, shallow } from 'enzyme';
 import { assert, expect } from 'chai';
+import cloneDeep from 'lodash.clonedeep';
 import MUIDataTable from '../src/MUIDataTable';
 import TableFilterList from '../src/components/TableFilterList';
 import TablePagination from '../src/components/TablePagination';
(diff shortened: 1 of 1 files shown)


test/MUIDataTable.test.js
```diff
@@ -429,6 +430,24 @@ describe('<MUIDataTable />', function() {
...
+  it('should apply columns prop change for filterList', () => {
+    const mountShallowWrapper = mount(shallow(<MUIDataTable columns={columns} data={data} />).get(0));
+    const instance = mountShallowWrapper.instance();
+    instance.initializeTable(mountShallowWrapper.props());
+    // now use updated columns props
+    const newColumns = cloneDeep(columns);
+    newColumns[0].options.filterList = ['Joe James'];
+    mountShallowWrapper.setProps({ columns: newColumns });
+    mountShallowWrapper.update();
+    // mimic componentDidUpdate(prevProps) {
```
gabrielliwerant: I think we can remove these two comments. My reasoning is that they will be brittle; if the actual code changes, the comments will be out of date, and could lead to confusion.

test/MUIDataTable.test.js
```diff
@@ -429,6 +430,24 @@ describe('<MUIDataTable />', function() {
...
+    // if (this.props.data !== prevProps.data || this.props.columns !== prevProps.columns) {
+    instance.setTableData(mountShallowWrapper.props(), 1 /* instance.TABLE_LOAD.INITIAL */);
+
+    const updatedState = mountShallowWrapper.state();
+    const { columns: updatedColumns } = updatedState;
```
gabrielliwerant: This line doesn't look like it's doing anything in the test, so I think we can remove it.

test/MUIDataTable.test.js
ericpyle: Sounds good to me.

test/MUIDataTable.test.js
ericpyle: You're right. Thanks.

ericpyle:
@gabrielliwerant I think I've made the changes you suggested

PR #717, opened 2019-06-25
Fixing page out of bounds issue

lounsbrough:
Recreation Steps to produce bug:
- Load table with several pages of data
- Navigate to the last page
- Remove a large portion of the data
- The table component will throw an error and wreck your site

Resolution:
When `props.data` changes, we added a check to make sure the current page in state is not beyond the last available page. If necessary, the current page is set to the new last page.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -205,16 +205,17 @@ class MUIDataTable extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
+    // When search text or data changes, we need to reset page
+    if (this.props.options.searchText !== prevProps.options.searchText ||
+        this.props.data !== prevProps.data) {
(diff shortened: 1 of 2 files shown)


gabrielliwerant:
I believe I have a simpler fix here: https://github.com/gregnb/mui-datatables/pull/727. I can put it in the next release and we can see if it solves your problem. If not, I will another look at what you have here.

lounsbrough:
Perfect, that looks sounds good to me.

lounsbrough:
I tested this in our app and it behaves much better. 👍

gabrielliwerant:
@lounsbrough The fix is in `2.6.0`. If it works for you, we can probably close this.

lounsbrough:
Yes, we've already pulled it in and we're good to go. Thanks!

PR #722, opened 2019-06-26
Add support for UTF-8 when downloading CSV

rinorbytyci:
This pull should add UTF-8 support for special / accented characters when downloading CSV files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "coverage:html": "cross-env NODE_ENV=test nyc check-coverage --lines 55 --reporter=html --reporter=text mocha --require babel-register test/*.js && nyc report --reporter=html",
     "prettier": "find src/ docs/ test/ -type f -name \"*.js\" ! -path \"*/.next/*\" | xargs prettier --write",
     "lint": "eslint src",
-    "build": "cross-env NODE_ENV=production npm run prettier && rollup -c",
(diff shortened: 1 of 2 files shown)


patorjk:
This can also be done with the onDownload method:

```
onDownload: (buildHead, buildBody, columns, data) => {
    return "\uFEFF" + buildHead(columns) + buildBody(data); 
} 
```

The problem with always adding the BOM in is that it's Excel specific, and the actual character will show up in some spreadsheet programs (ex: https://github.com/react-csv/react-csv/issues/89).

gabrielliwerant:
This library has intentionally steered away from providing specific types of exports as this can become a source of bloat. Specific implementations are left up to the devs and external libraries.

Perhaps a better use of this PR would be to add something to the documentation or to an example that points out how to work with downloads intended for Excel (or other programs).

patorjk:
Excel specific, so closing for now. I'll add an example in the documentation though.

jose1208:
// option
  downloadOptions: {
                filename: 'excel-format.csv',
                separator: ';',
                filterOptions: {
                    useDisplayedColumnsOnly: true,
                    useDisplayedRowsOnly: true,
                }
            },
//function         
 onDownload: (buildHead, buildBody, columns, data) => {
                component.pdfReportAutoTable(                              
                    columns,
                    data
                );
                return false;
            },

PR #726, opened 2019-06-27
Removed checkboxes on single selection and allowing disabling the selection toolbar

luca-saggese:


gabrielliwerant:
Hey @luca-saggese. It looks like there are other PRs for disabling the select toolbar. Can you separate out the removal of the checkboxes from the selection toolbar changes so that there's just a single set of changes for a single feature in here?

patorjk:
Disabling the select toolbar is currently available in the table and an option to hide the checkboxes will be added in v3. Closing this for now.

Issue #718, opened 2019-06-25
Reducing table data can cause current page to be out of bounds


lounsbrough:
We have opened a PR that should fix this bug: https://github.com/gregnb/mui-datatables/pull/717

![Kapture 2019-06-25 at 15 43 44](https://user-images.githubusercontent.com/8313853/60132037-5168fb80-9760-11e9-87bd-f5c096d0c47e.gif)

## Current Behavior

When I am on page 10 of a large table, and I modify the data in the table such that there are now only 9 pages, the component blows up.

## Steps to Reproduce (for bugs)

1. Load table with several pages of data
2. Navigate to the last page
3. Remove a large portion of the data
4. The table component will throw an error and wreck your site

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.5.1     |
| MUI-datatables  |     2.5.1    |
| React        |    16.8.4     |
| browser      |    Chrome     |

gabrielliwerant:
Unfortunately, these steps weren't specific enough for me to reproduce the issue. :(

I loaded an example from the site with several pages of data, set the data to only a single entry, and everything appears to work fine:
![capture](https://user-images.githubusercontent.com/1779564/60149259-14255d80-97a2-11e9-9b56-f6d31f3238ad.gif)

Can you provide a code sample in codesandbox which demonstrates the issue?

Thanks!

lounsbrough:
You got it: https://codesandbox.io/s/gifted-brattain-ty0j6

I made this as simple and concise as possible. You can navigate forward a couple pages, and then click the button to "Reduce data to 1 row" and you will notice a couple things:
- The pagination shows an invalid state, for example, it might be `31-1 of 1`, which is clearly invalid.
- The more concerning issue is the prop type error which you can see in the console. This causes higher order components to blow up, and seems like an issue in general.

This is what our PR is intended to fix.

Let me know if you want more clarification on what we're seeing, and thanks for your help with this!

gabrielliwerant:
So, in your bug report here, you listed your version of material ui as `3.5.1`, which is correct as the highest the library officially supports at the moment. But in the example in the codesandbox, your version is the latest. If I switch it down to `3.5.1`, there is no pagination issue, and again, everything seems to work fine.

Can you break this on `3.5.1` with an HOC?

lounsbrough:
[...]

PR #727, opened 2019-06-28
Issue warning instead of throwing an error

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/718.

We have a means to clean up the page out of bounds issue elsewhere, so this error is no longer needed and was causing breakage in certain circumstances.

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -55,13 +55,7 @@ class TableBody extends React.Component {
     const toIndex = Math.min(count, (page + 1) * rowsPerPage);
 
     if (page > totalPages && totalPages !== 0) {
-      throw new Error(
-        'Provided options.page of `' +
-          page +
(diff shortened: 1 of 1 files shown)


richard-szegh:
I still get an error message instead of a warning in my project. I am using MUI 4.2.0 and mui-datatables 2.6.4. Am I missing something?

gabrielliwerant:
If you are using `serverside: true`, you may not be protected from this error, as the onus would be on you to keep your settings up to date. Otherwise, I'm not sure, it may be a different issue. Would you be able to open up a new issue with steps to reproduce and a minimal example we can both run and verify in codesandbox?

richard-szegh:
I am not using serverside rendering. I just reproduced the error in codesandbox. Opening a new issue soon with the details and the link to the codesandbox.

Issue #737, opened 2019-07-02
Custom Filter doesn't accept filterList (needed for serverside)


Avd6977:
## Expected Behavior
Passing in filterList with the column will default the filter to a value. For server-side, this is how to keep the filter between loads.

## Current Behavior
deepClone clears out the array when it has string keys

## Steps to Reproduce (for bugs)
const customFilterList = [];
customFilterList["min"] = "20"
Pass customFilterList in as the column's filterList and see that it isn't loaded.

## Your Environment

| Tech         | Version |
|--------------|---------|
| MUI-datatables  |     2.6.0    |

gabrielliwerant:
Hm, I'm a little confused by this one. Maybe a more fleshed out example would help. This doesn't look like supported behavior. If you want to keep the `filterList` between loads, you should be able to update it as part of state change on your end. I don't see the need to pass in an object.

Avd6977:
I think saying it was needed for just serverside is confusing. If you take the change to the custom-filter example I made and revert the Object.assign, then it doesn't use the default filter because deepClone doesn't clone it to the new array.

gabrielliwerant:
I think what I'm saying is that if your problem is maintaining filter state between serverside updates, I don't believe you need to this the way you're doing it. I'm pretty sure there's a way to do this with existing functionality.

In general, the direction the table is moving is for you to manage more of the state yourself rather than rely on the internal derived state of the table, so I'm reluctant to go the opposite direction.

Avd6977:
From what I'm seeing happen though, the chip is disappearing despite maintaining the state on our end. Have you looked at the PR? The main issue is that deepClone has known issues (especially with stringed-key arrays) so the fix is just cloning with the Object.assign method.

Unless I'm missing what you're meaning by being able to do this already. I'm passing in the filterList prop with the column. Is there a better way to maintain the filter state and pass it down so the Chips appear still?

gabrielliwerant:
I'd like to verify that the issue is blocking people from being able to manage filter state, and I don't yet have that verification because I don't think it's being done in the recommended way.

```
const customFilterList = [];
customFilterList["min"] = "20"
Pass customFilterList in as the column's filterList and see that it isn't loaded.
```

[...]

PR #738, opened 2019-07-02
Fix for 737

Avd6977:
Fix for #737


Avd6977:
@gabrielliwerant Found a small issue with the custom filter stuff.

gabrielliwerant:
Ok, so if I'm understanding correctly, this is an issue with the way the example was constructed? It looks like the example should not have been using arrays in way that it was, which I missed when I reviewed it. And this PR is a fix for that?

Avd6977:
Yep, this would just be fixing the example to be correct.

PR #746, opened 2019-07-05
Add Select On Drag Functionality

17gcolestock:
User can click and drag to select multiple rows at once.


patorjk:
This is a really cool feature, however, I'm not so sure it should be wired into the table. Dragging to select rows could be done by using the setRowProps feature and setting up the drag methods there. Rows could be selected by updating the rowsSelected option. So this feature is possible without internal updates. 

I think for now I'm going to close this PR, though I'd be open to one that was an example on how to do draggable selections using the existing table options (setRowsProps and rowsSelected). Thank you for putting this together and sorry it has taken so long for someone to review this.

Issue #762, opened 2019-07-10
Paging resets when serverside search and change page


Avd6977:
## Expected Behavior
Page passed in as prop should be taken

## Current Behavior
Page prop is ignored when search is not null. This disables paging when search exists and done serverside with table reload

## Steps to Reproduce (for bugs)

1. Set serverside true and maintain state
2. Search for a value
3. Select Next Page
4. Page stays at 0

## Your Environment

| Tech         | Version |
|--------------|---------|
| MUI-datatables  |     2.6.1    |

gabrielliwerant:
Thanks for finding this, there is a bug here, but it's a bit more subtle than it appears. We definitely need to override the page value the first time a search is entered. However, in subsequent page updates, we should allow the page to change. I put some notes in the PR. It's possible there won't be a fix without a state refactor overhaul for the whole table. :(

gabrielliwerant:
Should be resolved in `2.8.0`.

PR #763, opened 2019-07-10
Page reset762

Avd6977:
Fix for #762


gabrielliwerant:
Seems to me like this PR would cause a regression with this: https://github.com/gregnb/mui-datatables/pull/647.

If the page is set, and we don't reset it on searches (and filters and probably some other things), then the results may not show and we may get errors. The page has to be reset on searching.

gabrielliwerant:
What we need to know here is two things:

1. Whenever the `searchText` goes from empty to something, we need to reset page to 0
1. If the previous `searchText` is not empty, then we need to make sure the page number does not exceed the list of results (something like RoundUp(`rowsPerPage` / results) = max page number.

Avd6977:
Why don't we just not reset it and have that be the responsibility of the consumer? If they're passing in page and search text, then they know that the page should be reset and can simply add page = 0 when search text is changed (that's what we're doing and how we caught this). So if the consumer is managing the page state, then there shouldn't be a need to reset it.

gabrielliwerant:
I understand your point, but it would be unclear to me that I would need to reset the page after a search. I would expect searching to handle this for me. In particular, there are definitely issues with out of bounds pages after searching that should at least be checked against occurring, and then handled appropriately.

I'd also rather not go back on fixes that existed previously if at all possible (there are exceptions, but I don't think this should be one). People are relying on the fixes that were put in place previously, so I don't want to break it again. I think with additional logic, we should be able to achieve both aims here. If you like, I can help out here when I get a chance.

Avd6977:
Yeah, I understand not wanting to revert back a bug. My thought was that the whole project was moving away from managing the internal state and moving toward putting more onus on the consumer of the table. Some documentation to reset page when searching or filtering may just be needed? Does the paging update when the filter is changed, as well?

Could be documented as a breaking change, as well. But if you'd rather put it in the code to fix this, that's fine. What would be your ideas to do that and not break the old bug? I'm not sure exactly what you're wanting to fix it with so some help would be great.

src/MUIDataTable.js
```diff
@@ -202,7 +201,7 @@ class MUIDataTable extends React.Component {
     this.setHeadResizeable(this.headCellRefs, this.tableRef);
 
     // When we have a search, we must reset page to view it
-    if (this.props.options.searchText) this.setState({ page: 0 });
+    if (this.props.options.searchText && !this.props.options.page) this.setState({ page: 0 });
```
gabrielliwerant: Ok, how about `!this.props.options.serverside` instead of `!this.props.options.page`? Same for below. That way, we keep the old bugfix for people that are using custom pages, but we prevent `serverside` behavior from changing.

Avd6977: That should work for us and keep the serverside behavior the same, but would it be confusing for anyone using it? I made the change so if we're good with it you can just accept it.

Issue #608, opened 2019-05-10
is it possible to customize search box ?


rushalipatel:
![Screen Shot 2019-05-10 at 12 31 58 PM](https://user-images.githubusercontent.com/48195555/57508645-ad4a0300-731f-11e9-8436-6938f6099259.png)

I have to customize search-box like above screenshot..is It possible in MUI Datatable ??

gabrielliwerant:
At the moment, it is not. This feature was requested a while ago here https://github.com/gregnb/mui-datatables/issues/91. If there is a lot of interest in this feature, I can take it on next, as it has languished quite a bit since it was initially opened.

vnellore:
Hi @gabrielliwerant - I also would like some sort of customization or control over the search box. Currently it overlaps the title of the table when you type a query. Also the length of the search field seems very long. Thanks for your efforts on this project.

elilondon:
@gabrielliwerant I am trying to override the color of the line (default black/ when focused blue) but cant seem to figure out how to call to this. Is this also not possible?

gabrielliwerant:
@elilondon The only way to customize the search box at the moment is to use style overrides. You'll find examples for how to do this is the customize-styling example of this project. Also, the material-ui documentation will give you pointers about overriding styles.

gabrielliwerant:
The ability to add a custom search bar is now available in `2.8.0`. See documentation or release notes for more information.

Issue #494, opened 2019-03-13
Only Search when a mininum number of characters where entered


ivaqmo123:
When users do searchs they usually expect to do a search using more than 3 or 4 characters. Currently as soon as a user type in a character the mui-datatable start searching in all records, this could take 1 to 5 seconds to allow the user to type in the next character, this wait time depends on the quantity of rows in the table.

## Expected Behavior
Once the user enter a minimum # of characters (e.g. 4), the search would start search across all records in the mui-datatable OR another option could be that the user could trigger the search using the "ENTER" key (I would prefer this last option)

## Current Behavior
As soon as the user type in a character, the MUI-datatable starts searching over all records which could take seconds depending on the # of rows. So for ~400 records, the user needs to wait ~2-4 seconds to enter the next character. Even when using customSearch function, it always goes across all records in the table for the 1st character and all the process again for the subsequent characters.

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.9.2     |
| MUI-datatables  |    2.0.0-beta.58     |
| React        |   16.8.4      |
| browser      |    Firefox-chrome     |
| etc          |         |

aliaanis:
You can use lodash- debounce for this.
` handleSearchChange = debounce((text) => {
 this.doAPICall(text);
    }, 400)`

patorjk:
I second a debounce solution for this problem. I have a table with 2k+ rows, and there's a noticeable typing lag when I type into the search field. From what I can tell, the "handleSearch" function in https://github.com/gregnb/mui-datatables/blob/master/src/components/TableToolbar.js would simply need to be wrapped in a debounce method (similar to what @aliaanis posted above).

Issue #91, opened 2018-07-09
Customize search element


NickToye:
Expose search component so that it can be customized, such as adding a Placeholder.

NickToye:
Good news, although you've spelt Customise wrong. ;)

gregnb:
Really? http://www.dictionary.com/browse/customize

NickToye:
Americanisms.

chiran-subedi:
Hello Gregnb,

I hope you can help me with this. I have been using mui-datatable and its
too good. Firstly i'd like to thankyou for this awesome project.

I want some customization on the tooltips.

Right now, whenever i hover through each of the column head at the table,
the tooltip popover shows "sort" for all. Is there any way so that we can
provide different titles for each of the columns ?
Expected Behavior

const options = {
textLabels: {
body: {
toolTip: "sort"
}
}

There is an element for this inside textLabels which is toolTip, but this
shows for all of the column's head. Can you please help me with having this
functionality(tooltip popover) work for each column heads with different
strings ?

I have a screenshot for more clarification.

Thankyou in advance. :)
Chiran

gregnb:
@chiran-subedi Thanks for the feedback, but please file an issue. Posting your own issues in other people's issues can be inconsiderate. So if you can please remove this comment and the other and file an issue. Thank you

gregnb:
So #96 has a similar question about more customization support, but for the column header. I have given it some time and I'm trying to come up with a way to solve the current issues and get ahead of future issues.  I think it's going to require some breaking changes and what I'm proposing is the following changes to options:

Change `options` object on custom columns from `customRender` to `customBodyCell` which will then setup nicely `customHeadCell`:

```
const columns = [
      {
        name: "Active",
        options: {
          filter: true,
          customHeadCell: {
             render: (value, tableMeta, updateValue) => {}
          },
          customBodyCell: {
            render: (value, tableMeta, updateValue) => {
              return (
              <FormControlLabel
                label={value ? "Yes" : "No"}
                value={value ? "Yes" : "No"}
                control={
                  <Switch color="primary" checked={value} value={value ? "Yes" : "No"} />
                }
                onChange={event => {
                  updateValue(event.target.value === "Yes" ? false : true);
                }}
              />
            );

          }
        }
      }
    ];

```

[...]

PR #771, opened 2019-07-16
Allow to custom table search render

Vallher:
Our designer like to make crazy things and we need to adjust searchRender a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ The component accepts the following props:
 |**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender(data, dataIndex, rowIndex) => React Component`
 |**`customSort`**|function||Override default sorting with custom function. `function(data: array, colIndex: number, order: string) => array`
 |**`customSearch `**|function||Override default search with custom function. `customSearch(searchQuery: string, currentRow: array, columns: array) => boolean`
+|**`customTableSearch `**|function||Render a custom table search. `customTableSearch(searchText: string, handleSearch, hideSearch, options) => React Component`
 |**`elevation`**|number|4|Shadow depth applied to Paper component
 |**`caseSensitive `**|boolean|false|Enable/disable case sensitivity for search
 |**`responsive`**|string|'stacked'|Enable/disable responsive table views. Options: 'stacked', 'scroll'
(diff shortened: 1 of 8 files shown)


jagribble:
Would it not be better to just allow the passing of props to the search Component?

gabrielliwerant:
@jagribble The precedent in this library is to allow full customization of components where possible as this maximizes customization potential. I don't think that props would allow for the same things that a custom render would.

Additionally, this PR would satisfy some longstanding open issues like https://github.com/gregnb/mui-datatables/issues/608 and https://github.com/gregnb/mui-datatables/issues/91.

patorjk:
This PR would also solve https://github.com/gregnb/mui-datatables/issues/494.  Taking a closer look, my comment in that thread is incorrect. Since the TableSearch component is controlled, "handleSearch" can't be wrapped and still have the component work (since any debouncing would cause the text field not to update with one's typing). The only way to do what 494 wants is to use a different search component that isn't controlled.

README.md
```diff
@@ -152,6 +152,7 @@ The component accepts the following props:
 |**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender(data, dataIndex, rowIndex) => React Component`
 |**`customSort`**|function||Override default sorting with custom function. `function(data: array, colIndex: number, order: string) => array`
 |**`customSearch `**|function||Override default search with custom function. `customSearch(searchQuery: string, currentRow: array, columns: array) => boolean`
+|**`customTableSearch `**|function||Render a custom table search. `customTableSearch(searchText: string, handleSearch, hideSearch, options) => React Component`
```
gabrielliwerant: To keep the API somewhat consistent, lets call this option `customSearchRender`.

gabrielliwerant: To keep the API somewhat consistent, lets call this option customSearchRender.

examples/customize-tablesearch/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+    const options = {
+      filter: true,
+      selectableRows: true,
+      filterType: 'dropdown',
+      responsive: 'stacked',
+      rowsPerPage: 10,
+      customTableSearch: (searchText, handleSearch, hideSearch, options) => {
+        return (
+          <CustomTableSearch
```
gabrielliwerant: I think you forgot to provide a `onHide` function here. The searchbar can't be closed in your example.

test/MUIDataTableToolbar.test.js
```diff
@@ -124,6 +124,22 @@ describe('<TableToolbar />', function() {
     assert.strictEqual(actualResult.length, 0);
   });
 
+  it('should render a toolbar with custom TableSearch when option.customTableSearch is method ', () => {
```
gabrielliwerant: Let's change the wording here to 'should render a toolbar with custom TableSearch when option.customTableSearch is provided' (`provided` instead of `method`). The test itself documents that what we're looking for is a React component. `method` is a little confusing to me.

Vallher:
@gabrielliwerant updated

PR #791, opened 2019-07-24
Fix issue with width on custom filter component 

### LINKED ISSUE #790: Width on custom filter component
## Expected Behavior
FormControl should allow components to fill width
image

## Current Behavior
FormControl limits the width of components
image

We are using React-Select within our project, and attempting to make a custom filter using it results in this.

pkmnct:
Fix #790

### DIFF
--- a/src/components/TableFilter.js
+++ b/src/components/TableFilter.js
@@ -192,7 +192,7 @@ class TableFilter extends React.Component {
     return (
       <GridListTile key={index} cols={1}>
         <div className={classes.selectRoot}>
-          <FormControl className={classes.selectFormControl} key={index} fullWidth>
+          <FormControl className={classes.selectFormControl} key={index}>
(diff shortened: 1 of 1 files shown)


gabrielliwerant:
Hm, not sure I see how this relates to custom filters. Custom filters are rendered using the `renderCustomField` method, not the `renderSelect` method. What are your `filterOptions`?

pkmnct:
@gabrielliwerant You are correct. It turns out this was an issue on my implementation. I've closed the issue and PR!

pkmnct:
I did some more research on this, and I figured out what was causing the weirdness. The FormControl components for the Select and TextField filter options had a style that did the equivalent of the fullWidth prop. I've removed this style and used fullWidth on all of the FormControls that had used it, in addition to the one for the custom filter type.

Issue #735, opened 2019-07-02
Expand Rows


leducsang97:
I would like to add a feature of expanding more than just one child row. Currently I used a trick to create a table inside a Tablerow but it will shift the parent row content to the right. I would like to ask if there is any official way of supporting this feature. Thank you very much.

| Tech         | Version |
|--------------|---------|
| Material-UI  |     ^3.9.2    |
| MUI-datatables  |   ^2.0.0-beta.59      |
| React        |   ^16.8.1   |
| browser      |    chrome     |

tsuz:
Try this @leducsang97 


```

import TableCell from '@material-ui/core/TableCell';
import TableRow from '@material-ui/core/TableRow';

...
options={{
  renderExpandableRow: () => (
            <TableRow>
              <TableCell>A</TableCell>
              <TableCell>b</TableCell>
              <TableCell>c</TableCell>
            </TableRow>)
}}
```

tsuz:
Something like this should do
```
<TableCell colSpan={3}>A</TableCell>
```

leducsang97:
> Try this @leducsang97
> 
> ```
> 
> import TableCell from '@material-ui/core/TableCell';
> import TableRow from '@material-ui/core/TableRow';
> 
> ...
> options={{
>   renderExpandableRow: () => (
>             <TableRow>
>               <TableCell>A</TableCell>
>               <TableCell>b</TableCell>
>               <TableCell>c</TableCell>
>             </TableRow>)
> }}
> ```

This is exactly what I did. But the issue is it only works with 1 row. I dont know how to do it with more than 1 <TableRow>

tsuz:
Have you tried
```
<>
  <TableRow>
    <TableCell>A</TableCell>
    <TableCell>b</TableCell>
    <TableCell>c</TableCell>
  </TableRow>
  <TableRow>
    <TableCell>A</TableCell>
    <TableCell>b</TableCell>
    <TableCell>c</TableCell>
  </TableRow>
</>
```
or build your own table within specific cell like this:
```
  <TableRow>
    <TableCell />
    <TableCell colSpan={2}>
      <YourDataTable />
    </TableCell>
  </TableRow>
```

leducsang97:
> Have you tried
> 
> ```
> <>
>   <TableRow>
>     <TableCell>A</TableCell>
>     <TableCell>b</TableCell>
>     <TableCell>c</TableCell>
>   </TableRow>
>   <TableRow>
[...]

Issue #433, opened 2019-02-20
Expand all rows by default


QuynhNguyen:
Is there a way to expand all rows by default?

gabrielliwerant:
I don't believe so, but if you'd like to work on this as an option, I will review the PR.

gabrielliwerant:
As of `2.7.0`, you can now set the expanded state of all rows in the options. See the release notes for more information.

kitaroar:
I use this code in [options] to expand all rows by default
data is the array to display in table

```
rowsExpanded: data.map((el, i) => {return i}),
```

PR #800, opened 2019-07-30
Add table option: show expanded row(s) by default

lnaia:
Add option to render rows expanded by default.

There's some scenarios, specifically for my employer current needs, to sometimes have a few rows, expanded by default, when the data is rendered, while still have the option to toggle it.

This PR is first attempt to try and add that behaviour. 
There's an optional `isRowDefaultExpanded` function, that will determine weather the row should expanded or not. 

Related with: #433 , #735

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ The component accepts the following props:
 |**`expandableRows`**|boolean|false|Enable/disable expandable rows
 |**`expandableRowsOnClick`**|boolean|false|Enable/disable expand trigger when row is clicked. When False, only expand icon will trigger this action.
 |**`renderExpandableRow`**|function||Render expandable row. `function(rowData, rowMeta) => React Component`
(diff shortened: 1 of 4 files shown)


patorjk:
I'm in need of a similar feature, though in addition to showing some rows open by default, it'd also be useful for me to open/close all.

I checked out your PR and noticed something interesting: If I set "sortDirection: 'desc'" on the Salary column, the first row is not expanded. Instead the last row is expanded. I'm guessing the isRowDefaultExpanded calculation happens before the sorting is done.

gabrielliwerant:
Check out `2.7.0` and see if this satisfies your use case.

lnaia:
@gabrielliwerant I adapted to use your solution just a few minutes after you released it :-)
Thank you very much, it is the kind of API we were looking for to dealt with our issues.
:100:

PR #805, opened 2019-08-01
Fix issue with display filter on IE11

ntt2k:
There is a minor display issue with the filter component on IE11.
Please see the screenshot -->

**_Before:_**

before

**_After:_**

after


ntt2k:
the issue has already been fixed in v2.8.0

gabrielliwerant:
Oh nice, it must have been related to the simplification of styles for the filter fields. Always nice when one fix helps another!

Credit goes to @pkmnct with https://github.com/gregnb/mui-datatables/pull/791!

PR #808, opened 2019-08-02
Add searchProps

jagribble:
One of the things that I can see is quite a problem with a lot of people with the search customization not being the best. So I have added the searchProps to option which spreads against the Seach text field. 

This allows people to define their own props (onBlur, onChange etc.)

Would there be a chance of reviewing this soon as we kind of need this functionality soon @gregnb

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4498,7 +4498,8 @@
           "version": "2.1.1",
           "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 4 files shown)


gabrielliwerant:
There is a PR already open for an entire custom search render, which I think would handle your use cases as well, so I'd like to see how that goes first: https://github.com/gregnb/mui-datatables/pull/771.

Marking as duplicate for now, but we can revisit after the other PR if there is still a need for something like this.

gabrielliwerant:
@jagribble The new option for custom search bars (`customSearchRender`) is available in `2.8.0`. Does that feature satisfy your use case?

gabrielliwerant:
@jagribble If you'd like to bring this PR up to date, I would like to consider adding it to the library, as it may solve some problems for people a bit more elegantly than a custom render solution.

Additionally, I think it would be good not to require specific properties here, and allow all props from an object to be spread to the component, as this matches similar features elsewhere.

gabrielliwerant:
This has fallen out of date, and @patorjk offered to submit a new PR with this feature, so this is on hold for now.

patorjk:
Sure, I'll put something together, I'll loop back to it later this week or early next week.

patorjk:
Feature added into v3 branch.

Issue #783, opened 2019-07-20
remove toolbar with trash button when user select any row


suiteappdev:
How to remove toolbar with trash icon when user has selected checkbox row? i dont want show this toolbar.

can any help me please?

patorjk:
You can use the "customToolbarSelect" method to provide a custom toolbar without the trash icon. I don't think you can override the "X row(s) selected" title or prevent the selected toolbar from appearing though.

gabrielliwerant:
Yes, `customToolbarSelect` will help you with what options show in this toolbar. At this point there isn't a built-in option to remove the toolbar entirely. I'm interested in hearing the use case for not having any toolbar (e.g. how would users then take an action based on their selections?).

In any event, you can forcibly remove the select toolbar by overriding css and setting to display: none or something, but be aware that this will not restore the original table toolbar, if that is your intent.

If you want to restore the original toolbar on selections, you can see https://github.com/gregnb/mui-datatables/issues/712 and https://github.com/gregnb/mui-datatables/issues/451 for discussion around this, as it has come up a few times. I'm still not sure it's worth adding that option due to UX issues, but I suggest a path forward to allow easier select toolbar customization, it's just not a priority at the moment.

sankho:
I'm on a project rn working with a design team that doesn't want this feature and is not using any filters beyond basic checkbox selection so there's no worries UX wise -- it would be useful to have a direct approach to hiding things as the ultimate UX can be beyond the ideal definition of the codebase as is IMO.

gabrielliwerant:
@sankho Hello, can you explain a bit more about your use case and user flow?

mattspahr:
> You can use the "customToolbarSelect" method to provide a custom toolbar without the trash icon. I don't think you can override the "X row(s) selected" title or prevent the selected toolbar from appearing though.


Have you found a way to customize the "X row(s) selected"?  I was able to change the "row(s) selected" but cannot get rid of the X . It seems to get it from the length of selectedRows.

This is causing me a lot of trouble. Thanks!

raphael10-collab:
@gabrielliwerant @patorjk How can we keep the toolbar with the trash icon and customize it (adding other options), or, how can we use a customToolbarSelect adding the pre-defined trash icon with its pre-defined functionality?

Xaxis:
[...]

Issue #712, opened 2019-06-21
Allow option to prevent toolbar select from displaying


mattgbrich:
It would be really nice to allow us to pass a flag to prevent this toolbar from appearing. 

I need to enable row selection, but I don't want that toolbar to appear, and I don't want users to be able to delete rows.

It looks like it's currently just hardcoded to always appear if `selectedRows.data.length`

```javascript
{selectedRows.data.length ? (
          <TableToolbarSelect
```

I can eliminate the delete action/icon with `customToolbarSelect`, but then I still get an empty toolbar covering up the main toolbar (which I want users to still have access to after making a selection).

It seems like it would be a common use case to want row selection without having that toolbar pop up.

gabrielliwerant:
Marking as a duplicate of https://github.com/gregnb/mui-datatables/issues/451, as a similar issue was raised there, and there's already a bit of conversation around it. The short story is that there is currently no plan to disable the select toolbar, and this alone won't do what you're hoping anyway, as the regular toolbar isn't "select-aware."

Instead, I think it should be possible to add any options you want through the custom toolbar, thereby recreating the options you see fit to have available, and choosing the behavior you want.

Issue #451, opened 2019-02-28
toobarSelect replaces toolbar when rows are selected


b1rt:
I'm not sure if it's bug or not, but it is negative user experience.

If you select any rows after applying search of filters, toolbarSelect (which shows how many rows you are selected) replaces and hovers standard toolbar, thus make it impossible to discard or change search, or change filters (but you can discard filters, as they appear under toolbar).

So, if user select any rows after searching, he must discard his selection first to discard or change search string, making it useless to search and then select rows.

## Expected Behavior

After selecting rows, user can change search string of filters.

## Current Behavior

After selecting rows, toolbar disappears (and replaces with toolbar select)

## Steps to Reproduce (for bugs)

1. Search smth in your table.
2. Select any of found rows.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  3.3.1       |
| MUI-datatables  |  2.0.0-beta-52       |
| React        |  16.6.0       |
| browser      |   chrome      |

jakesyl:
I'm having the same issue

byb121:
I'm having the same issue, I hope I can place the whole bar elsewhere, or disable it completely.

gabrielliwerant:
I think the UX would be troublesome to have both toolbars available at the same time. For example,

- With select bar and toolbar open, should the download/print options work only on selected rows or not?
- With select bar and toolbar open, changing filters can remove already selected rows. Should they remain selected and then return selected if the filters change again in a way that reveals them?
- With select bar and toolbar open, changing the search can also remove already selected rows.

Due to the UX difficulties of having both of these options available at the same time, there is often a toggle that enables/disables select checkboxes in interfaces.

What I think would help here is the capacity to add the main toolbar options to the select toolbar, thereby making the fact that they are operating in a select context clear. That might be a bit of a work, but I think it's a worthy goal.

mattgbrich:
@gabrielliwerant 

I would suggest that the Title/textCopy area should also be considered. It's often much more desirable to keep the table Title visible rather than being forced to replace it with `<num> row(s) selected`. 

This is especially relevant to single-select tables. Constantly displaying `1 row(s) selected` is meaningless information in that case, and creates a poor UX by removing relevant information from the users view, and replacing it with information they don't care about. 


[...]

PR #815, opened 2019-08-06
Make selectToolbar an option: options.displaySelectToolbar = false

lamyfarai:


gabrielliwerant:
A request for something like this has come up numerous times (https://github.com/gregnb/mui-datatables/issues/783, https://github.com/gregnb/mui-datatables/issues/712, https://github.com/gregnb/mui-datatables/issues/451), but unfortunately, I haven't yet received an answer as to what would happen in this case if the user selects something and wants to take an action based on the selection?

I'd like to hear a good UX answer as to what a user should do without a select toolbar before I can consider adding a feature like this.

lamyfarai:
@gabrielliwerant  What if you have, maybe 20 or 30 pages, and need to select items and not have an option to search  for "what to select". Such was my case anyways.

gabrielliwerant:
@lamyfarai Keeping a search bar showing during the selection of rows I think is a valid use case, and I like that idea. However, I'd prefer to see an implementation that did just that: maintain the search bar during selection, rather than remove the select toolbar altogether.

Also, marking this is duplicate for now, because there has been a PR up already for this exact thing.

patorjk:
I think this is a good idea. I needed to keep the search visible in one of my tables and wound up just implementing my own checkbox column to get around the select toolbar.

However, I'm not a fan of the implementation in https://github.com/gregnb/mui-datatables/pull/602 since you get 2 headers showing when you select an item (example here: https://codesandbox.io/s/mui-datatables-ehndq).

This PR also has the double toolbar issue (ex: https://codesandbox.io/s/github/lamyfarai/mui-datatables).

I think the optimal solution would be to have an option (say disableToolbarSelect) that if true, would keep the normal toolbar visible and not show the select toolbar if something was selected. Then as far as row selection goes, it'd be up to the developer to use the onRowsSelect callback to know what the selected items were.

I think all it would take is updating the PropTypes in MUIDataTable.js (line 134-ish) and modifying this line: https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L1184

    {selectedRows.data.length ? (

to be:

    {(selectedRows.data.length && this.options.disableToolbarSelect !== true) ? (

manoj-raghuvanshi:
can someone merge this, I need this option.

gabrielliwerant:
Version `2.13.0` has `disableToolbarSelect` which allows you to remove the select toolbar, which reveals the regular toolbar.

PR #825, opened 2019-08-09
Enable functionality to limit maximum number of rows to select

### LINKED ISSUE #751: Limit checkbox based selection for more than 'n' rows
Would like to see this feature:
Currently i cant limit how many rows i can select, i would want that there should be a limiter 'n' where if number of rows selected is equal to 'n' then any further row selection is not allowed.

For example if i have table with 10 selectable checkbox based normal rows, i would want a limiter of 3 lets say. If i try selecting more than 3 rows, there should be some message in the header that "Cannot select more than 3 rows" or something.

| Tech         | Version |
| Material-UI   v4
| MUI-datatables  v6
| React      v16
| browser     chrome
| etc

endrjuskr:
Closes #751.

Interface change:
```
Before:

isRowSelectable: (dataIndex) => boolean;

Now:

isRowSelectable: (dataIndex, selectedRows) => boolean;
```


endrjuskr:
Question: How we would like to handle a situation (quite common) when a maximum number of selected rows is lower than a number of rows and a user clicks on the header (select all)?

Current behavior: Selects all and unclicking any of row disables this row to be selected again.

gabrielliwerant:
Hello @endrjuskr. Thank you for taking an interest in contributing to the project!

I've been thinking about this feature, and I think what I'd like to do instead, is extend the functionality of `isRowSelectable` to include the `selectedRows` object. This way, devs can use that information to insert any custom logic that they want to enable/disable selecting a particular row. As of this moment, `isRowSelectable` only tracks the dataIndex of the current row about to be selected. Although all the selected rows could be tracked separately by the dev, adding this argument would make it easier.

The advantage this has is that it allows customization based any logic the dev wants, and it extends an already existing API rather than create a new one. I'm a little wary of creating new APIs if we can use existing ones to do the same job, as I want to be careful about creating an API that is so big and complex that it's difficult to reason about and lacks discoverability for those new to the project.

How does that sound?

endrjuskr:
Sounds reasonable. To be honest, I was not fully convinced about the approach that I proposed, mostly due to selecting all rows from header.

README.md
```diff
@@ -142,7 +142,7 @@ The component accepts the following props:
 |**`pagination`**|boolean|true|Enable/disable pagination
 |**`selectableRows`**|string|'multiple'|Numbers of rows that can be selected. Options are "multiple", "single", "none".
 |**`selectableRowsOnClick`**|boolean|false|Enable/disable select toggle when row is clicked. When False, only checkbox will trigger this action.
-|**`isRowSelectable`**|function||Enable/disable selection on certain rows with custom function. Returns true if not provided.  `function(dataIndex) => bool`
+|**`isRowSelectable`**|function||Enable/disable selection on certain rows with custom function. Returns true if not provided.  `function(dataIndex, selectedRows) => bool`
```
gabrielliwerant: I think we should also explain the structure of the `selectedRows` object so people know what it contains.

test/MUIDataTableBody.test.js
```diff
@@ -493,4 +493,31 @@ describe('<TableBody />', function() {
 
     expect(html).to.contain('Test_Text');
   });
+
+  it('should selectedRows be passed in isRowSelectable', () => {;
```
gabrielliwerant: The wording of this threw me off a little bit. Can we change it to "should pass in selectedRows to isRowSelectable"?

gabrielliwerant:
@endrjuskr Almost there! We're still missing the `selectedRows` argument in what is now this line: https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L1034.

README.md
```diff
@@ -142,7 +142,7 @@ The component accepts the following props:
 |**`pagination`**|boolean|true|Enable/disable pagination
 |**`selectableRows`**|string|'multiple'|Numbers of rows that can be selected. Options are "multiple", "single", "none".
 |**`selectableRowsOnClick`**|boolean|false|Enable/disable select toggle when row is clicked. When False, only checkbox will trigger this action.
-|**`isRowSelectable`**|function||Enable/disable selection on certain rows with custom function. Returns true if not provided.  `function(dataIndex) => bool`
+|**`isRowSelectable`**|function||Enable/disable selection on certain rows with custom function. Returns true if not provided. `function(dataIndex, selectedRows) => bool`. `selectedRows` contains information about selected rows in following format: `{ data // array of data objects, lookup // map of selected ids }`
```
gabrielliwerant: Just to keep the descriptions parallel between other areas of documentation, I think we should use this format for the `selectedRows` argument: "object(lookup: {dataindex: boolean}, data: arrayOfObjects: {index, dataIndex})"

test/MUIDataTableBody.test.js
```diff
@@ -271,6 +271,42 @@ describe('<TableBody />', function() {
     assert.strictEqual(toggleExpandRow.callCount, 0);
   });
 
+  it('should not gather selected row data when clicking row with selectableRowsOnClick=true when it is disabled with isRowSelectable via selectedRows.', () => {
```
gabrielliwerant: Thank you for adding this! Additional tests are always appreciated.

To keep the types in line with what's used in the table, how about changing the `selectedRows` argument to contain `{ lookup: { dataindex: 2 }, data: [{ dataindex: 2, index: 2 }] }` and then updating `selectedRows.length` in the `isRowSelectable` function to `selectedRows.data.length`?

gabrielliwerant:
Hey @endrjuskr, thanks for the update, however, if you look at the comment, I was actually referring to the test on line 274. :)

endrjuskr:
I hope this time I fixed correct test :)

Also split one test to check exactly if `isRowSelectable` actually determines selectability of row.

test/MUIDataTableBody.test.js
```diff
@@ -493,4 +533,69 @@ describe('<TableBody />', function() {
 
     expect(html).to.contain('Test_Text');
   });
+
+  it('should isRowSelectable be used to determine if row is selectable', () => {
```
gabrielliwerant: Actually, I think we can eliminate this test, or recombine it. (or change it as I'll discuss below). It looks like it has a couple of issues once separated out. Mainly, it just seems to be testing whether or not the `isRowSelectable` function we create here in the test has the proper return values, which basically means that the test is testing itself. It doesn't test that the table works because it doesn't simulate an interaction with the table select rows.

So, if we just want to show that the function is called, the test below (on line 571) does that, so we could combine it back with that one.

If instead we want to show the positive case, where we demonstrate that a click on a selected row that is allowed by `isRowSelectable` works, then I think we should model it after the one on line 274 and simulate the clicks.

I leave it to you as for how you'd like to resolve this one.

Issue #663, opened 2019-06-04
Shift + click to select multiple rows


patorjk:
Enhancement suggestion: I noticed that the common keyboard short-cut Shift + Click doesn't seem to work with selectable rows (selectableRows="multiple"). It's a pretty useful keyboard shortcut (ex: in GMail, select a row, and then shift+click another row, all of the rows in-between will be selected).

patorjk:
Added in v2.10.0

PR #827, opened 2019-08-11
Added support for Shift + Click selection of rows

patorjk:
An implementation of the feature suggested in https://github.com/gregnb/mui-datatables/issues/663

The UX for shift+click is modeled off of how GMail does it. For example, if you select/unselect row A and then hold shift and select/unselect row B:

- If row B is being selected, then the rows between A and B (including A) will be selected.
- If row B is being unselected, then the rows between A and B (including A) will be unselected. 

If any filters are removed or columns resorted or selection headers toggled or data fetched (in the case of Server Side tables), between the selection of rows A and B, then there will be no special effect.


patorjk:
This PR is most likely not compatible with https://github.com/gregnb/mui-datatables/pull/825, however, I could adjust it so that in the MUIDataTable.js file, when it was in the process of adding the adjacent rows, it would stop if it hit the maxSelectedRows threshold.

gabrielliwerant:
This is a cool feature, thank you for working on this!

I just reviewed and suggested a number of changes to https://github.com/gregnb/mui-datatables/pull/825, so how about we wait for that to be merged and integrated with yours before review?

patorjk:
Sounds good, I’ll be unavailable next week but could make any adjustments tomorrow or the week after that.

gabrielliwerant:
@patorjk The `isRowSelectable` changes are in latest master now, if you want to pull those.

patorjk:
Sounds good, I’ll pull them and update this PR early next week!

Issue #789, opened 2019-07-24
Reusable TablePagination for customFooter


patorjk:
I'm looking to do a custom footer, though all I really want to do is add a "jump to page" option (maybe a dropdown or buttom that brings up a dialog, haven't decided yet). I was poking around to see if I could reuse the existing footer and just stick another component next to it, but I noticed two things:

* The footer options aren't passed to the customFooter callback. These options are needed for the TablePagination component (https://github.com/gregnb/mui-datatables/blob/master/src/components/TableFooter.js#L28).
* The TablePagination component starts with MuiTableFooter and MuiTableRow tags, so even if I had the options, I wouldn't be able to reuse this component (https://github.com/gregnb/mui-datatables/blob/master/src/components/TablePagination.js#L56).

I can recreate everything, but it would be nice to reuse what the table uses. Not sure if it would break anything, but spliting TablePagination into 2 components (a reusable BaseTablePagination component and the normal TablePagination component that has the MuiTableFooter/MuiTableRow tags wrapping BaseTablePagination), plus passing the pagination options to customFooter would seem to allow this in theory.

gabrielliwerant:
Should be resolved via your PR for `customFooter` options in `2.9.0`.

PR #828, opened 2019-08-11
Added options parameter to customFooter callback

patorjk:
Fix for https://github.com/gregnb/mui-datatables/issues/789

I've updated the customFooter callback to also take in the options parameter. With this parameter, one can more easily re-create the table pagination. I've also fleshed out the customize-footer example by showing how the Material UI TablePagination component can be used inside of the footer with other custom options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ The component accepts the following props:
 |**`resizableColumns`**|boolean|false|Enable/disable resizable columns
 |**`customToolbar`**|function||Render a custom toolbar
 |**`customToolbarSelect`**|function||Render a custom selected rows toolbar. `function(selectedRows, displayData, setSelectedRows) => void`
-|**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage, textLabels) => string`&#124;` React Component`
+|**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage, `[`textLabels: object`](https://github.com/gregnb/mui-datatables/blob/master/src/textLabels.js)`) => string`&#124;` React Component` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-footer/index.js)
 |**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender(data, dataIndex, rowIndex) => React Component`
 |**`customSort`**|function||Override default sorting with custom function. `function(data: array, colIndex: number, order: string) => array`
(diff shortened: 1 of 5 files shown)


gabrielliwerant:
Hello @patorjk. So in looking through this, it looks like the only options being used in `TablePagination` are `textLabels` and `rowsPerPageOptions`. `rowsPerPageOptions` is easy enough to supply directly without `options`. `textLabels` can also be supplied by whoever is creating the custom footer, although it might be nice to have access to the internal labels used by the table, so maybe there's a case to just supply that.

I think I would be confused about an generic `options` argument, wondering what is there that I would be reusing in my custom footer vs. whatever I'd be responsible for supplying myself. So I only really see a case for providing `textLabels`, and all of it is already quite doable with the custom component. Am I missing a particular use case or something?

patorjk:
Yeah it’s mainly so the footer has access to the same labels as the table. I passed the options object since that’s what the table passes to it’s own footer.

The main motivation for the PR was to show (what I think) is a common use case of adding something to the footer. When I first saw the current example I thought I was going to have to add all the buttons/labels/drop downs myself. It wasn’t obvious to me that there was a pagination component that I could plug all that stuff into.

This is just something I thought might be useful, it’s non-essential so no worries if you’re not interested in adding it.

gabrielliwerant:
I think most of this PR is really useful. I'd love to keep the example, because as you say, it demonstrates how to recreate the internal table footer. Additionally, I think we should change `options` to `textLabels` and just use it explicitly for that purpose for now. We can revisit an `options` object if the footer becomes more complicated and there's a greater need.

How does that sound?

patorjk:
That sounds good, I'll make the updates later this afternoon.

README.md
```diff
@@ -149,7 +149,7 @@ The component accepts the following props:
 |**`resizableColumns`**|boolean|false|Enable/disable resizable columns
 |**`customToolbar`**|function||Render a custom toolbar
 |**`customToolbarSelect`**|function||Render a custom selected rows toolbar. `function(selectedRows, displayData, setSelectedRows) => void`
-|**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage) => string`&#124;` React Component`
+|**`customFooter`**|function||Render a custom table footer. `function(count, page, rowsPerPage, changeRowsPerPage, changePage, textLabels) => string`&#124;` React Component`
```
gabrielliwerant: How about we point out that text labels is an object or perhaps link to the `textLabels.js` file so people know how to implement it themselves if they want to change the wording?

PR #833, opened 2019-08-14
Code cleanup

pawel-schmidt:


src/MUIDataTable.js
```diff
@@ -1025,9 +1008,7 @@ class MUIDataTable extends React.Component {
           const { displayData } = prevState;
           const selectedRowsLen = prevState.selectedRows.data.length;
           const isDeselect =
-            selectedRowsLen === displayData.length || (selectedRowsLen < displayData.length && selectedRowsLen > 0)
-              ? true
-              : false;
+            selectedRowsLen === displayData.length || (selectedRowsLen < displayData.length && selectedRowsLen > 0);
```
gabrielliwerant: Now that I'm looking at this line more closely, another confusing bit here is that the order of operations in evaluating the comparison is really important, so could we further change this to add parens?

```
(selectedRowsLen === displayData.length) || (selectedRowsLen < displayData.length && selectedRowsLen > 0)
```

gabrielliwerant:
Continued here https://github.com/gregnb/mui-datatables/pull/883.

PR #835, opened 2019-08-14
Allow async onRowsDelete

pkmnct:


gabrielliwerant:
Hey @pkmnct, trying to get a handle on the exact use case here, as this was a spinoff from an issue that was asking specifically about a confirmation dialog. In my experimentation, there is no issue with using a confirmation dialog because it takes focus and prevents other code execution until the dialog is dealt with, allowing true/false to be returned from the callback.

Were you trying to do something other than use a confirmation dialog?

Additionally, I think this needs some babel plugin and configuration love, as I couldn't run any examples using `onRowsDelete` otherwise. I've done the work to set it up, so I can chip in on this if you like.

pkmnct:
@gabrielliwerant We were wanting to use a Material UI Dialog component, rather than a window.confirm dialog.

Here's an Code Sandbox: https://codesandbox.io/s/material-ui-datatables-async-onrowsdelete-example-wktpj

gabrielliwerant:
@pkmnct Gotcha, that makes sense.

Any thoughts on how you want to handle the babel configs to handle `async`/`await`?

pkmnct:
@gabrielliwerant I'm not having any issues using this PR branch locally testing, what error/warning/issue are you running into with the async/await and babel?

gabrielliwerant:
Webpack can't run the sections with `async` because it's missing some plugins like `babel-plugin-syntax-async-functions` and `babel-polyfill`. It's possible that you have extra packages installed locally or browser updates which are helping. My research indicates that those are needed, and this does seem to fall in line with the fact that there have been no other uses of `async` in the library until now.

In particular, I get a `regenerator runtime` error (which is connected to `async` if I understand correctly), when I try to load the `selectable-rows` example in Chrome (v76).

aberkb (approved):

PR #863, opened 2019-08-24
Fix #778 Enhancement/switch between examples

### LINKED ISSUE #778: how to switch between demo of examples without restarting npm
How can I switch between the demo of examples like expandable-rows, serverside-options, etc without having to change webpack.config.js and restart?

I have checked out mui-datatables code and want to see different examples live and possibly change them and see the changes.

I figured out that by changing webpack.config.js:

module.exports = {
  entry: {
    app: "./examples/customize-filter/index.js"
  }

and restarting the app I could accomplish this.

But is there a way to do it without restarting? and without changing the webpack?

One solution I'm hoping for is that there should be direct urls like :
http://localhost:5050/expandable-rows
http://localhost:5050/serverside-options

to see the respective examples.

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

alielkhateeb:
Fix #778 [Enhancement] Switch between examples.

- Added an entry point with a list of examples to choose from
- Hash Router to all the existing examples

You can use `examples.js` to easily add more examples in the future, just add the label and the component to render.


alielkhateeb:
And I also think we could improve the documentation by adding live demo links to each option we have an example for, it would be much easier for the users to have a look on what this option actually does visually.

For example, in the description of `customFilterListRender`, we could have [Demo](https://rp4tq.sse.codesandbox.io/#customize-filter) along with the included [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/column-filters/index.js)

Just a thought.

gabrielliwerant:
@alielkhateeb Live demo links would be great, but probably make more sense in a full documentation website, like what material ui has. As the library matures we could definitely use something like that, but I think we should hold off for now. I know the repo owner attempted this a while ago, so he may have thoughts on it first.

gabrielliwerant:
@alielkhateeb Looks like some of the other PRs were touching the `webpack.config.js` to switch examples, so that created a conflict for you. After your changes are merged, we'll never need to deal with that again!

alielkhateeb:
@gabrielliwerant Merged.
Yup, no one will have to push this file again 🍾

PR #864, opened 2019-08-25
Added a notice about column sorting when serverSide option is enabled

tomgazit:


patorjk:
I'm not sure this notice is needed anymore after the 2.15.0 update, since the table won't erase the sortDirection when rendered with new data. I'm closing this for now, but if you think a notice should be added I'm open to it.

PR #874, opened 2019-08-27
Added disableToolbarSelect option

patorjk:
An alternative implementation of the idea in PRs https://github.com/gregnb/mui-datatables/pull/815 and https://github.com/gregnb/mui-datatables/pull/602

This option allows a developer to not display the select toolbar when rows are selected. This can be useful for searching and filtering. 

Only thorny part was how to handle the Select All toggle. I setup it up to not allow you to unselect rows that are filtered out (since you don't select them when you Select All).

Quick view of PR: https://codesandbox.io/s/github/patorjk/mui-datatables/tree/disable-select-toolbar

I added these options when trying it out:

      disableToolbarSelect: true,
      rowsSelected: [0],
      isRowSelectable: function(dataIndex, selectedRows) {
        //return true;
        return dataIndex > 0;
      }


patorjk:
thanks! Couple random suggestions for the next major release:

onRowsSelect -> onRowSelectionChange
onRowsExpand -> onRowExpansionChange

Just makes them a little more clear that they handle expand/collapse and select/deselect cases.

src/MUIDataTable.js
```diff
@@ -1069,11 +1071,24 @@ class MUIDataTable extends React.Component {
...
-          }, []);
+           }, []);
 
           let newRows = [...prevState.selectedRows, ...selectedRows];
           let selectedMap = buildMap(newRows);
 
+          // if the select toolbar is disabled, the rules are a little different
+          if (this.options.disableToolbarSelect === true) {
+            if (selectedRowsLen > displayData.length) {
+               isDeselect = true;
```
gabrielliwerant: Looks like the merge eliminated the declaration change from `const` to `let` for `isDeselect`.

patorjk: Updated. Also, I've given you commit privileges to the branches I've PR-ed, just so you're not waiting on me if you see anything small like this.

patorjk:
Looks like they were pretty simple, should be resolved now.

Issue #869, opened 2019-08-27
Closing the search bar doesn't trigger onSearchChange.


Vanilor:
When something is typed in the search bar and then the bar is closed it does not trigger a onSearchChange. 

Example of the issue: https://codesandbox.io/embed/muidatatables-onsearchchange-bug-qfxsq

## Expected Behavior

Clicking the close icon on the search bar and by that clearing the search and closing the bar should trigger onSearchChange.

## Current Behavior

onSearchChange is not called but onTableChange is triggered with action "search".

## Steps to Reproduce (for bugs)

1. Open https://codesandbox.io/embed/muidatatables-onsearchchange-bug-qfxsq
2. Open the search bar and type anything.
3. Click on the X in the search bar to close it.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  | 3.2.2/4.1.3  |
| MUI-datatables  |  2.9.0/2.6.0  |
| React        |  16.8.6  |
| browser      |  Chrome  |

PR #876, opened 2019-08-28
Fix for #869

patorjk:
https://github.com/gregnb/mui-datatables/issues/869

* Added documentation for onSearchClose and made a corresponding onTableChange event for it.
* Moved triggering of onSearchChange to MUIDatatable component so that it mirrors the "search" event. This also makes it so customSearch components don't have to manage this event.
* Made sure customSearch components can't double invoke this method if they're already setup to call it.

Quick view: https://codesandbox.io/s/github/patorjk/mui-datatables/tree/onSearchChange

Used this setup to follow event call order:

      onSearchOpen: () => {
        console.log('onSearchOpen');
      },
      onSearchClose: () => {
        console.log('onSearchClose');
      },
      onSearchChange: (txt) => {
        console.log('onSearchChange:'+txt);
      },
      onTableChange: (action, tableState) => {
        console.log("onTableChange: " + action + ', searchText:'+tableState.searchText);
      },

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ The component accepts the following props:
 |**`onChangeRowsPerPage`**|function||Callback function that triggers when the number of rows per page has changed. `function(numberOfRows: number) => void`
 |**`onSearchChange`**|function||Callback function that triggers when the search text value has changed. `function(searchText: string) => void`
 |**`onSearchOpen`**|function||Callback function that triggers when the searchbox opens. `function() => void`
+|**`onSearchClose`**|function||Callback function that triggers when the searchbox closes. `function() => void`
 |**`onFilterChange`**|function||Callback function that triggers when filters have changed. `function(changedColumn: string, filterList: array) => void`
 |**`onColumnSortChange`**|function||Callback function that triggers when a column has been sorted. `function(changedColumn: string, direction: string) => void`
 |**`onColumnViewChange`**|function||Callback function that triggers when a column view has been changed. `function(changedColumn: string, action: string) => void`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -179,6 +179,7 @@ The component accepts the following props:
 |**`onChangeRowsPerPage`**|function||Callback function that triggers when the number of rows per page has changed. `function(numberOfRows: number) => void`
 |**`onSearchChange`**|function||Callback function that triggers when the search text value has changed. `function(searchText: string) => void`
 |**`onSearchOpen`**|function||Callback function that triggers when the searchbox opens. `function() => void`
+|**`onSearchClose`**|function||Callback function that triggers when the searchbox closes. `function() => void`
```
gabrielliwerant: 👍

src/components/TableToolbar.js
```diff
@@ -196,12 +198,18 @@ class TableToolbar extends React.Component {
     const { search, downloadCsv, print, viewColumns, filterTable } = options.textLabels.toolbar;
     const { showSearch, searchText } = this.state;
 
+    // prevent a customSearchRender component from calling the onSearchChange callback
+    let optionsCopy = Object.assign({}, options);
```
gabrielliwerant: I actually don't think we need this bit. Custom search render doesn't rely on `onSearchChange` at all, so we don't need to override here. It may also be a good thing to allow people to use this function in their own custom search renders if they want.

patorjk: Sounds good, just took it out.

Issue #880, opened 2019-08-29
Empty strings don't work in Dropdown filters


patorjk:
In my data there can be fields with empty strings. I noticed that the dropdown filtering doesn't work for them - it treats them the same as the "All" keyword, which leads to odd results when selecting them. You can see it in action here:

https://codesandbox.io/s/muidatatables-custom-toolbar-sisef

The problem stems from the handleDropdownChange method in TableFilter.js:

```
  handleDropdownChange = (event, index, column) => {
    const labelFilterAll = this.props.options.textLabels.filter.all;
    const value = event.target.value === labelFilterAll ? '' : event.target.value;
    this.props.onFilterUpdate(index, value, column, 'dropdown');
  };
```

It treats an empty string as well as the "All" keyword as the same, and then passes an empty string to the onFilterUpdate method (filterUpdate in MUIDataTable.js). filterUpdate hits the "default" in the switch statement for dropdowns and removes the filters on an empty string:

```
  filterUpdate = (index, value, column, type) => {
        ...
        switch (type) {
          case 'checkbox':
            filterPos >= 0 ? filterList[index].splice(filterPos, 1) : filterList[index].push(value);
            break;
          case 'multiselect':
            filterList[index] = value === '' ? [] : value;
            break;
          case 'custom':
            filterList[index] = value;
            break;
          default:
            filterList[index] = filterPos >= 0 || value === '' ? [] : [value];
        }
```

## Expected Behavior
Empty strings should be valid filter options.

## Current Behavior
Empty strings, as well as the All keyword, are treated the same.

## Steps to Reproduce (for bugs)

1. https://codesandbox.io/s/muidatatables-custom-toolbar-sisef
2. Select the Filter icon
3. Select the empty string option

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    2.9.0     |
| MUI-datatables  |    3     |
| React        |    16.4     |
| browser      |    Chrome     |

PR #881, opened 2019-08-29
Fix for 880

patorjk:
https://github.com/gregnb/mui-datatables/issues/880

* Updated how dropdown filters work so that only the "All" keyword is treated as a sign to remove the filters (thus allowing empty strings to be used as filter values).

Quick view: https://codesandbox.io/s/github/patorjk/mui-datatables/tree/empty-string-filter

The filterUpdate function only gets passed to 3 files: TableToolbar.js, TableFilter.js and TableFilterList.js. And only the latter 2 call it, and only TableFilter.js calls it with the type "dropdown" (TableFilterList.js removes filters by setting the type as "checkbox" or "custom", regardless of filterType).

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -936,6 +936,8 @@ class MUIDataTable extends React.Component {
             filterList[index] = value === '' ? [] : value;
             break;
           case 'dropdown':
+            filterList[index] = value;
+            break;
           case 'custom':
             filterList[index] = value;
             break;
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -935,6 +935,7 @@ class MUIDataTable extends React.Component {
           case 'multiselect':
             filterList[index] = value === '' ? [] : value;
             break;
+          case 'dropdown':
           case 'custom':
             filterList[index] = value;
             break;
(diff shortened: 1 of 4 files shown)


src/MUIDataTable.js
```diff
@@ -935,6 +935,7 @@ class MUIDataTable extends React.Component {
           case 'multiselect':
             filterList[index] = value === '' ? [] : value;
             break;
+          case 'dropdown':
```
gabrielliwerant: It could become a source of subtle bugs to allow `dropdown` to fall through to `custom` here. If this is intended as a placeholder, I recommend a `break` and if not, I think it's ok to duplicate what's in `custom`.

src/MUIDataTable.js
```diff
@@ -936,6 +936,8 @@ class MUIDataTable extends React.Component {
             filterList[index] = value === '' ? [] : value;
             break;
           case 'dropdown':
+            filterList[index] = value;
```
gabrielliwerant: Actually, do we need the `default` option anymore? `dropdown` used to fall into it, but now nothing will except invalid options, right? We could replace it with a warning or something.

patorjk: Yes, the type "textField" falls through and gets set there.

Issue #887, opened 2019-08-30
Column header "hint" should have a fallback for ADA compliance


Avd6977:
## Expected Behavior
Title to have a fallback

## Current Behavior
Empty column hint/title. Not ADA compliant.

## Steps to Reproduce (for bugs)
Don't pass in a column hint

## Your Environment

| Tech         | Version |
|--------------|---------|
| MUI-datatables  |   2.10      |

gabrielliwerant:
I think I need more information about this. Sort and hint are two separate things as far as column options are concerned. When sorting is possible, there is a `aria-sort` attr on the header. It's possible this is a dupe of https://github.com/gregnb/mui-datatables/issues/875, but even with that one, I'm still trying to figure out what the exact compliance issue is and what would be sufficient to solve it.

What would be helpful is documentation about this specific accessibility issue.

PR #888, opened 2019-08-30
#887 and#878

Avd6977:


gabrielliwerant:
See the issues themselves for more detail, but I'm unsure exactly what's needed.

src/components/TableHeadCell.js
```diff
@@ -68,6 +68,8 @@ class TableHeadCell extends React.Component {
     hint: PropTypes.string,
     /** Column displayed in print */
     print: PropTypes.bool.isRequired,
+    /** Tooltip label to fall back to */
+    label: PropTypes.string,
```
gabrielliwerant: This doesn't look like it's being used anywhere. Was this a holdover from the previous changes?

src/MUIDataTable.js
```diff
@@ -2,6 +2,7 @@ import Paper from '@material-ui/core/Paper';
 import { withStyles } from '@material-ui/core/styles';
 import MuiTable from '@material-ui/core/Table';
 import classnames from 'classnames';
+import assign from 'lodash.assign';
```
gabrielliwerant: We can eliminate this; as of latest master, we're not using `assign` anymore (it's `assignwith`), but also doesn't appear needed given the change sets here.

src/MUIDataTable.js
```diff
@@ -317,6 +318,9 @@ class MUIDataTable extends React.Component {
     if (props.options.responsive === 'scroll') {
       console.error('This option has been deprecated. It is being replaced by scrollMaxHeight');
     }
+    if (props.options.textLabels) {
```
gabrielliwerant: I'm not sure we need this as it looks like a holdover from a previous fix, but we're merging in the default `textLabels` in the props `textLabels` elsewhere now.

src/components/TableHeadCell.js
```diff
@@ -110,7 +112,11 @@ class TableHeadCell extends React.Component {
       <TableCell className={cellClass} scope={'col'} sortDirection={ariaSortDirection}>
         {options.sort && sort ? (
           <Tooltip
-            title={options.textLabels.body.toolTip}
+            title={
+              options.textLabels.body.columnHeaderTooltip
+                ? options.textLabels.body.columnHeaderTooltip(column)
+                : options.textLabels.body.toolTip
```
gabrielliwerant: I think `toolTip` is intended to be the default fallback option, right? If so, I think we'll need to add it to the default `textLabels.js` file or this errors out when `columnHeaderTooltip` isn't provided.

I think we'll also need to add documentation around proving `columnHeaderTooltip` in the localization section. Unless I'm misunderstanding, in which case, let me know.

Avd6977: No, you're right. Forgot to clean up the PR :) my bad.

src/MUIDataTable.js
Avd6977: Removed.

src/components/TableHeadCell.js
Avd6977: Forgot to change it with "column". Updated.

src/components/TableHeadCell.js
gabrielliwerant: No worries! Default is still missing from `textLabels.js`.

gabrielliwerant:
As far as documentation goes, we should add the new options to https://github.com/gregnb/mui-datatables/blob/master/README.md#localization.

Avd6977:
> As far as documentation goes, we should add the new options to https://github.com/gregnb/mui-datatables/blob/master/README.md#localization.

Added

README.md
```diff
@@ -340,6 +340,7 @@ const options = {
     body: {
       noMatch: "Sorry, no matching records found",
       toolTip: "Sort",
+      columnHeaderTooltip: "Sort for `${column.label}`"
```
gabrielliwerant: Can we use a function in the example, since that's how people would have to implement it? So more like `columnHeaderTooltip: column =>` `Sort for ${column.label}`.

Avd6977: Done.

src/components/TableHeadCell.js
gabrielliwerant: Looks like this is working now, I may have been missing something locally last time I found this problem.

Avd6977: Good to go then? Or do I still need to add something to textLabels.js?

gabrielliwerant: Nope, we're good to go.

Issue #871, opened 2019-08-27
Add `setCellHeaderProps`


mikroware:
Add `setCellHeaderProps` just like how `setCellProps` is there already. 

## Expected Behavior
`setCellHeaderProps` should work the same as `setCellProps`. It should be passing those props to the underlying `TableCell`. With this feature one can style the header cell just like the rest of the cells.

## Current Behavior
Does not exist for header cells (only for normal cells through `setCellProps`).

(feature request created as of discussion in #400)

marek-jarab:
Hi. I would like to start working on that . Could you pls assign me to this issue?

gabrielliwerant:
There is already a PR open for this feature, so I have to check on the status of that.

shawnmitchell:
Would it not make sense for `align` to be a first-class column option? 

```
const columns = [
 {
  name: "Name",
  options: {
   filter: true,
   sort: false,
   align: 'right',
  }
 },
 ...
];
```

austin557:
Has any progress been made on this?
I cannot get my column headers to center align while using the responsive: "scrollFullHeight" option.

gabrielliwerant:
Yes, this feature has been added as of 2.13.0, but `align` on the th won't help you because the column name is wrapped in `flex` display mode when there is a sort button. So you'll actually need to override styles instead: `MUIDataTableHeadCell sortAction` and set `justify-content`.

Issue #714, opened 2019-06-23
Add support for dense table


mohemos:
Just as material-ui dense table and bootstrap small table, Kindly support dense option to make tables more compact by cutting cell padding.

patorjk:
This looks like it could be done via a setTableProps option, similar to https://github.com/gregnb/mui-datatables/issues/871, if I have time, I'll see put a PR together tomorrow.

ltalirz:
Just in case anyone else needs this, the example https://github.com/gregnb/mui-datatables/issues/871 mentioned above uses

```javascript
      ...
      options: {
        setTableProps: () => {
          return {
            // material ui v4 only
            size: 'small',
          };
        },
       ...
```

alexkrytsky:
any updates on this? MUI 5 is out so I take that this isn't compatible anymore?

eplatek-safyr:
> any updates on this? MUI 5 is out so I take that this isn't compatible anymore?

It just works, all you need is trying ;)

emileWhispa:
> Just in case anyone else needs this, the example #871 mentioned above uses
> 
> ```js
>       ...
>       options: {
>         setTableProps: () => {
>           return {
>             // material ui v4 only
>             size: 'small',
>           };
>         },
>        ...
> ```

saved my day, thanks

alxndrsn:
Thanks @ltalirz.  I've updated your fix for MUI v5:

```javascript
      ...
      options: {
        setTableProps: () => {
          return {
            // material ui v5 only
            size: 'small',
          };
        },
       ...
```

Hubbsy:
I am trying to use this for setting my table density on click of a free-action.  I am just calling the setTableProps() method in the click handler of the free action.  But no luck.  Are there css styles that may conflict with this option setting?

Issue #857, opened 2019-08-23
Using onBlur to activate the filter instead of using onFilterChange


hungbang:
I am using server-side filtering, I overrode onFilterChange to activate filter every time the filter list is changed, but the problem is every time the filter list is changed, the backend API will be called to update the table's data. It makes the layout quite lag and the backend is called too much. I want to filter every time the user leaves the field or maybe we will add the button in TableList called `Submit Filter`, as soon as the user enters all the field they need to filter then click the button to activate the filter. 

| Tech         | Version |
|--------------|---------|
| Material-UI  |     3.9.2    |
| MUI-datatables  |      2.8.0   |
| React        |     16.8.5    |
| browser      |    Chrome     |
| etc          |         |

gabrielliwerant:
Right now, the way filtering and the callback work, it does not support server-side filtering. In my experience, filtering data is the exclusive domain of the UI, and I would expect it to operate only on what I'm looking at, at the moment, rather than change the data state first. Doing otherwise would tend to be very confusing from a user perspective. Maybe if I understood more about the specific needs of your project, I could empathize a bit more with your use case. Otherwise, I would consider this lower priority.

I don't think there's probably a need at some point to enhance the callback API to make it more clear when they fire (e.g. `onFilterChangeFinished`).

hungbang:
@gabrielliwerant thank you for your answer, I agree with you that we should have at least one event like `onFilterChangeFinished` or more simple solution is we will have one event called `onCloseTableFilterList`, as soon as the `TableFilterList` is closed we will call the API to reload data.

mbniebergall:
I'm also looking for a solution to the same problem @hungbang . Having one or more new events to be able to call my search API once the user has completed typing in the search field would be very helpful. Either a new event is needed, or the option to have a 'Search' button to trigger the search would help solve the problem.

hungbang:
@gabrielliwerant @mbniebergall 
Regarding the user experience, I think we should open to the user decides when is the search or filter performed. of course, the real-time update will be really helpful, but in some specific situation, we should have an option such as : 
`disableAutoSearch`: true or `disableAutoFilter: true`, when this flag is enabled, the button `Search` or `Filter` will be displayed.

patorjk:
[...]

Issue #681, opened 2019-06-10
How to set column width and row height?


emagombe:

roballsopp:
Apparently they aren't planning to support this: https://github.com/gregnb/mui-datatables/issues/92#issuecomment-403651830. There is at least one potential semi-hacky example of how to get around this in that thread. I'd just go for the native [material-ui tables](https://material-ui.com/components/tables/) if you have the flexibility.

gabrielliwerant:
The way to do this is the same way you'd do it with material-ui, which is to override mui classes with the styles you want, or provide your own components/styles via the custom render options (`customBodyRender`, `customRowRender`). This library is based off of the material-ui tables, so generally speaking, anything you can do with that library, you can do with this one. See `customize-rows` and `customize-styling` for some examples.

PR #889, opened 2019-08-30
Feature/issue 871 (setCellHeaderProps) and Feature/issue 714 (setTableProps)

patorjk:
https://github.com/gregnb/mui-datatables/issues/871 and https://github.com/gregnb/mui-datatables/issues/714

Added two new options:

* setTableProps (options)
* setCellHeaderProps (columns)

They're modeled after the existing setCellProps and setRowProps. The dense table is really only available in Material v4, so the example only removes some padding in Material v3.

Quick view: https://codesandbox.io/s/github/patorjk/mui-datatables/tree/set-props


patorjk:
I'm glad this is on your radar. Since discovering setCellProps and setRowsProps I've found them incredibly useful. They provide a lot of power and a nice reprieve from render functions. Some examples:

* Using setCellProps I'm able to eliminate 1,700 divs from one of my tables (previously I was wrapping every cell in a div with a style of "white-space:pre").
* Using setCellProps and setCellHeaderProps I'm able to have more concise code.
* setCellProps and setCellHeaderProps make the issue discussed here: https://github.com/gregnb/mui-datatables/issues/681 - completely trivial. 

Render functions technically allow for the same thing, but I find them dreadful and they're way more complex than would seem. For example, to override a cell header, you really need to add in most of the functionality of [TableHeadCell.js](https://github.com/gregnb/mui-datatables/blob/master/src/components/TableHeadCell.js), and there's a lot going on in there. I don't think it's reasonable to assume that most devs will take the care to reproduce what's in there, leading to bugs in their code.

Anyway, some time has passed since I put this in. setTableProps may be better off as "tableProps", as there's only 1 table. Additionally, if you decided you like this type of API, you may want to reconsider the searchProps idea of https://github.com/gregnb/mui-datatables/pull/808 (though that implementation is buggy - it's very close though). searchProps also makes a lot of customization trivial, and itsolves a lot of issues people brought up in https://github.com/gregnb/mui-datatables/issues/857#issuecomment-530179241.

src/MUIDataTable.js
```diff
@@ -1229,6 +1233,10 @@ class MUIDataTable extends React.Component {
         break;
     }
 
+    let tableProps = this.options.setTableProps ? this.options.setTableProps() : {};
```
gabrielliwerant: Maybe we should be more defensive here. In case the function doesn't return an object, it will error out below. What do you think?

src/components/TableBodyCell.js
```diff
@@ -12,18 +12,26 @@ const defaultBodyCellStyles = theme => ({
     [theme.breakpoints.down('sm')]: {
       display: 'inline-block',
       fontSize: '16px',
-      height: '24px',
```
gabrielliwerant: Could we hold off on the stylistic changes here or put in another PR? I believe there were some modifications here since you opened this that might have solved some of the style issues, but also, if we keep them separate, then we can discuss them separately and then they won't hold up the other work here.

src/components/TableHeadCell.js
```diff
@@ -89,7 +89,8 @@ class TableHeadCell extends React.Component {
 
   render() {
     const { isSortTooltipOpen, isHintTooltipOpen } = this.state;
-    const { children, classes, options, sortDirection, sort, hint, print } = this.props;
+    const { children, classes, options, sortDirection, sort, hint, print, cellHeaderProps } = this.props;
+    const { className, ...otherProps } = cellHeaderProps;
```
gabrielliwerant: Similar to the other comment about checking for the `className` property, as the dev will get an unhelpful error if they return the wrong type.

gabrielliwerant:
Understood about props vs render, I agree that custom render functions are a bad option when you want to replicate the functionality almost entirely, but with some slight changes. I think it would be a nice option to allow various parts of the table to be exported separately for just this kind of usage, but that's probably a ways away. I think render functions work best when the functionality being added is very custom and much different from exists, and probably has a bunch of nested components and such. But use cases like those you mention are better served via adding props, so it's nice to have both.

gabrielliwerant:
Hm, I don't see a big difference between `setTableProps`  and `tableProps` for the API. I think it's better to be consistent with the other options for now. Maybe there's an even better organizing principle that can be used for API naming, like how callbacks all begin with `on`. I've also been toying with the idea of adding a prefix for boolean options because as of now they are difficult to distinguish from many of the non-boolean options. So this is something I do think about.

Agreed about `searchProps` having valid use cases outside of the `customSearchRender`. I'll ping the PR creator in the next day or so and see if he wants to update it. If not, It'll be up for grabs.

src/MUIDataTable.js
patorjk: I don't know, that's not how the existing setCellProps and setRowProps functions work. Trying out other return values in the console, it doesn't look like anything breaks with the spread operator, they just wouldn't be setting any fields if they had the function return a non-object (with the exception of a "string" type, where they'd set numerical fields on the component).

src/components/TableBodyCell.js
patorjk: They're tied to the functionality a setTableProps method would add (ie, "dense table"). Without these changes, applying a dense table breaks stacked mode. There's actually bugs in the existing code, if you view the Customize Styling example you'll see one of them:

https://codesandbox.io/s/github/gregnb/mui-datatables/

The CSS solution I wound up with is simple. Initially I tried to be fancy, but it wasn't needed and didn't cover certain edge cases (basically a height parameter is needed for causes when cell values are empty or just a space).

src/components/TableHeadCell.js
patorjk: I'm not sure I follow. Check that it's set to a string?

patorjk:
I can circle back a little later for tests.

src/MUIDataTable.js
gabrielliwerant: The others don't work this way because they aren't checking an object for a specific property. The new functions you added are checking to see if `classNames` appears on the object, and this will error for non-objects, where the others won't, with the exception of strings.

src/components/TableBodyCell.js
gabrielliwerant: Right, so the problem is just that the PR is a two-for-one deal. In the future, I'd prefer to have them kept separate as this makes reviewing easier, but it's also much easier to trace back any future issues or regressions to a specific commit (since we squash merges) and enables us to rollback specific sets of functionality without losing others that work well, if need be.

gabrielliwerant:
@patorjk If you want to take on `setSearchProps`, go for it. I'm going to assume the other dev that submitted the old PR isn't checking up on it, since it's gotten old and he hasn't responded to new messages.

patorjk:
Sure, I'll put something together, I'll loop back to it later this week or next.

Issue #878, opened 2019-08-28
Text Label props


Avd6977:
## Expected Behavior
Passing in text label props to use defaults if not overridden

## Current Behavior
Passing custom text label props that don't include the default breaks loading
e.g.
```
textLabels: {
	body: {
		noMatch: this.props.emptyMessage
	}
}
```

## Steps to Reproduce (for bugs)
Pass in the prop above

## Your Environment

| Tech         | Version |
|--------------|---------|
| MUI-datatables  |    2.9.0     |

gabrielliwerant:
I wasn't able to reproduce the issue with the information here. I loaded the `customize-search` example from `v2.10.0` and then modified the options:

```
const options = {
  // ... a bunch of options
  textLabels: {
    body: {
      noMatch: 'empty message'
    }
  }
}
```

Nothing breaks for me when I do that.

nhattextiq:
I had this problem today as well. When I removed the `textLabels` part in my options, everything would work.
``` 
  const finalOptions = (
      Object.assign({},
        DEFAULT_OPTIONS,
        { textLabels: {
          body: {
            noMatch: isLoading
              ? <CircularLoading classes={classes} />
              : intl.formatMessage({ id: 'noRecord' })
          }
        } },
        otherOptions
      )
  )
```
Stack trace is as follow:
```
(t, ) TypeError: Cannot read property 'search' of undefined

Stack trace:
TypeError: Cannot read property 'search' of undefined
at t.value (webpack:///./node_modules/mui-datatables/dist/index.js?:1:45681)
at t.componentRender [as render] (webpack:///./node_modules/react-hot-loader/dist/react-hot-loader.development.js?:2243:64)
at finishClassComponent (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:17036:31)
at updateClassComponent (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:16991:24)
at beginWork$1 (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:18502:16)
at beginWork$$1 (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:23190:14)
at performUnitOfWork (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:22205:12)
at workLoopSync (webpack:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js?:22182:22)
[...]

PR #890, opened 2019-09-02
Issue 878 Overwriting Props

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/878.

We want to override options when the component updates to enable changes in props to be reflected in the table. However, this overrides entire object properties when we really want to keep any defined nodes.

The fix is to merge/assign based on the property.

Additionally, there is some light refactoring here to make the deprecation notices and changes easier to manage.

Aside from the written test, I used the following code, added to the `selectable-rows` example and toggled the switch to test table/prop updates.

```
state = {
  text: 'foo',
  separator: ':'
  rows: [10, 200, 500]
};

const options = {
  textLabels: {
    body: {
      noMatch: this.state.text
    }
  },
  downloadOptions: {
    separator: this.state.separator
  },
  rowsPerPageOptions: this.state.rows,
};

toggleLoading = event => {
  this.setState({
    separator: ';',
    text: 'bar',
    rows: [10, 200, 300]
  });
};

return (
  <div>
    <FormGroup row>
      <FormControlLabel
        control={
          <Switch
            checked={this.state.isLoading}
            onChange={this.toggleLoading}
            value="true"
            color="primary"
          />
        }
        label="Loading"
      />
    </FormGroup>

    <MUIDataTable title={"ACME Employee list"} data={data} columns={columns} options={options} />
  </div>
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6570,11 +6570,6 @@
       "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==",
       "dev": true
     },
-    "lodash.assign": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/lodash.assign/-/lodash.assign-4.2.0.tgz",
-      "integrity": "sha1-DZnzzNem0mHRm9rrkkUAXShYCOc="
-    },
     "lodash.assignwith": {
       "version": "4.2.0",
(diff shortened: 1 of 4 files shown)


patorjk:
Nicely done! I kicked the tires a little bit and this looks good! I especially like the way the way the deprecated options were handled.

gabrielliwerant:
@patorjk Thank you for taking a look!

cHewN:
We stumbled upon this bug in our project a few days back and upgrading to the latest version solved it. Thanks for your work @gabrielliwerant

PR #893, opened 2019-09-03
Feature/Allow to display search bar from init

Vallher:
Added option to show search from init. (in my case I want to resign from using searchIcon hide/display). In current version it is only possible when searchText is defined, but I want to have placeholder at init.


brycewalter-univa:
Bump, I also require this for my use case

brycewalter-univa:
Bump one more time, this feature is critical to a use case

README.md
```diff
@@ -165,6 +165,7 @@ The component accepts the following props:
 |**`sort`**|boolean|true|Enable/disable sort on all columns
 |**`filter`**|boolean|true|Show/hide filter icon from toolbar
 |**`search`**|boolean|true|Show/hide search icon from toolbar
+|**`showSearch`**|boolean|false|Initially displays search bar
```
gabrielliwerant: What do you think about calling the option `searchOpen` instead? My concern is that I want to distinguish it from the numerous search options that exist now, and I worry it may cause confusion with the option that shows/hides the icon form the toolbar.

acberk:
@Vallher I've made the above changes in a pull request to your fork.
https://github.com/Vallher/mui-datatables/pull/2

Vallher:
Ups. I merged above changes but now I see that there was to much things renamed. I will fix it in 1-2h.

Vallher:
Thanks @acberk. Now should be okay:D

nikbakhtenb:
Hi. I really appreciate for this amazing table. 
As it was my problem too , when it will be release?

PR #910, opened 2019-09-10
No Pointer style cursor for table head when no hint or sorting

lalong13:
With no sorting or a hint, the cursor still shows a pointer. I'm not saying this is the best solution, but I think the header shouldn't shouldn't show a pointer style of cursor when there is nothing to click on or additional behavior from hovering.

### DIFF
--- a/src/components/TableHeadCell.js
+++ b/src/components/TableHeadCell.js
@@ -27,10 +27,6 @@ const defaultHeadCellStyles = theme => ({
   data: {
     display: 'inline-block',
   },
-  noSortAction: {
-    display: 'flex',
-    verticalAlign: 'top',
-  },
   sortAction: {
     display: 'flex',
     verticalAlign: 'top',
@@ -165,7 +161,7 @@ class TableHeadCell extends React.Component {
             </span>
           </Tooltip>
         ) : (
-          <div className={hint ? classes.sortAction : classes.noSortAction}>
+          <div className={hint ? classes.sortAction : null}>
             {children}
             {hint && (
               <Tooltip
--- a/src/components/TableHeadCell.js
+++ b/src/components/TableHeadCell.js
@@ -27,6 +27,10 @@ const defaultHeadCellStyles = theme => ({
   data: {
     display: 'inline-block',
   },
+  noSortAction: {
+    display: 'flex',
+    verticalAlign: 'top',
+  },
   sortAction: {
     display: 'flex',
     verticalAlign: 'top',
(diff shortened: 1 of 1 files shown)


lalong13:
@gabrielliwerant Would you be looking to get this change in soon? Is there anything I need to do with this? I imagine you may want to do this change in less of a hacky type fix, but I can make changes to the PR to help get it in faster.

gabrielliwerant:
@lalong13 Just about to look at a few PRs right now. Will keep you updated.

src/components/TableHeadCell.js
```diff
@@ -161,7 +165,7 @@ class TableHeadCell extends React.Component {
             </span>
           </Tooltip>
         ) : (
-          <div className={classes.sortAction}>
+          <div className={hint ? classes.sortAction : classes.noSortAction}>
```
gabrielliwerant: What if we just do `hint ? classes.sortAction : null`? Then we don't even need replacement styles and it looks like it prevents the pointer cursor from showing.

gabrielliwerant: @lalong13 Any interest in submitting that change? If not, I can open up something based on this, just wanted to make sure before using your work or going a different way.

lalong13: @gabrielliwerant Sorry for the delay. I had assumed that display and verticalAlign style properties were still desired. I don't mind not giving it a class name personally. It just might be weird if the other stylings change.

gabrielliwerant: I appreciate the cautious approach, but those styles are only needed when the div acts as a container for other things, like sort and hint. They serve no useful function when we don't have a sort or hint.

If some other thing is added here in the future, we can add styles back, but for now, I think it's cleaner to just remove them.

lalong13: I made the suggested change. I can roll the commits up as well if needed/desired.

gabrielliwerant: Excellent, will verify soon! What do you mean by "roll the commits up?"

lalong13: Make the two commits into just one commit.

gabrielliwerant: Oh, gotcha. No need, I squash them when I merge.

lalong13: @gabrielliwerant Do I need to do anything else for this at this point?

gabrielliwerant: Not sure, I need to re-checkout and look everything over, just a bit backed up with PRs and such, but I'll swing around to it before the next minor release.

PR #929, opened 2019-09-13
Fix crash on downloading an empty dataset as CSV

Skn0tt:
When downloading an empty dataset as CSV, there is a crash because you cannot `.trim()` an array.

This PR adds a fix and a regression test.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -28,7 +28,7 @@ function sortCompare(order) {
 
 function assembleCSV(columns, data, options) {
   const replaceDoubleQuoteInString = columnData =>
-  typeof columnData === 'string' ? columnData.replace(/\"/g, '""') : columnData;
+    typeof columnData === 'string' ? columnData.replace(/\"/g, '""') : columnData;
 
   const buildHead = columns => {
     return (
@@ -56,7 +56,7 @@ function assembleCSV(columns, data, options) {
             .map(columnData => replaceDoubleQuoteInString(columnData))
             .join('"' + options.downloadOptions.separator + '"') +
           '"\r\n',
-        "",
+        '',
       )
       .trim();
   };
--- a/src/utils.js
+++ b/src/utils.js
@@ -66,6 +66,11 @@ function assembleCSV(columns, data, options) {
     ? options.onDownload(buildHead, buildBody, columns, data)
     : `${CSVHead}${CSVBody}`.trim();
 
+  const downloadHasBeenAborted = csv === false;
+  if (downloadHasBeenAborted) {
+    return null;
+  }
+
   return csv;
 }
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -26,9 +26,9 @@ function sortCompare(order) {
   };
 }
 
-function createCSVDownload(columns, data, options) {
+function assembleCSV(columns, data, options) {
   const replaceDoubleQuoteInString = columnData =>
-    typeof columnData === 'string' ? columnData.replace(/\"/g, '""') : columnData;
+  typeof columnData === 'string' ? columnData.replace(/\"/g, '""') : columnData;
 
   const buildHead = columns => {
     return (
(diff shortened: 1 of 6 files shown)


Skn0tt:
If somebody needs this fix before it's reviewed, they can use the published PR package: https://github.com/Skn0tt/mui-datatables/packages/28551

Skn0tt:
Hey @gabrielliwerant, thanks for reviewing my code.

I tried to make this PR easy to review by dividing up my thoughts into individual commits so it can be reviewed step by step.
I consider regression tests as necessary. Please have a look at the code before I extracted parts of it - there were side effects all over the place, basically untestable. That's why I decided to refactor it.
Do you mean the removal of unused imports by "removing old code"? This was one of the warnings that eslint gave me.

I see that `if (!data.length) return '';` would have been easier to review since it perfectly fits the stated issue.
On the other hand, I consider the underlying issue to be in the initial value of the `reduce` call (it just doesn't make sense to use `[]` and then make a string out of it after the first run, does it?)

src/utils.js
```diff
@@ -66,15 +66,15 @@ function createCSVDownload(columns, data, options) {
     ? options.onDownload(buildHead, buildBody, columns, data)
     : `${CSVHead}${CSVBody}`.trim();
 
-  if (options.onDownload && csv === false) {
```
gabrielliwerant: I think we still need this, as this allows people to switch off download functionality by returning `false` in the callback.

test/utils.test.js
```diff
@@ -0,0 +1,55 @@
+import { assembleCSV } from '../src/utils';
+import { expect } from 'chai';
+
+describe('utils', () => {
```
gabrielliwerant: Can we use `'utils.js'` here just to avoid any potential ambiguity?

Skn0tt: Fixed in 07c797ea6cedbe1e9583d113d3ccd73cd6b54fa8

src/utils.js
Skn0tt: Fixed in 9343413843cd3d375aa6f3ad0799f2f0135342e9.

Skn0tt:
Thanks a lot for your response. I totally understand that you want to keep the code base consistent. I've made the requested changes and converted from using `expect` to comply wit `assert` usage.

src/utils.js
```diff
@@ -74,15 +74,20 @@ function createCSVDownload(columns, data, options) {
...
-    return;
+  const downloadHasBeenAborted = csv === false;
```
gabrielliwerant: I'm not sure that assigning to a variable here that is only used once just below is clearer than using `options.onDownload`. This part wasn't broken before, so I think let's just leave it as it was for now.

src/utils.js
```diff
@@ -74,15 +74,20 @@ function createCSVDownload(columns, data, options) {
...
+  if (downloadHasBeenAborted) {
+    return null;
   }
 
+  return csv;
+}
+
+function downloadCSV(csv, filename) {
```
gabrielliwerant: Great idea to break this out into its own function! Helps a lot with the testing.

test/utils.test.js
```diff
@@ -1,60 +1,127 @@
...
+        name: 'firstname',
+        download: true,
+      },
+      {
+        name: 'lastname',
+        download: true,
+      },
+    ];
+
+    describe('when given two rows', () => {
```
gabrielliwerant: So, I think we still have too many nested `describe`s. If you look at the `getPageValue` tests, you'll see that we just use one `describe` block for the whole function, breaking up the bits into `it` blocks instead. So, this one would become `it('renders a two-record csv when given two rows`... Let's do that for all the new tests here, just one `describe` for `assembleCSV` and everything else is an `it` block.

gabrielliwerant:
I put this on hold and continued in a separate PR here https://github.com/gregnb/mui-datatables/pull/1064. Feel free to take a look and offer feedback.

gabrielliwerant:
Fix should now be in `2.13.0`.

Issue #928, opened 2019-09-13
tableData changes if sorted/filtered


wroughtec:
#923 fixes the issue of data going missing when sorting etc. however it now appears to change from an array of arrays to an array of objects:

https://codesandbox.io/embed/muidatatables-custom-toolbar-txw6d

(added a couple of 'bugged' examples to give you an idea what I was achieving before)

The reason having the original index was so useful was I could enrich the table with extra data that I don't necessarily want to display in the table ie add tooltips or a link etc.

If you click sort now you will see that I have added a link but as its the data order and rowIndex don't match the link always stays on the first row. I don't expect the index to make an appearance where it was but it would be useful it it was returned somewhere.

gabrielliwerant:
Ok, I see that. It's happening when arrays of objects are fed to the table. It doesn't happen with arrays of arrays. I'll work on it.

You can still add all of those things by using `customBodyRender` and mapping your data to whatever else you need. It was never a supported feature to allow the internal table state to store a bunch of data that isn't be used by the library. That's likely to cause breakage in a number of areas, and it certainly may break in the future as the data structure might change, or add things, or remove things, etc. I know it's unfortunate in that there was a loophole that appeared to be helping you out, but it's just not reliable, and it's better to catch it sooner rather than later.

wroughtec:
Sure. Got to love a loophole ;)

gabrielliwerant:
Haha! It looks like I may need to go back to array of objects, but only with the index/dataindex values, as it might be the only way to have consistency here. But again, if you add other props there, you do so at your own risk! :p

PR #933, opened 2019-09-17
Force consistency with table metadata between array and object data

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/928.

Sending in array of arrays data vs array of objects data into the table was resulting in different `tableData` in the `tableMeta` object. Now, either should result in data that matches passing in array of arrays `tableData`.

It's possible this will be cause breakage for people relying on the inconsistent behavior, so saving this for a major release, unless I can trace it back as a relatively new issue, at which point, it could be released with a minor release.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -480,7 +480,8 @@ class MUIDataTable extends React.Component {
 
         if (typeof column.customBodyRender === 'function') {
           const rowData = tableData[rowIndex].data;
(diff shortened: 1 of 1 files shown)


patorjk:
Added manually to v3.

patorjk:
Actually, upon further testing, this update destroys performance with larger data sets. I'm going to leave this inconsistency in the table for now until a better performing work around can be developed.

patorjk:
Issue resolved in https://github.com/gregnb/mui-datatables/pull/1404 (new tableMeta parameter currentTableData should be used).

PR #951, opened 2019-09-21
Add support for custom sort function for columns

GuyShaanan:
Adds a `sortFn` prop to `column.options`.
When given, the column sorting will be made using that function.

sortFn Interface
---------
```typescript
(order: 'asc'|'desc') => ({data: val1}, {data: val2}) => number
```

example:
```javascript
columns = [
  {
    name: "Name",
    options: {
      sort: true,
      sortFn: order => ({data: val1}, {data: val2}) => (val1.length - val2.length) * (order === 'asc' ? 1 : -1)
    }
  }
]
```

Gotchas
---------
1. Doesn't work when `customSort` is provided (for the whole table).
The reason is that there is a specially crafted payload when calling `sortedData.sort(...)` (whereas usually we would expect an interface of `(a,b) => number` where `a`,`b` are the actual values).
2. `sortFn` (and `customSort`) need to know about the possible values of `order` instead of accounting for the comparison only. I think this should be refactored, something like:
```typescript
const sortWrapper = (order: number, compareFn: function) =>({data: val1}, {data: val2}) => (order === 'asc' ? 1 : -1) * compareFn(a,b)
sortedData.sort(sortFn(order, compareFn))
```
That way neither the custom `sortFn` (or `sortCompare`) need to know about `order`.

Any thoughts?


gabrielliwerant:
Hey @GuyShaanan, this is an interesting feature, thanks for working on it. I'm also glad you provided an example, and that certainly helped me understand the potential use cases better.

Here are some of my thoughts:

1. I would expect a custom column sort to override the behavior of the overall custom sort in cases where that column is specifically being sorted. I think we should be able to do this by checking if the `columnSortFn` argument is `null`. If it isn't, I think we should resort the table in that case.
1. It's an interesting proposal to handle `order` such that the custom functions don't need it, however, there are probably existing use cases for treating `asc` and `desc` in a special way, so I'm wary of changing this. Already, your sample code and the `customize-sort` example don't agree on how to value the order.
1. I'd prefer a name for the function that would keep it a bit more obviously distinct and consistent with the existing `customSort`: perhaps `customColumnSort` would do.
1. As far as the function shape/arguments, I'd also prefer this to use the same format (for consistency) as `customSort`, meaning just passing all the arguments in at once, without currying.

patorjk:
This is a pretty cool feature, and it greatly simplifies custom sorting. This feature has been mentioned a few times in the last 2 months, so I'm merging it into another branch to finish it up for the 3.3.0 release.

Issue #466, opened 2019-03-05
Support for load more option


anudeepjusanu:
Hi do we have a support for load more option instead of normal pagination?

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

gabrielliwerant:
You can provide this option for yourself with `customFooter`. All the tools are available to create such a button and insert it into the table. There are no plans currently to add any specific implementations for something like this, as is the case with most optional renderings.

Issue #361, opened 2019-01-11
Why mui-datatables do not have loading layer


Aaron-Lei:
## Expected Behavior

   In the server-side mode, when do pageChange or else, a loading layer upper table body is needed.
   This is a common and good interaction. 
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

jkeruzec:
I'm also interested in this feature when data pass to datatable is fetching but not yet available. For the moment we have the basic text : "Sorry, no matching records found". 

It would be interesting to have a property like "loadingText" to display a waiting message or icon while data is loading.

Aaron-Lei:
> I'm also interested in this feature when data pass to datatable is fetching but not yet available. For the moment we have the basic text : "Sorry, no matching records found". This text is sometimes "blinking" when loading is fast.
> 
> It would be interesting to have a property like "loadingText" to display a waiting message or icon while data is loading.

@jkeruzec Thank you for your attention, I am very concerned whether this demand is on your development agenda. I urgently need this demand. Thx

paianoa:
I need it too :(

kostin-softgrad:
It is very need

AlexYenkalov:
+1

gabrielliwerant:
I'm not 100% sure what you mean by a "loading layer." If you mean things like text or progress bars, there is no intention to add these to the library on data fetching because you can already do this yourself. See the `serverside-pagination` example for examples of how to do this.

PR #958, opened 2019-09-25
Add a loading option

joshleblanc:
This adds a loading option to the table. It defaults to false.

When true, it adds a loader on top of the current table.

ZnjLHg3p3n

This is to facilitate server-side data loading.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -17,11 +17,15 @@ import TableResize from './components/TableResize';
 import TableToolbar from './components/TableToolbar';
 import TableToolbarSelect from './components/TableToolbarSelect';
 import textLabels from './textLabels';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import { fade } from '@material-ui/core/styles/colorManipulator';
 import { buildMap, getCollatorComparator, sortCompare } from './utils';
 
 const defaultTableStyles = theme => ({
   root: {},
-  paper: {},
+  paper: {
+    position: 'relative'
+  },
   tableRoot: {
     outline: 'none',
   },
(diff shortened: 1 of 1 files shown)


loktar00:
The same can be accomplished by providing a loader (custom ect.) Right before `<MUIDataTable/>`

Quick dirty example using the serverside pagination as an example.

```javascript
const loadingComponent = (
      <div style={{position: 'absolute', zIndex: 110, top: 0, left: 0, width: '100%', height: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center', background: 'rgba(255,255,255,0.8)'}}>
        <CircularProgress size={24} />
      </div>
    );

    return (
      <div style={{position:'relative'}}>
        {isLoading && loadingComponent}
        <MUIDataTable title={<Typography variant="title">
          ACME Employee list
          </Typography>
          } data={data} columns={columns} options={options} />
      </div>
    );
```

gabrielliwerant:
This isn't a feature that I think should be supported at this time. It has come up a few times (Eg: https://github.com/gregnb/mui-datatables/issues/361 & https://github.com/gregnb/mui-datatables/issues/466). It's easy enough to add this in and around the table without making it core functionality (see `serverside-pagination` example).

The library needs to leave out some things to keep it from becoming bloating and overcomplicated. When it comes to additional component rendering, I think it's best left to custom renders or author-provided functionality.

anthonywebb:
@gabrielliwerant it would be helpful in your `serverside-pagination` example if the loading indicator placement was not covered/hidden by the search box.  Sure adding it outside the "paper" works, but does not feel right.  I think a useful example of this as the example you reference attempts to be would be a good starting point for those wanting to elegantly add a loading indicator.

gabrielliwerant:
@anthonywebb Agreed, the examples provided are not the best. Would you have any interest in submitting something to enhance the example?

henrique-exacta:
Is It ok?

thefat32:
As opposed to what has being said in this thread. I think load state should be part of server-side data source support. +1 to get this implemented.

Thanks for this awesome lib btw!

ShawnXU0208:
I also want to have this feature added, I am trying to have a loading indicator appear in the center of table body, and while it is showing, all the actions (changing page, filters, search) should be disabled, there is currently no easy way to do this.

Issue #963, opened 2019-09-27
Using download option useDisplayedRowsOnly is causing issues with non-react objects like arrays


gabrielliwerant:
Originally brought up here https://github.com/gregnb/mui-datatables/issues/709#issuecomment-535637754

GuyShaanan:
Thanks.
I think I could sort it out by setting `dataIndex: index` in https://github.com/gregnb/mui-datatables/blob/f0ea9e58d82839c57a4dfd4a342b6649932660a7/src/MUIDataTable.js#L791
Some tests failed but I think they need to be changed as well.
(Not sure what are the side effects of this, but from skimming the examples they all seem to work fine).

gabrielliwerant:
I've got a PR open that closes the issue (linked here) so I don't think we need to worry about potentially breaking tests with a different fix.

PR #964, opened 2019-09-27
Fix issue by properly checking for basic js objects and preserve displayed index

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/963.

Also closes https://github.com/gregnb/mui-datatables/issues/973.

### DIFF
--- a/src/components/TableToolbar.js
+++ b/src/components/TableToolbar.js
@@ -105,8 +105,9 @@ class TableToolbar extends React.Component {
             data: row.data.map(column => {
               i += 1;
 
-              // if we have a custom render, we must grab the actual value from data
-              return typeof column === 'object' ? data[row.dataIndex].data[i] : column;
+              // if we have a custom render, which will appear as a react element, we must grab the actual value from data
+              // TODO: Create a utility function for checking whether or not something is a react object
+              return typeof column === 'object' && column !== null && !Array.isArray(column) ? data[row.dataIndex].data[i] : column;
             }),
           };
         });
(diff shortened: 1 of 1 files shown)


src/components/TableToolbar.js
```diff
@@ -105,8 +105,9 @@ class TableToolbar extends React.Component {
             data: row.data.map(column => {
               i += 1;
 
-              // if we have a custom render, we must grab the actual value from data
-              return typeof column === 'object' ? data[row.dataIndex].data[i] : column;
+              // if we have a custom render, which will appear as a react element, we must grab the actual value from data
+              // TODO: Create a utility function for checking whether or not something is a react object
+              return typeof column === 'object' && column !== null && !Array.isArray(column) ? data[row.dataIndex].data[i] : column;
```
GuyShaanan: So, this doesn't resolve the original bug.
Here the "faulty" column is
```javascript
{$$typeof: Symbol(react.element), key: null, ref: null, props: {…}, type: ƒ, …}
```
which is not of type array, so we do the wrong calculation again.

gabrielliwerant: Not sure which bug you mean. The only bug I've been able to reproduce is one in which, with `useDisplayedRowsOnly` is set to `true`, some of the data is in array form, the table is sorted, and then the csv is downloaded. In that case, the arrays were not returning the correct values and it showed up to me as exactly the mismatched downloaded csv in your example. It definitely fixes this exact set of circumstances. If there is another set of circumstances, then I would need those steps spelled out and demonstrated in another issue.

GuyShaanan: Sorry, my bad. I just checked out your branch and everything indeed works!

Just wanted to say thanks for this library, it saves a lot of boilerplate and was really easy to set up.

gabrielliwerant: Thank you, on behalf of the community! Glad to hear it's working for you.

GuyShaanan:
Sorry about nagging - What is the release cycle of this lib / when should we expect this fix to land?

PR #968, opened 2019-09-30
Exclude toolbar on printing and update material-ui to latest version

andfs:
When printing, hide the toolbar and it's icons, besides the selection column, so only data will be printed.
In this pull request, material-ui and react was updated to the latest version.
Unfortunately, I have prettier installed in my vscode and the files were formatted. If it would be a problem I can make another pull request without formatting.


gabrielliwerant:
Hey @andfs, thanks for submitting this! That's a good catch about the display of the toolbar on print, though it might be best to release it in a major so that print behavior doesn't change in an unexpected way before then.

It would actually be helpful if you could open a PR without the formatting changes (or revert them here; we squash and merge so there won't be any excessive commit history). This library uses `prettier`, but I think the configs for this library are probably different than yours, as there is a lot of diff noise that would probably be reverted when I run prettier on it before a release.

You can run `npm run prettier` before each of your commits to make sure that the formatting is in line for this library, which would help reduce diff noise even further.

andfs:
Hello @gabrielliwerant 
I made some changes to revert my prettier formatting, but I was not able to squash my previous commits.
Let me know if we really need to squash it and how I can do this.

gabrielliwerant:
Oh, sorry if I wasn't clear @andfs, you don't need to squash them, I'll do that during the merging process. I only mentioned it so that you would know that adding a revert wouldn't be a problem.

andfs:
Oh sorry for misunderstood you haha
Anyway, with my last push, I think the merge will be easier
Thank you :)

patorjk:
Closing as a new version of this PR was submitted here - https://github.com/gregnb/mui-datatables/pull/1269

Issue #969, opened 2019-09-30
Improve accessbility


michalica:
Allow user to "tab" among rows and allow him to open and close expandable row with Space or enter 

## Expected Behavior
    We should allow user tabbing among rows.
    We should wrap SVG icon which opens expandable row with HTML element button:
`<button><svg></svg></button>
`
## Current Behavior
It in not wrapped -> tabbing doesnt work.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |     4.0.1    |
| MUI-datatables  |     2.1.2    |
| React        |    latest      |
| browser      |    all     |

gabrielliwerant:
Available in `2.12.0`!

PR #970, opened 2019-09-30
adding button wrapper for icon

michalica:
I want to improve accessbility of expandable rows in  https://github.com/gregnb/mui-datatables/issues/969


michalica:
Can you please include this PR in next release? @gabrielliwerant

michalica:
I would be glad to work on this. I think pressing space or enter on focused element triggers onClick event by default. In our case, row will be expanded or collapsed. Do we need onKeyDown handler there?

gabrielliwerant:
You are correct, it looks like the default is to trigger on space and enter.

Ok, so let's just switch it to `IconButton` and we should also disable it for the header or we'll have an odd looking empty header button as well. Something like `disabled={isHeaderCell}` should work.

Issue #966, opened 2019-09-29
sortDirection cannot be applied for more than one column.


T-pirithiviraj:
If we apply sortDirection for more than one column, it is throwing an error.

## Expected Behavior
sortDirection should work for all the column in case we update the state.

## Current Behavior
Open the console, apply sort direction for more than one column it is throwing an error as sortDirection can be applied for only one column.

## Steps to Reproduce (for bugs)

1. open the console.
2. In the state, define sortDirection in an array.
3. Include this sort direction from state in column.
4. Now reload the dataTable, It is throwing an error as sortDirection can applied for only one column, if we try to use more than one column it will take the first column only.
Sample code:
```
this.state: {
             page: 0,
          count: 1,
          changedColumn: null,
          sortDirection: ['asc','none','none','none','none','none','none', 'none'], ...
     }
sortColumnChange(sortDirection) {
this.setState ...
}
render() {
const columns = [
            {
                name: 'reAlias',
                label: 'Alias',
                options: { sortDirection: this.state.sortDirection[0] }
            },
            {
                name: 'cookieGroup',
                label: 'Group',
                options: { sortDirection: this.state.sortDirection[1] }
            }, ...
}
## Your Environment
<!--- Include as many relevant details about the environment with which you experienced the bug. -->

| Tech         | Version |
|--------------|---------|
| Material-UI  |  |
| MUI-datatables  |    2.11.1     |
| React        |         |
| browser      |     chrome    |
| etc          |         |

gabrielliwerant:
In order to apply a new sort direction, you have to remove one of the existing sort directions. Sorting along one than one column at once is not possible, as I understand it (not possible even in theory).

danielbyun:
Hi @gabrielliwerant ,

seems pretty redundant he's creating new issues for the basically same issue, but I'm not sure I'm understanding correctly.

I understand we have to keep the values in the state that we want to track while using `serverSide = true`, and I am doing that to toggle from 'asc' and 'desc' when listening to an event from `onColumnSortChange` from the options object I'm defining.

example:
      onColumnSortChange: (changedColumn, direction) => {
        console.log(changedColumn, direction);
        let order = "desc";

        if (direction === "ascending") order = "asc";
        if (direction === "descending") order = "desc";

        switch (changedColumn) {
          case "id":
            console.log(`changedColumn: ${changedColumn}, order: ${order}`);
            return this.sortChange(changedColumn, order);
          case "name":
[...]

PR #971, opened 2019-10-01
Allow sort direction of "none" to be applied multiple times

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/966.


T-pirithiviraj:
@gabrielliwerant , When this will change be available for us to use?

gabrielliwerant:
It should be available now in `2.12.0`.

PR #986, opened 2019-10-06
Serverside sorting example

danielbyun:


examples/serverside-sorting/index.js
```diff
@@ -0,0 +1,215 @@
...
+          customFilterListRender: v => `Title: ${v}`,
+        },
+      },
+      {
+        name: 'Location',
+        options: {
+          sortDirection: this.state.columnSortDirection[2],
+          customFilterListRender: v => `Location: ${v}`,
+          customBodyRender: (value, tableMeta, updateValue) => {
+            return <Cities value={value} index={tableMeta.columnIndex} change={event => updateValue(event)} />;
```
gabrielliwerant: Since the initial data is just a loading message, we're missing `value` at first, which is causing a prop check verification error in the console. We could fix this by using `value || ''`.

danielbyun:
Hi, added empty string for the props check.

Thank you!

PR #995, opened 2019-10-10
Refactor column direction conversion

### LINKED ISSUE #151: Unusual conversion of sort order string in onColumnSortChange
Column sorting direction uses 'asc' and 'desc' internally however getSortDirection() converts this to 'ascending' and 'descending'.

I can understand why you would want to convert to log form for the announce text, which is one of the uses of getSortDirection().

However in the onColumnSortChange callback I would generally want the value to be the same as the value that is used internally.

Maybe getSortDirection() should actually be called getSortDirectionLabel() and only be used for the announce and other interface specific stuff?

bkiac:
Fixes #151

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -824,7 +824,7 @@ class MUIDataTable extends React.Component {
     );
   };
 
-  getSortDirection(column) {
+  getSortDirectionLabel(column) {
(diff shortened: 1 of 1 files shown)


patorjk:
I agree, that callback should be using "asc" or "desc". This is a breaking change, I've put it in the v3 branch.

patorjk:
Also, thank you for noticing this!

PR #997, opened 2019-10-11
Stop mutating column.options props and support immutable data object

n3tr:
I notice that the `MUIDataTable` is mutating the props.columns.options value and it causes error when passing an immutable object to it. 

Example, I'm using `immer` to update state and it produces immutable data object (using deep-freeze under the hood).

I think it is better to not mutate the props values inside the component.

**Change**
- Replace code that mutate `props.columns.options` by spreading `column.options` into new options before mutating

**Example**
https://codesandbox.io/s/dazzling-dijkstra-kmuf2

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -409,7 +409,7 @@ class MUIDataTable extends React.Component {
       };
 
       if (typeof column === 'object') {
-        const options = { ...column.options }
+        const options = { ...column.options };
         if (options) {
           if (options.display !== undefined) {
             options.display = options.display.toString();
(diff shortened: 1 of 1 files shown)


n3tr:
@gabrielliwerant Fixed :)

Sorry, I forgot to run test before push the change. Didn't realize it breaks a test case.

gabrielliwerant:
Looks like you accidentally push your local yarn lockfile with the last commit! I noticed when the line changes jumped by about 9k. :)

n3tr:
@gabrielliwerant removed 😅

gabrielliwerant:
@n3tr Hm, I guess that wouldn't hurt. Could also take it out if we switched to yarn in the future. I'll leave that up to you.

n3tr:
@gabrielliwerant I will open separate PR for removing `yarn.lock`.

Do you know why TravisCI doesn't send the build status here?

gabrielliwerant:
@n3tr Not sure. I actually don't have access to the CI administration for this library. My guess is that it just wasn't set up.

n3tr:
@gabrielliwerant look like TravisCI deprecated Github commit status (https://blog.travis-ci.com/2018-09-27-deprecating-github-commit-status-api-for-github-apps-managed-repositories) and require owner to activate Travis CI app for this repo.

Issue #1011, opened 2019-10-17
Cannot accept objects for cell data - Date


wroughtec:
Unfortunately the fix introduced in 2.12.2 to validate data is either a string or an int causes an issue with things like Dates which is seen as an object. Previous versions this was not an issue.

I can work around this issue by converting to a timestamp (to avoid borking sort) prior to hitting the view but either there should be a check if its an object and not a Date or document that raw dates can not be passed as I am sure this may catch a few people out

gabrielliwerant:
Understood, will look into it.

gabrielliwerant:
So, when I use a date object as data in the table, I get the `Uncaught Invariant Violation: Objects are not valid as a React child` error, which is what the fix in `2.12.2` is preventing from occurring. Using date objects in the same way as other objects in the data should be giving you the same problems. So I'm actually not sure how you were able to do this before. Can you share the relevant bits of your code so I can understand how you were using it?

wroughtec:
Ok let me get up some examples that should 'hopefully' show how I have it working. I do wonder if its because I use the custom body render.

wroughtec:
This shows how I have been doing it prior to 2.12.2:
https://tv8fd.csb.app/

This is with 2.12.2:
https://70o59.csb.app/

Potentially  (am thinking without looking at the code) the check should maybe only happen on those fields that do not have a custom render.

drabelo:
Also getting this error ... Everything was working fine until today. I have a custom button as a cell and now it doesn't allow it. How do you fix that?

wroughtec:
do you have a demo? You  may be able to use the customBodyRender

lalong13:
Also seeing this error. Using a moment object that can be formatted with `customBodyRender`.

gabrielliwerant:
So, this type of functionality was never officially supported. The docs tell you what data is ok to use:
![image](https://user-images.githubusercontent.com/1779564/67122706-6872a400-f1bc-11e9-950e-729e0dafadf9.png)

It seems to work ok when you have a custom render function that knows just how to treat it, but without that, it breaks. The way to do this is to keep object data separate, if you have it, use an id of some kind in the actual table data, and then map and transform it how you wish in the custom render function.

gabrielliwerant:
Here is a modification of your sandbox example to show what I mean (one possible way of avoiding objects in the table data but still getting the values you want): https://codesandbox.io/s/muidatatables-custom-toolbar-bsb81?fontsize=14

[...]

PR #1012, opened 2019-10-17
is valid date

wroughtec:
add check to see if object is a js date object

fixex #1011


gabrielliwerant:
I don't think we'll need this anymore, as I've changed to a console error in cases where an object is being used. Can you confirm?

Also, when the ability to use objects is prevented (to fix other issues), this will be a wontfix.

PR #1015, opened 2019-10-20
Update overrides README for TypeScript

Imballinst:
Hi! Thanks for this great library. I've been using it for quite some time now, and recently I have the need to customize the style (in this case, overriding). But, turned out, my IDE yelled at me that there was no overrides such as `MUIDatatable*`.

When I first time encountered this error, I had absolutely 0 idea on how to fix this (there is no issue around this I believe). Luckily, I already knew how to extend Material UI's interfaces with `declare module` syntax. So, I just needed to find which module in Material UI that contains the overrides typings and re-declared it in my own project.

I hope this little PR can help other people to override the styles of this library properly in TypeScript.

Signed-off-by: Try Ajitiono <ballinst@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,8 +323,6 @@ class BodyCellExample extends React.Component {
 If you are using TypeScript, you will need to add this as well:
 
 ```ts
-import { Palette } from '@material-ui/core/styles/createPalette';
-
 declare module '@material-ui/core/styles/overrides' {
   interface ComponentNameToClassKey {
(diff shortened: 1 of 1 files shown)


patorjk:
Sorry for the delay in someone reviewing this but thank you for taking the time.

This PR seems very IDE and TS specific. I'm not sure it would make sense to add this. However, it'll now be searchable in this repo if someone runs into this problem.

PR #1024, opened 2019-10-22
Update expandable rows example

Imballinst:
`expandedRows.data.length` should be `dataIndex`, because the example is referring to the row 5th and above.

Working _correct_ demo: https://codesandbox.io/s/aged-lake-mhiwn.

Signed-off-by: Try Ajitiono <ballinst@gmail.com>


Imballinst:
Actually, I'm closing this because it's already taken care in https://github.com/gregnb/mui-datatables/pull/1025 as well.

Issue #1023, opened 2019-10-22
Hide toggle button when the row isRowExpandable resolves to false


Imballinst:
## Expected Behavior
When `isRowExpandable` is false, the row should have no toggle button.

## Current Behavior
It still have the toggle button -- and is clickable, but it doesn't expand (this is correct behavior).

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/aged-lake-mhiwn

1. Open the link
2. Click on the first row and second row (it works)
3. Click on the third row (it doesn't expand because we set only row 1 and 2 that should expand)
4. But the button is still there!

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.5.1     |
| MUI-datatables  |    2.12.3     |
| React        |     16.8.6    |
| browser      |    Firefox, Chrome     |

gabrielliwerant:
Changing this behavior might break how other people want it to work, so I'm not comfortable with assuming so strongly what the desired UX should be. Instead, what I'd be glad to accept is an additional css class being added somewhere to rows that are not expandable so that devs can use this  to change the experience however they wish.

Imballinst:
Ah, I see! Yeah, that makes more sense and should be safer. So perhaps, I'll leave the default to show, and it can be overridden with, say, `MUIDataTable-expandDisabled`. What do you think?

gabrielliwerant:
That sounds good to me! I assume you mean that you will add that class rather than replace any existing ones. If so, all good.

Imballinst:
Yeah, I'll do that and will ping you once I have re-patched the PR. Thanks for the guidance!

endurance:
Anyone googling who wants to do this code, follow this example.

https://github.com/gregnb/mui-datatables/blob/master/examples/expandable-rows/index.js

The CSS override in the Theme shows you how to hide the toggle button when the expansion is diabled on the row.

z-ax:
Alright, works.
Is there an option to override the row's styles for that case? I don't want cursor pointer on that, it is misleading.

patorjk:
One way would be to use the setRowProps option:

      setRowProps: (row, dataIndex) => {
        if (dataIndex % 2 === 0) { // your check here
          return {
            style: {
              cursor: 'default'
            }
          };
        } else {
          return {};
        }
      },

However, it would be nice if the table did this by default.  I'll put this on my list of things to look into.

PR #1025, opened 2019-10-22
Hide toggle button when the row is not expandable

Imballinst:
Fixes https://github.com/gregnb/mui-datatables/issues/1023. Adds a class `MUIDataTableSelectCell-expandDisabled` to the expand button, when `isRowExpandable(dataIndex)` returns `false`.

This allows us to hide the `IconButton` [shown by default] -- and therefore, won't break existing behavior.

## Expandable toggle button -- does not have class

image

## Non-expandable toggle button - has class `expandDisabled`

image

### GIF of the Example

In this GIF, I will show case 2 examples -- the new example for the hidden button one, and the existing example (just to show the current features are not breaking).

ezgif com-video-to-gif(5)


gabrielliwerant:
I don't want to force this behavior by default, and it's not how the parallel feature of row selection works, see my comment on your issue for another way to move forward here.

Imballinst:
PTAL @gabrielliwerant and let me know if you have further requests! Thanks.

examples/expandable-rows/index.js
```diff
@@ -84,7 +84,7 @@ class Example extends React.Component {
       expandableRowsOnClick: true,
       isRowExpandable: (dataIndex, expandedRows) => {
         // Prevent expand/collapse of any row after the 5th (but allow those already expanded to be collapsed)
-        if (expandedRows.data.length > 4 && expandedRows.data.filter(d => d.dataIndex === dataIndex).length === 0) return false;
+        if (dataIndex > 4 && expandedRows.data.filter(d => d.dataIndex === dataIndex).length === 0) return false;
```
gabrielliwerant: I think we can revert this, as it doesn't operate as intended, to disable any expandable row after the 4th (out of any rows). With your change, it will only disable according to the position of the row, which wasn't the intent.

examples/expandable-rows-hidden-expand/index.js
```diff
@@ -0,0 +1,118 @@
...
+        return (
+          <TableRow>
+            <TableCell colSpan={colSpan}>Custom expandable row option. Data: {JSON.stringify(rowData)}</TableCell>
+          </TableRow>
+        );
+      },
+      onRowsExpand: (curExpanded, allExpanded) => console.log(curExpanded, allExpanded),
+    };
+
+    const theme = createMuiTheme({
```
gabrielliwerant: I think it's overkill to have an entirely new example for this feature. I think we can just take this portion and add it to one of the existing expandable row examples.

src/components/TableBody.js
```diff
@@ -240,6 +240,11 @@ class TableBody extends React.Component {
                     fixedHeader={options.fixedHeader}
                     checked={this.isRowSelected(dataIndex)}
                     expandableOn={options.expandableRows}
+                    // When rows are expandable and selectable, but this row isn't expandable, set this to enabled.
+                    // This will add a new class, MUIDataTableSelectCell-expandDisabled and can be overridden when wanted.
+                    hideExpandButton={
+                      !this.isRowExpandable(dataIndex) && options.expandableRows && options.selectableRows
```
gabrielliwerant: Hm, shouldn't this option be independent of what's happening with `selectableRows`? I would expect an expand disabled class to only matter given expandable options. For example, we could have this situation occur even when selectable rows are turned off entirely in the options.

examples/expandable-rows/index.js
Imballinst: Hmm, I think the comment says otherwise? "Prevent expand/collapse of any row after the 5th". If we want to follow what the comment says, this change is kinda needed.

## Before my change (using expandedRows.data.length, the 5th row is still expandable (it shouldn't)

ezgif com-video-to-gif

## After my change

ezgif com-video-to-gif(1)

examples/expandable-rows-hidden-expand/index.js
Imballinst: Alright, I will defer this to the previous example once our discussion has been resolved!

src/components/TableBody.js
Imballinst: ooops yeah, I think this is my bad. Not sure what went to my head. I will remove the `selectableRows` clause.

examples/expandable-rows/index.js
Imballinst: Oh, and the "After my change" GIF section above was before the merging from 2 files into 1. This is the one after the merge. (see the non-expandable rows don't have toggle anymore)

ezgif com-video-to-gif(2)

gabrielliwerant: Haha, so what you pointed out to me made me realize that my comment was not clear. What I meant was, 5th as in total number of rows expanded. That was my intent. Feel free to change the comment to make that more clear!

Imballinst: ohhhh right, sorry, I thought it means the "order" of the row instead of the "number of rows". All good now!

src/components/TableBody.js
```diff
@@ -240,6 +240,11 @@ class TableBody extends React.Component {
                     fixedHeader={options.fixedHeader}
                     checked={this.isRowSelected(dataIndex)}
                     expandableOn={options.expandableRows}
+                    // When rows are expandable and selectable, but this row isn't expandable, set this to enabled.
```
gabrielliwerant: This comment is probably out of date now that we no longer check for selectable, right?

Imballinst: iuhh correct, I will delete the selectable part.

Imballinst: Updated it! let me know what you think.

Issue #1031, opened 2019-10-24
Option to hide ToolbarSelect


victorhporto:
Hello, I'm making some tests with this component and it's amazing, makes everything that i needed, but i think we could have an simple options, to disable the ToolbarSelect, that shows x row(s) selected. 

Something like:
options.toolbarSelect = {true|false}

I don't find any way to make it desapear, the option `customToolbarSelect` only desaper with the delete button, but i wish we could hide the whole component.

## Expected Behavior

Hide the component ToolbarSelect with an option
## Current Behavior

There's no options to do that
## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  | 4.3.0        |
| MUI-datatables  | 2.12.3        |

gabrielliwerant:
Hello @victorhporto. This has been a highly requested feature with many PRs and issues open for it, so marking this one as duplicate and closing. But feel free to run a search and you find the other issues/PRs if you want to follow them.

PR #1032, opened 2019-10-24
Added option to hide Toolbar Select

victorhporto:
I added a simples option to hide the whole Toolbar Select in reponse to my issue #1031


gabrielliwerant:
See my comments in the issue.

PR #1042, opened 2019-11-01
Add open options to the webpack dev server

wdh2100:
https://webpack.js.org/configuration/dev-server/#devserveropen

Add open options to the webpack dev server

You can check it immediately when you run it(npm run dev)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -16,6 +16,7 @@ module.exports = {
     hot: true,
     inline: true,
     host: "0.0.0.0",
+    open: true,
     port: 5050
   },
   module: {


wdh2100:
What do you think about switching host '0.0.0.0' to 'localhost'

gabrielliwerant:
Hello @wdh2100.

'localhost' is probably better than '0.0.0.0', so I'm ok with that change, however, it might be even better to choose a port, so that devs running locally can better manage multiple concurrent running projects (by managing their port usage accordingly). What do you think?

Regarding `open` by default, I wonder what the community feedback would be. For me personally, I actually prefer not opening by default, as it means I don't collect a bunch of extra tabs if I happen to stop/restart the server, which I do quite a bit when testing for various reasons.

oliviertassinari:
> Regarding open by default, I wonder what the community feedback would be. For me personally, I actually prefer not opening by default, as it means I don't collect a bunch of extra tabs if I happen to stop/restart the server, which I do quite a bit when testing for various reasons.

I would prefer not to open default as well.

Issue #1049, opened 2019-11-04
Pagination row exceeds table bounds on mobile


joshwaiam:
On mobile devices, the pagination row in the footer of the table exceeds the bounds.  The left and right arrows in particular are what actually break out of the right edge of the table.

This appears to be happening because of a 24px padding applied to the last-child in MUIDataTablePagination.  I was able to fix this in my project by applying an override to the element, but it would be great to not need to create an override for this!

```typescript
              <ThemeProvider
                theme={{
                  overrides: {
                    MUIDataTablePagination: {
                      root: {
                        "&:last-child": {
                          padding: 0
                        }
                      }
                    }
                  }
                }}
              >
```

joshwaiam:
I should have said thank you for you and your team's efforts and work on this project.  I have used it in many projects and thank you all for this wonderful tool and your continued commitment to refining and improving it!

gabrielliwerant:
Haha, thanks @joshwaiam! Team of 1 at the moment, but much appreciated.

I would be open to a PR that addresses this at the proper responsive breakpoints. Let me know if you're interested and I can assign you!

joshwaiam:
I would love to try and help with this!  I've been a coder/tinkerer half my life, but plagued by doubt.  I've been wanting to contribute to a project, but I'm been scared because I've never done it!  This seems like as good a time as any to take a swing!

gabrielliwerant:
That's great @joshwaiam, and I agree I think this is a great opportunity to get your feet wet! Don't worry, we all start out that way, and I'm glad to help you through it if need be. Let me know if you'd like me to start you off by pointing you in the right direction, or if you wanted to tinker around on your own first; I'm good with either.

joshwaiam:
I appreciate your willingness to help!  I'll take a crack at it and see what I can do before taking your time on it!

joshwaiam:
Sorry for everyone's time on this.  I thought I did a good amount of testing before reporting this, but it must have been something on my end as two new projects do not replicate the issue.  Thank you for taking the time to respond and guide me, and apologies for any wasted effort on your part!

PR #1051, opened 2019-11-05
padding: 0 for defaultPaginationStyles root

joshwaiam:
Bug fix for issue #1049.

### DIFF
--- a/src/components/TablePagination.js
+++ b/src/components/TablePagination.js
@@ -9,7 +9,7 @@ import { getPageValue } from '../utils';
 const defaultPaginationStyles = {
   root: {
     '&:last-child': {
-      padding: '0px 24px 0px 24px',
+      padding: 0,
     },
   },
   toolbar: {},


gabrielliwerant:
Hey @joshwaiam, thanks so much for getting involved!

I'm trying to test the change, but I'm having trouble finding where the breakage occurs normally. On `master`, this is how the table looks in mobile mode in my dev tools:

image

patorjk:
I also can't reproduce this. I went to https://www.material-ui-datatables.com/ on my phone and then clicked one of the examples in the sidebar (which pops a codesanbox example) and it seems to be displaying correctly.

joshwaiam:
I'm sorry for wasting everyone's time on this.  I tried it in a new project and it does appear to be working fine without this fix.

patorjk:
No worries! I’m a heavy user of this table so I appreciate people looking into things like this.

joshwaiam:
I think I've been able to figure out when this happens, but not sure it's something readily fixable.  When trying to use the MUIDataTable inside of MUI's Container or a Grid, the padding of those parent containers causes the footer row to shift.  Looking through the mui-datatables docs to see if there is something I'm missing to address this, as using the datatable within a layout seems a common use-case!

PR #1053, opened 2019-11-06
Reset `activeIcon` to `null` when Search button is clicked while search is already open

pgoforth:
Currently, clicking the Search button while search is already open will close the search bar, but will not revert the `activeIcon` state to `null`.

This PR will toggle the click method of the Search button based on the `showSearch` state property so that clicking the button while search is open will close the search bar and revert the icon **as long as there is no search query entered**.

When there is a search query, the icon will remain in its active state until the query is cleared.

### Without Search Query
1. Click Search Icon
    * Query area appears
    * Search Icon gains active state
2. Click Search Icon Again
    * Query area closes
    * Search Icon looses active state
### With Search Query
1. Click Search Icon
    * Query area appears
    * Search Icon gains active state
2. Enter search query
3. Click Search Icon Again
    * Query area remains open
    * Search Icon retains active state
4. Click Filter Icon
    * Query area remains open
    * Search Icon retains active state
    * Filter Icon gains active state
5. Click outside Filter Icon, closing popover
    * Query area remains open
    * Search Icon retains active state
    * Filter Icon looses active state


src/components/TableToolbar.js
```diff
@@ -177,7 +177,12 @@ class TableToolbar extends React.Component {
   };
 
   getActiveIcon = (styles, iconName) => {
-    return this.state.iconActive !== iconName ? styles.icon : styles.iconActive;
```
gabrielliwerant: I believe we can simplify all of the logic here, thanks to the other work that you did. We can keep the old return statement, but just add this above it:

```
    if (!this.state.iconActive && iconName === 'search') return styles.icon;
```

What we care about now is just that when the icon is inactive for search, we return the inactive styles. I've tested and it looks like it works. I'd rather not rework any logic we don't have to, so that we reduce the chance for regressions. So the whole function would be

```
    if (!this.state.iconActive && iconName === 'search') return styles.icon;
    return this.state.iconActive !== iconName ? styles.icon : styles.iconActive;
```

and I think we're good.

pgoforth: @gabrielliwerant 
The logic I added to `getActiveIcon` was implemented for the cases where the icon needed to be set to an `active` state when the search field was initialized in some way outside of a click on the icon...be it through `searchText` being set or `searchOpen` being `true`. I have added two additional tests to this PR which will both fail if we only care about the `iconActive` state.

Realistically, we should not be paying attention to `iconActive` state at all due to the ability for search to be active regardless of which icon was last pressed.

Now, if we are using `iconActive` as more of a focus indicator for WCAG, I can see a use for that...but I would think that should be a different CSS class altogether so that a user can tell the difference between the button currently being acted upon, and a button indicating an active table state. Regardless, that would be a whole other, more involved PR.

gabrielliwerant: Good point, thank you for pointing that out and for adding the additional tests!

Issue #977, opened 2019-10-03
Filtered values get reset in case of multiple selection of items.  


manojakuri:
Filtered values get reset when you have multiple selection data-table. On selecting a single all all rows from header, the filter reset and displays all rows 

## Expected Behavior
If you filtered a data table and have a multiple selectable rows, the filter should stay unless user removes the filter or reset filter or changes any filter values. Data- Table shouldn't keep the state of filter and should only show filtered rows on single or multiple row selection

## Current Behavior
On multiple data-table adding a filter and selecting a single or multiple rows on filtered results, control removes filter and shows all rows

| Tech         | Version |
|--------------|---------|
| Material-UI  |     4.4.3    |
| MUI-datatables  |   2.11.1      |
| React        |   16.10.1      |
| browser      |   Chrome/Safari      |
| etc          |         |

gabrielliwerant:
Hey @manojakuri, not sure I completely understand how to reproduce this. Can you add explicit steps? What does selecting "all all rows from header" mean? The select all checkbox? That doesn't seem to have any impact on the filters for me:

![capture](https://user-images.githubusercontent.com/1779564/66152564-b944a300-e5e7-11e9-99bc-d8f5bb0f1c71.gif)

acberk:
I have the same issue using the steps in your screenshot @gabrielliwerant.

Haven't had time to dig in yet, here are the custom table options I'm using, maybe one of the custom function handlers is the cause?

`
  const tableOptions = {
    customSearch: searchUtil,
    customToolbarSelect: () => (
      <Box my={'6px'} marginRight={'24px'}>
        <BulkEditModal
          executeCommand={executeCommand}
          selectedData={selectedData}
        />
      </Box>
    ),
    download: true,
    downloadOptions: {
      filename: 'test.csv',
      separator: ','
    },
    filter: true,
    filterType: 'checkbox',
    fixedHeader: true,
    pagination: true,
    print: false,
    responsive: 'scrollMaxHeight',
    rowsPerPage: 6,
    rowsPerPageOptions: [6, 12, 18],
    rowHover: true,
    search: true,
    selectableRows: 'multiple' as SelectableRows,
    searchText: searchText || '',
    sort: true,
    viewColumns: false,
    onRowsSelect: handleRowSelect,
    rowsSelected,
  };`

Tech | Version
-- | --
Material-UI | 4.0.0
MUI-datatables | 2.11.0
React | 16.8.5
browser | Chrome

[...]

Issue #807, opened 2019-08-02
TableState lost during re-render.


Fabbok1x:
The table state is lost during a re-render. All filters, page-rows, and other related information are reset.

Example to re-produce the issue:
```
import React from "react";
import ReactDOM from "react-dom";
import MUIDataTable from "mui-datatables";
import {
  createMuiTheme,
  MuiThemeProvider,
  withStyles
} from "@material-ui/core/styles";

class App extends React.Component {
  constructor(props) {
    super(props);

    window.table = this;
  }
  getMuiTheme = () =>
    createMuiTheme({
      overrides: {
        MUIDataTable: {
          root: {
            backgroundColor: "#FF000"
          },
          paper: {
            boxShadow: "none"
          }
        },
        MUIDataTableBodyCell: {
          root: {
            backgroundColor: "#FF0000"
          }
        }
      }
    });

  render() {
    const columns = [
      {
        name: "Name",
        options: {
          filter: true
        }
      },
      {
        name: "Title",
        options: {
          filter: true
        }
      },
      {
        name: "Location",
        options: {
          filter: false
        }
      },
      {
        name: "Age",
        options: {
          filter: true
        }
      },
      {
        name: "Salary",
        options: {
          filter: true,
          sort: false
        }
      }
    ];

    const data = [
      ["Gabby George", "Business Analyst", "Minneapolis", 30, 100000],
      ["Aiden Lloyd", "Business Consultant", "Dallas", 55, 200000],
      ["Jaden Collins", "Attorney", "Santa Ana", 27, 500000],
      ["Franky Rees", "Business Analyst", "St. Petersburg", 22, 50000],
      ["Aaren Rose", "Business Consultant", "Toledo", 28, 75000],
      ["Blake Duncan", "Business Management Analyst", "San Diego", 65, 94000],
      ["Frankie Parry", "Agency Legal Counsel", "Jacksonville", 71, 210000],
      ["Lane Wilson", "Commercial Specialist", "Omaha", 19, 65000],
      ["Robin Duncan", "Business Analyst", "Los Angeles", 20, 77000],
      ["Mel Brooks", "Business Consultant", "Oklahoma City", 37, 135000],
      ["Harper White", "Attorney", "Pittsburgh", 52, 420000],
      ["Kris Humphrey", "Agency Legal Counsel", "Laredo", 30, 150000],
      ["Frankie Long", "Industrial Analyst", "Austin", 31, 170000],
[...]

PR #1058, opened 2019-11-08
#977 Maintain filter state in buildColumns updates

acberk:
Addresses Issue https://github.com/gregnb/mui-datatables/issues/977

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -388,9 +388,10 @@ class MUIDataTable extends React.Component {
    */
 
   buildColumns = newColumns => {
-    let columnData = [];
-    let filterData = [];
-    let filterList = [];
+    let { columnData, filterList, filterData } = this.state;
+    columnData = columnData || [];
+    filterData = filterData || [];
+    filterList = filterList || [];
     let sortDirectionSet = false;
 
     newColumns.forEach((column, colIndex) => {
@@ -442,8 +443,8 @@ class MUIDataTable extends React.Component {
 
       columnData.push(columnOptions);
 
-      filterData[colIndex] = [];
-      filterList[colIndex] = [];
+      filterData[colIndex] = filterData[colIndex] || [];
+      filterList[colIndex] = filterList[colIndex] || [];
     });
 
     return { columns: columnData, filterData, filterList };


gabrielliwerant:
Hey @acberk, thanks for looking into this. Just so I'm clear on what's being fixed while I'm testing, is it true that the filters are reset when filter state in the column options is manually updated? Or is there another cause?

acberk:
@gabrielliwerant I'm not sure that's quite it. 

The contrived example showed the filters resetting when _any_ state was manually updated in the column options (in this case, the _onRowsSelect_ function).
https://codesandbox.io/s/muidatatables-example-u6u8b

From testing, that state update triggers _componentDidUpdate_ --> _setTableData_ --> _buildColumns_

The stack trace:

1. https://github.com/gregnb/mui-datatables/blob/ce277b48104de80120d64cecfa9d931282b6f483/src/MUIDataTable.js#L247
2. https://github.com/gregnb/mui-datatables/blob/ce277b48104de80120d64cecfa9d931282b6f483/src/MUIDataTable.js#L482
3. https://github.com/gregnb/mui-datatables/blob/ce277b48104de80120d64cecfa9d931282b6f483/src/MUIDataTable.js#L391

---

The filters are reset by any manual state call within the column options. It appears that the component updates when state is called. Is that intended? My fix passes the filter data through the _buildColumns_ function so that any previous filters remain active on component updates.

patorjk:
@acberk in the way the table currently works, all state is erased when the table re-renders - filters, sorting, search text, selected rows, etc etc (see this issue: https://github.com/gregnb/mui-datatables/issues/807). I've put in another PR that will cause state to persist during re-render if it's not being overwritten by inputted options or column data. You can see it here if you're interested: https://github.com/gregnb/mui-datatables/pull/1086

acberk:
@gabrielliwerant Did you have a chance to look at this? I've seen a lot of talk around derived state, but this smaller PR should take care of cases where selecting a row incorrectly resets the filters.

In other news, I found that I can work around it by self-managing filter state, and passing it down to each column with the filterList option. But this feels like a workaround.

patorjk:
@acberk Using the filterList option is how you're currently supposed to preserve state with filters (see @gabrielliwerant's comment here: https://github.com/gregnb/mui-datatables/issues/807#issuecomment-517817791). An in-depth discussion of the problem can be found in https://github.com/gregnb/mui-datatables/issues/807.

This PR seems to only address the filterList option, and it doesn't allow the user to override the filterList from the option's object once the table has created a filterList (which prevents developers from loading filters externally after the table is loaded). There is another PR https://github.com/gregnb/mui-datatables/pull/1086 that would make all options, including filterList, persist on rerender if not overwritten on the options object.

acberk:
@patorjk Fair enough, that makes sense. I don't think I knew about the filterList option during this fix. Sounds like that'll be the answer in the meantime. Thanks for chiming in.

Issue #444, opened 2019-02-26
CSV and Filter Columns names takes from name column instead of Label column


Prasad-YS:
Technically The CSV column name should takre the labels whatever displayed in the UI, but it is taking the response filed name.

const columns = [
 {
  name: "name",
  label: "Name",
  options: {
   filter: true,
   sort: true,
  }
 },

CSV and Filter columns should have "Name" instead of "name"

gregnb:
Would you like to send in a PR to address this?

PR #1059, opened 2019-11-11
Fix(#444): downloaded csv should show column "label", not "name"

Beamanator:
See #444 

Here's an example to look at as well, with the latest `@material-ui/core`, `mui-datatables`, and `react` packages: https://codesandbox.io/s/muidatatables-custom-toolbar-dzhnx

After downloading the csv file, notice how column labels are all lower-case, as they are in the `'name'` properties of the column options, instead of the `'label'` property:

image

Near the bottom of the `buildColumns` method of the `MUIDataTable` class, you can see that the `label` column option will always populated, with its default being the column `name` (if that column's options is an object) or just that column string:

image


Beamanator:
I don't believe this warrants an extra setting in a column's `downloadOptions`, but I'd be happy to think about that case if y'all prefer.

Beamanator:
@gabrielliwerant I was thinking the following referenced line prevented `column.label` (in [`buildHead`](https://github.com/gregnb/mui-datatables/blob/12b2d2baadd2fe4815ea04fe9f9240a1687876d7/src/utils.js#L40)) from ever *not existing*, since it copies `column.name` into `label` if `column.label` doesn't exist (and it seems like this is where the main column data gets built and used in the rest of the library):

https://github.com/gregnb/mui-datatables/blob/a7db00268e4498fcee93f87b0ac8c1d95a5f4a9f/src/MUIDataTable.js#L468

However, maybe I'm missing another use case where `column.label` can be `undefined`. Either way, It's an easy fix to just use `column.name` if `column.label` doesn't exist - as you mentioned, and as seen in the latest commit. Hope this is what you were thinking!

gabrielliwerant:
@Beamanator You are right that the table sets `label` as `name` when none exists and you are using the table normally. However, when `buildCSV` is used/tested in isolation, it breaks without the default being set. I think it's good practice to keep as many pieces as possible working in isolation so that they could be used on their own if necessary, and not be dependent on side effects which are difficult to test and maintain.

To that end, your new commit is perfect, thank you!

thclark:
Thanks to you both, @Beamanator and @gabrielliwerant for your work here :)

I'm conscious that I'm kool-aid-guy-ing in here, and am sorry for that. But, I think it would be very helpful to have this as an option (I'd suggest it's introduced as default off, then at the next major version release changed to default on), because:
 - This is a breaking change for anyone who uses those downloaded sheets in any kind of automation or audit process
- This breaks the exported columns wherever a `label` contains `options.downloadOptions.separator`
- This potentially breaks data integrity/auditability, since there's no export of the actual data field names

Does it seem like a major change to make it an option?

gabrielliwerant:
Hey @thclark. Kool-aid guy references make everything ok. :)

What is the feature to which you are referring? The issue here is about label vs name showing in the exported csv, which is a bug and not a breaking change. Is there something I'm missing?

thclark:
Hi @gabrielliwerant

So, the first aspect was really asking the question "what do the end users do with the CSV files once they've downloaded them". If they have some kind of workflow based on those exports (like saving to a dropbox folder zaps the contents to a google sheet, for/ex), the workflow gets broken.

The second aspect was let's say we have columns:
```
[{
name: "wood_petrified",
label: "Wood, petrified"
},
{
name: "wood_fossilised",
label: "Wood, fossilised", 
}]
```
The current implementation will output the CSV:
```
Wood, petrified, Wood, fossilised
value1a, value1b
value2b, value2b
```

Of those two scenarios, the second is more serious since it's actually a broken export (as opposed to the first not being a breaking change here, but potentially meaning downstream workflows get broken).

Entirely your call but thought it worth mentioning these edge cases :)

gabrielliwerant:
@thclark I appreciate you keeping a lookout, but I don't think anything gets broken here. All the fields of data in the export are quoted, so there is not confusion between the separator and actual values. For example, using commas in the label as you show, results in:
```
"Wood, petrified", "Wood, fossilised"
"value1a", "value1b "
...
```

Issue #900, opened 2019-09-06
The selection box is misplaced when I scroll the table (only safari)


RomanKwok:
<img width="379" alt="Screenshot 2019-09-07 at 12 06 41 AM" src="https://user-images.githubusercontent.com/53133673/64442983-7d5f1080-d103-11e9-82c9-c402f3a3c54d.png">

gabrielliwerant:
Hello @RomanKwok. Can you create a codesandbox example that demonstrates the issue? I have a PC, so I cannot download the latest safai for testing. There are some options I can use for testing, but the free trials are limited in capabilities, so having an example ready to go that I can load in a browser will be a big help.

RomanKwok:
Hi, @gabrielliwerant. Thanks for your reply.
I created a sample with codesandbox, please check it.
thanks.

https://codesandbox.io/embed/polished-sun-2dj96?fontsize=14
<img width="1328" alt="Screenshot 2019-09-07 at 1 05 02 PM" src="https://user-images.githubusercontent.com/53133673/64470105-26435500-d170-11e9-8e6b-dac28a688722.png">

patorjk:
I can confirm this. It is also present in iOS devices on Chrome (and probably other browsers). If you alter @RomanKwok's example use to use "scrollMaxHeight", you'll see the problem on the iPhone in Chrome. If you create a table with 10+ rows, you'll also see the issue happens for the table headers when you scroll right.

The issue is due to "position: sticky" in TableSelectCell.js and TableHeadCell.js. It's unclear to me if the bug is in Safari and iOS Chrome's implementation of sticky, or the desktop browser's implementation of sticky. Whatever the case, it can be fixed: simply don't stick for the positions you don't want to stick. What this means is removing "top:0px" from the rule for checkboxes that aren't the header, and removing "left:0px" for table header labels. Ex:

TableSelectCell.js

```
    fixedHeader: {
      position: 'sticky',
      top: '0px',
      left: '0px',
      zIndex: 100,
    },
    fixedLeft: {
      position: 'sticky',
      left: '0px',
      zIndex: 100,
    },

...

  const cellClass = classNames({
    [classes.root]: true,
    [classes.fixedHeader]: fixedHeader && isHeaderCell,
    [classes.fixedLeft]: fixedHeader && !isHeaderCell,
    [classes.headerCell]: isHeaderCell,
  });
```

TableHeadCell.js:

```
    fixedHeader: {
      position: 'sticky',
[...]

Issue #983, opened 2019-10-04
CustomFilterListRender not working when ServerSide = true


danielbyun:
CustomFilterListRender does not work when ServerSide = true

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |    2.12.0   |
| React        |     16.9.0    |
| browser      |    chrome    |
| etc          |         |

akirwin:
Seeing this same issue - is there an update on this?

gabrielliwerant:
Server side filters have been overhauled as they didn't quite work before. The way you specify your rendered filter chips now is with `serverSideFilterList`. Not having this work alongside `customFilterListRender` was an oversight. I don't have any timeline, but feel free to jump in with a fix.

patorjk:
I'm not so sure this is a bug, looking at the code, customFilterListRender does look like it gets applied. However, in the example above, there's no serverSideFilterList option provided. If serverSide is set to true, then the filter list is rendered using the serverSideFilterList. Also, in the example above, the filter dialog has no footer for submitting filters.

I've been thinking about this and I believe the library would be improved if this line https://github.com/gregnb/mui-datatables/blob/master/src/components/TableFilterList.js#L102 was changed to this:

        {(serverSide && serverSideFilterList)

That way the table would continue it's previous behavior if the user wasn't using the serverSideFilterList option.

I have a few more ideas regarding server-side filters, I may put in a PR for them.

giorgiofellipe:
I can confirm this bug still happens on v2.14.0

Issue #740, opened 2019-07-03
Expand all rows Expandable Row Header?


patorjk:
I was looking for this option but didn't see it - is there a way to have the header column for the expandable rows have an icon that will Expand/Close all of the rows?

gabrielliwerant:
That option does not yet exist. Marking this as a duplicate as it has come up a number of times. You can follow the threads here for more information: https://github.com/gregnb/mui-datatables/issues/549, https://github.com/gregnb/mui-datatables/issues/518.

patorjk:
It’s similar, though slightly different in that I’m looking to control via the column header. However, marking as a dup is fine, I’ll keep an eye on those other issues.

gabrielliwerant:
You can already make custom column headers, so the only part that is missing is a way to programmaticly interact with the expand functionality. So if you had that, you would be able to do what you wanted.

patorjk:
You can only create custom headers for columns in the Columns array. The expandable row options are set on the main options object and there isn’t a way to set its header (at least from what I can tell).

gabrielliwerant:
It won't matter which object it's in, you could theoretically track any of it in your state and make any changes you wanted from the header, to the header, whatever, as long as you had the ability to trigger expandability at will.

patorjk:
There’s no customHeadRender for the expandable row column though, you can’t change header for that column.

gabrielliwerant:
Now I think that maybe I don't understand your issue. Not sure what you mean by "expandable row column." Expandable rows are rows, so not sure what columns have to do with it. If you want a column header that allows you toggle expandable rows, then that I understand. Maybe I need screenshots or something that demonstrate what you're trying to accomplish.

patorjk:
Sorry if I’m not clear - I’m looking for something akin to the Select All/Unselect All checkbox header that the selectable rows have. So for the expandable rows column (by this I mean the column representing the arrow buttons that are pressed to open a row), I’m looking for an arrow I could push in the column header that would expand all of the rows or collapse all of the rows. 

Part of this would be solved by what you’re talking about, but the other piece would be having the ability to have the column header there, since right now there’s no way to set that column header.

gabrielliwerant:
[...]

PR #1065, opened 2019-11-15
State/Prop fix, New features, Bug Fixes, Plug-ins

patorjk:
This PR is similar to https://github.com/gregnb/mui-datatables/pull/795 in that it's more for discussion, though technically a merge may work but smaller PRs might be better based on discussion.

Months ago I forked this repo. I'm using this table for a rather big project and I needed the flexibility to iterate quickly and add features that I needed. I'm moving that project to my internal codebase so I'll be discontinuing work on the public repo (which was only for my team). However, I added a number of improvements that I wouldn't mind giving back to this project since it's helped me tremendously.  I'll list below the major changes.

* packson.json and package-lock.json have been updated to use the latest packages. All vuneratabilies are gone, and all of the relevant test, rollup, etc, scripts have been updated. This was the first thing I did months ago and I've had no issues. This has also allowed me to use the latest version of react, material-ui, and important features like Hooks. (PR: https://github.com/gregnb/mui-datatables/pull/1077)
* I changed the way State/Props work which I think would eliminate the need for the version 4 refactor. Issue https://github.com/gregnb/mui-datatables/issues/807 is a good example of this. When the table re-renders, all internal state is lost. This is a development nightmare and causes people manage the state themselves. In my version, the table doesn't reset state that isn't overriden in the "options" or "columns" parameters. So if you're only controlling rowsSelected, you won't lose your sort direction when the table re-renders. This seems like the most logical way it should work and follows the KISS principle. Making devs manage all of the state is just too much work and leads to lots of problems. Doing it this way eliminates the need for drastic API changes and re-writes. (PR: https://github.com/gregnb/mui-datatables/pull/1086)
* Using serverside filters does not require a separate API (ie, serverSideFilterList). This was originally why I forked the project (https://github.com/gregnb/mui-datatables/issues/857). My implementation uses the same API for both client and server side filtering, and simply allows for some customization when server side is set to true (through filterPopoverOptions). It is much easier to use, and does not cause the problems that a separate API does (ex: https://github.com/gregnb/mui-datatables/issues/983). 
* A plug-in system has been setup. There are common renderers that users needs (ex: debounceSearchRender), and having the project include these types of plug-ins would make it friendlier to developers. (PR: https://github.com/gregnb/mui-datatables/pull/1089)
* A customBodyRenderLite method was developed that has better performance than customBodyRender. It is useful for tables with lots of data and columns (ie, 50k+ rows, 15+ columns, etc etc).
* Sorting was split into an option called sortOrder. Having the sortDirection on the column object is confusing and causes problems (ex: https://github.com/gregnb/mui-datatables/issues/966). It is also is a huge pain in the butt for people with server side tables. I think serverSide use is a blind spot for this project. The hand full of things I did in this area greatly eliminated pain points. Also, having a sortOrder option also opens the door for multi-column sorting down the road.
* The responsive option was split into "displayMode" and "tableBodyHeight ". This allows one to set the height of the table independent of the display mode (ie: scroll, stacked, responsiveScroll). Having the table height tied to the responsive option is not flexible for all use cases. Also, it's bad for cases where you have a table in the sidebar and a table in the main display area.
* This bug is fixed: https://github.com/gregnb/mui-datatables/issues/900 - this is a bad bug and makes the table unusable in Safari and mobile Chrome if you have checkboxes. (PR: https://github.com/gregnb/mui-datatables/pull/1088)
* An expand all header option was added (https://github.com/gregnb/mui-datatables/issues/740) (PR: https://github.com/gregnb/mui-datatables/pull/1091)
* checkboxColor option was added. Last time I checked this table still had issues with checkbox colors (ex: the filter checkboxes had the wrong highlight color). Also, allowing custom colors is nice.
* readme.md was alphabetized - The current state of things makes finding API methods hard.
* Various function signatures were changed to be consistent.

There may be more, but that's what I can remember off the top of my head.

This PR isn't meant as a slight and is simply an offer to better this project. Before chosing mui-datatables back in June, I did a head-to-head comparsion between it and material-table (which material-ui seems to be pushing and which has moved ahead of this table in popularity) and this table performed much better than material-table in all of my tests. 

Edit: 

Here's the link to the repo: https://github.com/patorjk/mui-dt

Looks like the PR only included the package-lock.json.


patorjk:
Couple of other things just came to mind:

* Handles dark mode for selection and hover colors.
* Performance improvement: doesn't render extra divs when in scroll mode.
* Alphabetized parameters in main file to make easier to track.

gabrielliwerant:
Hey @patorjk, agreed with many of the suggestions and improvements here, and would be happy to have a look at them! I think opening each as a separate PR would be best so we could discuss and review each. Thanks so much for your support.

patorjk:
I've begun putting in PRs, I'll update each bullet point above with the PR when I get it in. These are actually taking more time to do than I thought. Most of the PRs I'm planning are in that mui-dt repo. The two most important ones are the package.json updates and the state persistence update. However, after seeing another PR you have lined up, I really think the displayMode/tableBodyHeight idea should be considered. I'll try and work on a PR for that later this weekend.

patorjk:
Also, I realized what went wrong with this PR. I didn't want to delete my mui-datatables repo since it had my branches in it, so I created a new repo and imported this repo into it for the mui-dt project. When creating this PR I used my mui-datatables fork instead of mui-dt - though it looks like I'm unable to create merge using the mui-dt repo since it's not considered a fork in github's eyes.

Issue #937, opened 2019-09-19
Filter Chips are getting concatenated in filterType 'custom' mode, incompatible with multi-select.


buwanekab:
When column with filterType `custom` mode and with multi-select options. filter chips are getting concatenated in the same chip, when select on multiple options.

## Expected Behavior
filterType `custom` with multi-select options should support for multiple chips to be displayed in the chips list. Like in the current filterType `multiselect` mode.

### FilterType with `custom`
**Filter Options :**
Multiple options are selected.
![image](https://user-images.githubusercontent.com/53601805/65218490-36debe00-dad4-11e9-9060-8d2aa5c0ebe9.png)

**Chip :**
All selected options are concatenated into single chip.
![image](https://user-images.githubusercontent.com/53601805/65218395-01d26b80-dad4-11e9-8f5a-577bc67f6657.png)

### FilterType with `multiselect`
**Chip :**
All selected options are displayed as separate chips.
![image](https://user-images.githubusercontent.com/53601805/65218804-e3b93b00-dad4-11e9-8b6e-173ae24edd8b.png)

## Analysis
I have investigated mui-datatable `TableFilterList.js` component, and i found that, the filterType `custom` doesn't handle multi-values in the `filterList[index]`
https://github.com/gregnb/mui-datatables/blob/f0ea9e58d82839c57a4dfd4a342b6649932660a7/src/components/TableFilterList.js#L44

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |   4.3.3   |
| MUI-datatables  |   2.11.1   |
| React        |   16.9.0   |
| browser      |   any   |

gabrielliwerant:
Hello @buwanekab. So, if you are using `filterType: "custom"`, then you are required to supply your own filter component and logic, and so on. That means that you have to make your own multiselect component. Given that, I'm not sure that `custom` is really what you're looking for. Why doesn't simply using `multiselect` work for you?

gabrielliwerant:
I think I'll need to see a sample of code that isn't working to be of more help. Can you post something to codesandbox?

buwanekab:
@gabrielliwerant ,
can check on this codesandbox example :
https://codesandbox.io/s/monitoring-dashboard-hhb42?fontsize=14
FYI - 
`custom` filterType in  `Alert Source`
`multiselect` filterType in  `Alert Severity`

promerorgz:
Not sure if youve solved this yet but I came across this same issue.

[...]

PR #1067, opened 2019-11-17
Feature/issue 937 array and custom update filter support

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/937.

- Add support for array of values in custom filter list rendering, so as to support custom multiselect workflows
- Add custom update for custom filter list rendering, as this is needed to support the feature
- Create new API for custom filter list options and deprecate the old API


BjoernRave:
is there an example of the usage somewhere?

gabrielliwerant:
Yes, `customize-filter` was updated to use the new API.

PR #1073, opened 2019-11-21
Deprecation warnings

jdetle:
Currently there are some deprecation warnings that show up in production code, not sure what your opinion is @gabrielliwerant, but I think its enough to know about the deprecations when developing locally, no need for error reporters to see it I think.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -376,7 +376,7 @@ class MUIDataTable extends React.Component {
     if (options.expandableRows && options.renderExpandableRow === undefined) {
       throw Error('renderExpandableRow must be provided when using expandableRows option');
     }
-    if (this.props.options.filterList) {
+    if (this.props.options.filterList && process.env.node_env === 'development') {
       console.error(
         'Deprecated: filterList must now be provided under each column option. see https://github.com/gregnb/mui-datatables/tree/master/examples/column-filters example',
       );
@@ -501,7 +501,7 @@ class MUIDataTable extends React.Component {
       transformedData.filter(
         data => data.filter(d => typeof d === 'object' && d !== null && !Array.isArray(d)).length > 0,
       ).length > 0;
-    if (hasInvalidData)
+    if (hasInvalidData && process.env.node_env === 'development')
       console.error(
         'Deprecated: Passing objects in as data is not supported, and will be prevented in a future release. Consider using ids in your data and linking it to external objects if you want to access object data from custom render functions.',
       );
@@ -556,9 +556,11 @@ class MUIDataTable extends React.Component {
       if (column.filterOptions) {
         if (Array.isArray(column.filterOptions)) {
           filterData[colIndex] = cloneDeep(column.filterOptions);
-          console.error(
-            'Deprecated: filterOptions must now be an object. see https://github.com/gregnb/mui-datatables/tree/master/examples/customize-filter example',
-          );
+          if (process.env.node_env === 'development') {
+            console.error(
+              'Deprecated: filterOptions must now be an object. see https://github.com/gregnb/mui-datatables/tree/master/examples/customize-filter example',
+            );
+          }
         } else if (Array.isArray(column.filterOptions.names)) {
           filterData[colIndex] = cloneDeep(column.filterOptions.names);
         }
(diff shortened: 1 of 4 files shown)


gabrielliwerant:
Hey @jdetle, I think that hiding warnings in production environments makes sense. However, I think it might be nice to have a higher level abstraction around this so that's easier to use in future cases, and we're less likely to forget to add it. We can also use that to tweak the logic more easily in the future if we need to as well.

One idea would be to send all warnings to a warning function before logging them, but I'm open to other suggestions. What do you think?

jdetle:
My favorite experiences have been when the deprecation error is a warning that will occur once per session, multiple uses of the table in one session shouldn't show the warning again IMO. We could have 'warn' utility function somewhere that accepts an id for a deprecation warning, sends the warning, and records that in a global object somewhere so that future calls to the function dont do anything if that id has already been recorded. What do you think?

gabrielliwerant:
I don't think we should only show the warnings once per session. They will be too easy to miss for developers that don't know to look for them or to expect once-only behavior. I think hiding in production is good, but the warnings are important, and I do want developers to be bothered by them so that they will fix them rather than ignoring them.

jdetle:
Okay then in that case just special log function imported from a util file somewhere that hides the message if its in prod?

gabrielliwerant:
Yes @jdetle, that sounds good to me!

jdetle:
Okay, made an attempt, let me know if you want to change the naming

test/MUIDataTable.test.js
```diff
@@ -1179,7 +1179,10 @@ describe('<MUIDataTable />', function() {
     instance.selectRowUpdate('cell', { index: 1, dataIndex: 1 });
     shallowWrapper.update();
 
-    const expectedResult = [{ index: 0, dataIndex: 0 }, { index: 1, dataIndex: 1 }];
+    const expectedResult = [
+      { index: 0, dataIndex: 0 },
+      { index: 1, dataIndex: 1 },
+    ];
     const state = shallowWrapper.state();
```
jdetle: Sorry my idea formatted this

jdetle:
@gabrielliwerant, I made the changes, hopefully they meet your expectations :D

src/MUIDataTable.js
```diff
@@ -16,8 +16,8 @@ import TableHead from './components/TableHead';
 import TableResize from './components/TableResize';
 import TableToolbar from './components/TableToolbar';
 import TableToolbarSelect from './components/TableToolbarSelect';
-import getTextLabels from './textLabels';
-import { buildMap, getCollatorComparator, sortCompare, getPageValue } from './utils';
+import textLabels from './textLabels';
```
gabrielliwerant: Is there a reason this line was changed? We need the `getTextLabels` import from what I can tell.

test/MUIDataTable.test.js
gabrielliwerant: If you run `npm run prettier`, it will format everything for you.

src/utils.js
```diff
@@ -5,6 +5,12 @@ function buildMap(rows) {
   }, {});
 }
 
+function warnDeprecated(warning) {
+  if (process.env.NODE_ENV === 'development') {
+    console.warn(`DEPRECATED:  ${warning}`);
```
gabrielliwerant: Not totally sure how I feel about prefixing "DEPRECATED" in front of every message because the wording of some of the messages might make this is a bit awkward. Perhaps something like "Deprecation Notice:" would be better. What do you think?

src/MUIDataTable.js
jdetle: I actually dont remember, I'll go back and revert it.

src/utils.js
jdetle: Yeah thats fine, I'll make the change.

gabrielliwerant:
@jdetle Thanks for making those changes! One more thing. I didn't notice this last time, but since I'm able to test it now, I realized that you changed the errors to warnings. I think they should stay errors. My main priority here is to ensure that devs are adequately notified about breaking changes and I think warnings are much easier to ignore. Additionally, material-ui uses the convention of errors so I think it would be good to stick with how they do it since this library is built to work with it.

jdetle:
Done.

jdetle:
Not used to this workflow, whats in the way of this getting merged now?

gabrielliwerant:
Nothing, I just have to create the release and document everything. Sometimes I like to put in a certain number of changes at once so I don't have the overhead of creating lots of releases, it depends on numerous factors.

Issue #1063, opened 2019-11-14
Scroll Full Height Overflow-Y: scroll not being set on MUIDataTable-responsiveScrollFullHeight-274


ShaggyKris:
Table overflows parent div with `class="MUIDataTable-responsiveScrollFullHeight-274"`. Adding `overflow-y: scroll` to the element style fixes the issue.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

gabrielliwerant:
Hey @ShaggyKris, looks like a good find! Any interest in opening a PR for it?

PR #1075, opened 2019-11-22
Fixed "scrollFullHeight" overflow-y problem.

ShaggyKris:
Fixed "scrollFullHeight" overflow-y problem.
Added example for swapping responsive values for table.
Added backgroundColor: theme.palette.background.paper to root style of TableSelectCell.


gabrielliwerant:
Hey @ShaggyKris, thanks for contributing! Can you point me to an issue that describes problem being fixed so that I know what to look for when testing?

ShaggyKris:
> Hey @ShaggyKris, thanks for contributing! Can you point me to an issue that describes problem being fixed so that I know what to look for when testing?

https://github.com/gregnb/mui-datatables/issues/1063

examples/examples.js
```diff
@@ -35,36 +36,37 @@ import Themes from './themes';
  * Here you can add any extra examples with the Card label as the key, and the component to render as the value
  */
 export default {
-  'Array Value Columns': ArrayValueColumns,
```
gabrielliwerant: Let's undo these diffs. It's good to have indent in an object, and I don't want this sort of thing to go back and forth in various PRs, so the way it was before was fine.

patorjk:
Another idea: break "responsive" into two options:

* displayMode - stacked, responsiveStacked, and scroll
* tableBodyHeight - "auto" or a set pixel value (or any valid CSS value)

Doing it this way is much more flexible and will allow the scrollbar to be placed on the table body instead of on the surrounding div element. This will allow the footer to be visible too. Adding a scrollbar to the surrounding div doesn't look right.

Additionally, breaking this option into two options makes more sense because there's no reason to tie the table's height to the "responsive" parameter. In the current implementation it's hard to think about. Additionally, in the current implementation you can't have a "stacked" table if you have a large screen and there are cases where you'd want this (ex: you have a sidebar and a table in the main work area - you might want one of these in "stacked" mode, but they'll appear in "scroll" mode if window is too big).

I listed this idea in https://github.com/gregnb/mui-datatables/pull/1065, but I know I listed a ton of stuff there. I used it on mui-dt and it was really helpful though.

gabrielliwerant:
@patorjk That's an interesting idea, definitely worth considering. I'm leaning toward expanding the API we have for now, to keep things simpler to implement in the short term, but I will think about this for the major version bump, especially if there's additional feedback about the current API being confusing.

patorjk:
@gabrielliwerant Sounds good. The reason I broke it out into those 2 parameters in my setup was because I needed my tables to extend to the bottom of browser window and no futher, and to resize when the window was resized. I'm not able to do that in the current setup.

tableBodyHeight isn't the greatest parameter, and there may be a better solution for this part. It still required me knowing the height of the toolbar and footer when I resized (my tableBodyHeight was set to something like "calc(100% - 120px)"), and it required that I know if filters were present (since they affect the height of the table). So it's probably not the optimal solution, but it allows for things that aren't currently possible.

Oh, and one other thing I needed - I wanted to give users the ability to toggle between "stacked" and "scroll" mode, since sometimes looking at data in stacked mode is nicer, and sometimes scroll mode it's nicer. This isn't currently possible as the "stacked" mode is tied to the size of the browser window.

examples/examples.js
maxkarkowski: > As far as the actual fix for the scroll behavior, this turns out to be a little more complicated than we thought. Implementing the auto overflow behavior removes the ability to apply a fixed header (for y axis). So that's a bit of a problem, as it would be a breaking change. I experimented around a bit, and I found that if we apply the style
> 
> ```
> position: aboslute;
> ```
> 
> to the `MUIDataTable-paper` class at the top level `div` for the whole library component, it fixes the overflowing behavior problem, and allows fixed header options. However, it's also a different type of display from `scrollFullHeight`. So what I think we should do is keep `scrollFullHeight` as is, as this will still be a viable option for tables that don't have many columns, and add a new option (also to maintain backwards compatibility) called `scrollFullHeightFullWidth` which implements the `position: absolute` as the fix.

position:absolute seems to collide when you have the Material-UI Grid around the table.
respsonsive setting set to 'scrollFullHeightFullWidthÄ
````
<Grid container justify="center">
        <Grid item xs={12} md={10}>
          <MUIDataTable title={<Title />} data={rawData} columns={TableColumns} options={TableSettings}></MUIDataTable>
        </Grid>
      </Grid>```
screenshot with the position: absolute
<img width="988" alt="20-02-06-congstar VPP Order List-bq1fo" src="https://user-images.githubusercontent.com/622841/73924825-5e8cc200-48cd-11ea-983b-02a316b75de3.png">

<img width="1029" alt="table-with-pos-abs" src="https://user-images.githubusercontent.com/622841/73924896-72d0bf00-48cd-11ea-99f7-581dd112130f.png">



screenshot without ```.MUIDataTable-paperResponsiveScrollFullHeightFullWidth-106 {
    position: absolute;
}```
<img width="988" alt="table-without-pos-abs" src="https://user-images.githubusercontent.com/622841/73924905-75331900-48cd-11ea-907d-ffda3edb87df.png">

Issue #1057, opened 2019-11-08
Sort Column Gets Reset (Server-Side Data)


Ahsan-R-Kazmi:
The sort column direction is getting reset after new data is retrieved from the server and set in the table. It also gets reset on page changes, rows per page changes, and search text changes. Is this intended behavior, and is there any way to override it? 

## Expected Behavior
The sort column should not change when new data is retrieved and set from the server.

## Current Behavior
The sort column direction is getting reset after new data is retrieved from the server and set in the table. It also gets reset on page changes, rows per page changes, and search text changes.

gabrielliwerant:
This is expected behavior with sever side, at the moment. The table does not currently have any way to distinguish which options individually you are managing server side vs client, so server side opts you in to all of them, more or less. This means you have to take the reigns and make sure the various options have the values you want when you update your data.

Ahsan-R-Kazmi:
Could you point my to the right direction on how to set that for the table state? I tried looking through the options, but can't find which option I need to set to choose the sort direction of the column. 
I also tried setting the columns that are provided to the MUIDataTable component through set state, but it doesn't seem to maintain it in the tableState the next time `onTableChange` is called.

```
case 
"sort" : {
   let sortColumnIndex = tableState.activeColumn;
   let columnInfo = tableState.columns[sortColumnIndex];
   let orderByParam = columnInfo["name"];
   let orderByDirection = columnInfo["sortDirection"];
   this.setState({columns: tableState.columns});
   console.log("Table sorted");
   this.getData(this.state.page, this.state.rowsPerPage, this.state.searchText, orderByParam, 
   orderByDirection);
   break;
 }
```

gabrielliwerant:
Should be this one under column options:

![image](https://user-images.githubusercontent.com/1779564/68552071-1dc20180-03e1-11ea-9d5f-06f5e95b3a90.png)

T-pirithiviraj:
@Ahsan-R-Kazmi , You can refer to this. You might get some idea. 
```
https://github.com/gregnb/mui-datatables/issues/244#issuecomment-533260448
https://github.com/gregnb/mui-datatables/issues/947
https://github.com/gregnb/mui-datatables/issues/954
```

Ahsan-R-Kazmi:
So it seems, it works properly if I do something like:
``````
this.state.columns.forEach((column, index) => {
            if (column.name === orderByParam) {
                column.options.sort = true;
                column.options.sortDirection = orderByDirection;
[...]

Issue #947, opened 2019-09-20
Sorting Icon is not coming and sorting direction is always descending.


T-pirithiviraj:
When a user clicks sorting on specific column, It is always showing descending because of re-render of dataTable.

## Expected Behavior
The sorting Icon should display and it shouldn't set always as descending.

## Current Behavior
After API call is made the dataTable fetch data and the direction of sorting is always descending.

## Steps to Reproduce (for bugs)

1. In render Make server-side = true.
2. call onColumnSortChange().
3. Make API call to fetch sorted column data there.
4. After receiving data populate in dataTable and click on column to sort, it re-renders and erases sortDirection as well as icon

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  4.2.1  |
| MUI-datatables  |  2.10.2       |
| React        |   16.8.6      |
| browser      |   chrome      |
| etc          |         |

rhaicode:
I have encountered the same problem too before. But here's my workaround:

1.) You need to have state management, either using Redux or useContext with useReducer.

2.) On MUIDataTable's component, provide you own columns (which is an array of column objects, exported from a certain file (e.g. columns.js) ) for the columns prop.

3.) On each column object, provide your custom column header component on  `customHeadRender` value of each columns' `options` property

4.) On your custom column header component, import TableSortLabel from material-ui and define a div with an `onClick` event that:
   a.) calls the dispatch method responsible for updating your state that holds the records / data
   b.) updates the direction of TableSortLabel to either "asc" or "desc", set "asc" by default

If you're having difficulty getting it, I'll follow up this comment with a CodeSandbox example.

gabrielliwerant:
See my comment here https://github.com/gregnb/mui-datatables/issues/244#issuecomment-533260448. You don't have to provide your own `customHeadRender`, but if you are managing things yourself, serverside, then you need to keep the table up to date with all of the state you wish to be retained, including the sort direction. Otherwise the table doesn't know what your intention was, as it will read its state from the props passed in, and anything not passed in will be empty.

T-pirithiviraj:
Thanks @gabrielliwerant . Can you send the codeSandBox example which will be more helpful to proceed further?

pedroo21:
It would be really useful if it was the datatable managing this internal state for us. In a server-side context, the goal is almost always to preserve the state across remote refreshes.

[...]

Issue #679, opened 2019-06-07
v4 State/Props Refactor Roadmap


gabrielliwerant:
### Intro:

This issue is to track progress on and explain the major refactor that is underway to solve numerous issues relating to state/prop updates as well as to make the codebase simpler, easier to reason about, and better able to support additional features moving forward.

### Overview:

The table uses state, derived from props to manage the data passed into the table. This means that whenever the props sent into the table are changed, they cannot be properly displayed in the table unless the state is re-initialized. This is making it quite difficult for users of the library to modify the props sent to the table, and have those changes reflected, which users should be able to do easily. In addition, the code is becoming difficult to maintain, reason about, and test. See this blog entry https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html or the discussion here for more detail on this issue: https://github.com/gregnb/mui-datatables/pull/650#issuecomment-499262879.

### Solution:

The table should operate mainly from the props sent in and only maintain minimal state that is entirely controlled within the table components. Operating off of props will allow users to make any changes they wish to the their data, anywhere, at any time, and have those changes automatically reflected in the table data.

We can use `render` to capture changes to props and then transform as necessary, as well as filter for searches, deleted rows, filters, etc. without relying on a separate, internal displayed data state. In essence, we will derive the display state directly from the props in the `render` function, without storing it in state.

Since we will be performing transformations and filtering on each render, there are opportunities for performance enhancements here via memoization, but the initial refactor will focus mainly on feature parity. Some functions may be become obsolete and will be deprecated as a result of the refactor (e.g. the difficult to understand `updateValue` function may no longer be needed, as it exists mainly to keep state in sync with changes).

### Process:

As soon as a working example is ready, it will be released as `3.0.0-alpha` and tracked separately from `2.x.y` releases. Features or fixes in the v2 releases will need to be ported over, but it will focus mainly on important bug fixes to minimize churn (though some features that are high in demand can still be added during the transition).

Hopefully the community can then help to test and vet the v3 alpha to make it ready as the replacement for v2.

### Roadmap:

- [ ] Separate out all data modeling from state-specific functions (ideally in separate functions and/or files)
- [ ] Replace derived state management with direct transformations and filters on props
- [ ] Update API to include onChange functions to match prop-based data (e.g. if a sort option is set as an option on props, an `onChange` function will be required to be supplied to the table to update that sort, since it cannot be managed by internal state at that point).
[...]

PR #1086, opened 2019-11-29
Updated table to make state persistent between renders

patorjk:
This PR updates the table so that it maintains it's state between re-renders for properties that are not overwritten by the user (https://github.com/gregnb/mui-datatables/issues/807, https://github.com/gregnb/mui-datatables/issues/947, https://github.com/gregnb/mui-datatables/issues/1057). I've updated the "Simple" and "Data as Objects" examples with a "re-render" button that shows that search text, filters, selections, sorting, etc etc, are not lost when the table re-renders.

Some notes:

* Selected Rows and Expanded Rows reset if the data changes. This is why I've put the data on the component's state for the 2 examples I updated. In my own use, I found that if the table's data changed, it made no sense for the selections and expansions to stay in place.
* The Trash Can in the toolbar doesn't effect the external state data in those two examples, so any deleted data is returned to the table on re-render. This makes sense because the user is inputing non-modified data to the component, so the table should use the non-modified data rather than internal data that is modified.
* I'm not sure of a good way to test this. I've copied over my edits from the mui-dt repo I mentioned in the other thread, and I'd been using the changes over there in my app without issue.

### DIFF
--- a/examples/data-as-objects/index.js
+++ b/examples/data-as-objects/index.js
@@ -4,6 +4,22 @@ import MUIDataTable from "../../src/";
 
 class Example extends React.Component {
 
+  state = {
+    counter: 0,
+    data: [
+      { name: "Gabby George", title: "Business Analyst", location: "Minneapolis", age: 30, salary: "$100,000", phone: { home: '867-5309', cell: '123-4567' } },
+      { name: "Aiden Lloyd", title: "Business Consultant", location: "Dallas",  age: 55, salary: "$200,000", phone: { home: '867-5310', cell: '123-4568' } },
+      { name: "Jaden Collins", title: "Attorney", location: "Santa Ana", age: 27, salary: "$500,000", phone: { home: '867-5311', cell: '123-4569' } },
+      { name: "Franky Rees", title: "Business Analyst", location: "St. Petersburg", age: 22, salary: "$50,000", phone: { home: '867-5312', cell: '123-4569' } }
+    ]
+  }
+
+  rerender = () => {
+    this.setState((prevState, props) => ({
+      counter: prevState.counter + 1
+    })); 
+  }
+
   render() {
 
     const columns = [
@@ -14,7 +30,7 @@ class Example extends React.Component {
           filter: true,
           display: 'excluded',
         }
-      },      
+      },      
       {
         name: "title",
         label: "Modified Title Label",
@@ -22,7 +38,7 @@ class Example extends React.Component {
           filter: true,
         }
       },
-      {        
+      {        
         name: "location",
         label: "Location",
         options: {
@@ -58,14 +74,6 @@ class Example extends React.Component {
       }
     ];
 
-
-    const data = [
-      { name: "Gabby George", title: "Business Analyst", location: "Minneapolis", age: 30, salary: "$100,000", phone: { home: '867-5309', cell: '123-4567' } },
-      { name: "Aiden Lloyd", title: "Business Consultant", location: "Dallas",  age: 55, salary: "$200,000", phone: { home: '867-5310', cell: '123-4568' } },
-      { name: "Jaden Collins", title: "Attorney", location: "Santa Ana", age: 27, salary: "$500,000", phone: { home: '867-5311', cell: '123-4569' } },
-      { name: "Franky Rees", title: "Business Analyst", location: "St. Petersburg", age: 22, salary: "$50,000", phone: { home: '867-5312', cell: '123-4569' } }
-    ];
-
     const options = {
       filter: true,
       filterType: 'dropdown',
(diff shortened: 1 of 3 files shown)


pedroo21:
Hey @patorjk ,
Correct me if mistaken, but isn't this more like a quick-fix to the table state lost across re-renderings? I would assume the best solution would be the one you suggested in #679 (https://github.com/gregnb/mui-datatables/issues/679#issuecomment-518751789).
In other words, breaking the table state into smaller components, where we would pass both the state and a method to update it.

I see this working just fine, except if you need to control one of the tableOptions and have no way of notifying the parent about changes, rather than the burden of implementing the onTableChange.

patorjk:
> Correct me if mistaken, but isn't this more like a quick-fix to the table state lost across re-renderings?

No, this would be a permanent fix. Basically it makes it so state the table is managing isn't erased.

> I would assume the best solution would be the one you suggested in #679 (#679 (comment)).

That was my opinion a few months ago, but the more I've used the table, the more I'm not so sure. That would be a rather big API change and there are no clear benefits as a user - in fact, the API would become more complex and possibly more confusing. And I've never seen an API that looks like what I suggested. Re-reading [1], it doesn't say derived state shouldn't be used, it just says it's an advanced feature and that in many cases you don't need it.

I noticed at the bottom of the article it invites questions, so I've gone ahead and put in a [question](https://github.com/reactjs/reactjs.org/issues/2603) regarding this particular situation. Maybe someone there will have some insight on the best way to approach this time of problem.

However, from own experience, using derived state this way - as long as you know what you're doing - seems to be fine. It also keeps the complexity under the hood so that users don't have to deal with it. That's partly why people are having so many problems, when they have to manage the state themselves, it becomes a headache.

> have no way of notifying the parent about changes, rather than the burden of implementing the onTableChange.

I don't see this as an issue. Event handlers are traditionally used for this type of thing. 

[1] https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

pedroo21:
I think I was deviating a little bit from the main subject of this PR. I agree with the PR fix to have the table manage its own internal state when the props are not set/undefined. I'm also doing it myself. However, my current concern is regarding the table state that is set and controlled.

> I don't see this as an issue. Event handlers are traditionally used for this type of thing.

If I wanted to control the sortDirection for a certain column, I would have to go into doing something like this:

```
const columns = {
    name: 'foo',
    label: 'bar',
    options: {
        sortDirection: this.state.firstColumnSortDirection
    }
};

const options = {
    onTableChange: (action, tableState) => {
        if (action === 'sort') {
            if (tableState.columns[0].sortDirection !== this.state.firstColumnSortDirection) {
                this.setState({ firstColumnSortDirection: tableState.columns[4].sortDirection });
               // Note: Also, when the sortDirection is none, we're suppose to delete it, not even send it as props.
            }
         }
     }
};
```

While a simpler approach could be:

```
const columns = {
    name: 'foo',
    label: 'bar',
    options: {
        sortDirection: this.state.firstColumnSortDirection,
        onChangeSortDirection: (dir) => this.setState({ firstSortDirection: dir })
       // Or something else that includes this callback as props
    }
};
```

Plus, using the onTableChange is causing an extra re-render on the parent, considering the child has already updated its own internal state.
However, I believe this problem is already sufficiently exposed in #679 and should be a separate PR.

patorjk:
That's a good point and I forgot about that - yes, this fix solves the state loss problem for uncontrolled props but doesn't address that controlled state still has the "Double Render" problem (I'll call it that to make it easy to refer to). The Double Render problem would only be solved by an internal rewrite and API change.

In that sense, this PR would be a stop gap until https://github.com/gregnb/mui-datatables/issues/679 is completed.

Legogris:
Can confirm that this fixes our issues with filters dropping between mutations of table data (for context, using a non-rendered meta column to keep track of client-side editing state per row).

Flavyoo:
Can anyone review and approve this, or do we need someone specific to get this merged?

aamin21:
@patorjk You probably have seen this, but there is a conflict with the PR. 
@gabrielliwerant Can anyone do code review and approve or are there particular maintainers that need to approve for this to be merged?

Thanks for you work guys!

patorjk:
@aamin21 I'm using a fork of this repo that's on my client's servers, it's in current development but it's unavailable to the public. I'd prefer to use this repo, but none of the PRs I submitted months ago were reviewed. I'm not going to submit any more changes if they're not going to get looked at (though I've set this PR up so that maintainers can do any changes they want to it). This update is really important though, even if @gabrielliwerant has other plans for the future, in the current state, the loss of state on re-render is unexpected and leads to bugs. 

Also, PR https://github.com/gregnb/mui-datatables/pull/1088 is really important too, as this table is broken in Safari (see https://github.com/gregnb/mui-datatables/issues/900 for details) and on Chrome/Firefox on iPhone.

I talked to @gregnb on twitter a couple months ago and he expressed interest in coming back, but I understand why he stepped back (he had personal stuff to deal with). I think maybe if he came back, he could help excellerate the development that @gabrielliwerant is working on, but other than that, fixing this issue is in the hands of @gabrielliwerant.

hybmg57:
Can we merge this one in? This is causing problems for me also with filters getting reset

AntonioLobo:
LGTM

hacknaked:
I still see this issue in v3.5.0

uniqueuser-repo:
Why hasn't this been merged yet?

PR #1088, opened 2019-11-29
Fix 900 and refactored fixedHeaderOptions into fixedHeader and fixedSelectColumn

patorjk:
This PR fixes https://github.com/gregnb/mui-datatables/issues/900

It also refactors the fixedHeaderOptions into two parameters:

* fixedHeader (boolean)
* fixedSelectColumn (boolean)

Both of these features work in Safari on Mac as well as in Chrome on the iPhone.

It also addresses the problem of text scrolling "under" a select box when the user scrolls the table horizontally. The fixedHeader example was updated to demonstrate the updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,8 @@ The component accepts the following props:
 |**`rowsPerPage`**|number|10|Number of rows allowed per page
 |**`rowsPerPageOptions`**|array|[10,15,100]|Options to provide in pagination for number of rows a user can select
 |**`rowHover`**|boolean|true|Enable/disable hover style over rows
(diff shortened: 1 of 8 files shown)


patorjk:
I noticed that the Material UI API refers to these types of parameters as "sticky" rather than "fixed". If this PR goes in, it may make sense to change:

* fixedHeader -> stickyHeader (to be more inline with Material UI API: https://material-ui.com/api/table/)
* fixedSelectColumn -> stickySelectColumn

patorjk:
Closing in favor of updates in v3 branch.

PR #1091, opened 2019-11-30
Added expandable row header (expandableRowsHeader option)

patorjk:
Resolves https://github.com/gregnb/mui-datatables/issues/740

This PR adds the option expandableRowsHeader which defaults to true. Users will now be able to toggle all rows open or closed by clicking a button in the header. I've updated the Expandable Rows example to add the option in. You can toggle it to false to see the behavior of the table when the header isn't present.


patorjk:
Closing in favor of updates in v3 branch.

PR #1093, opened 2019-12-02
Added confirmFilters option

patorjk:
I have a feeling this PR may take some convincing, but it has a lot of subtle benefits. Here's what it does: It adds an option called "confirmFilters" that defaults to false and works in conjunction with the customFilterDialogFooter option. If it's set to true, then filters from the Filter dialog won't be applied to the master filterList object unless "confirmed". The customFilterDialogFooter is updated to receive an "applyFilters" function that will apply the filters that are set in the filter dialog.

There are 2 main use cases:

* Server-side filtering - I'll go through how this is different from serverSideFilterList below.
* Filtering on tables with large data sets. 

I would have set this option up to work with the large-data-set example I created in another PR, but it's not available yet so I created another serverside-filtering example to show it in action. You may look at it and see that it's not too different from the existing one, but here are the benefits for a confirmFilters option over a serverSideFilterList: 

* It dovetails nicely with https://github.com/gregnb/mui-datatables/pull/1086. If these two PRs are accepted, then no management of filterLists will be needed for server-side filters.
* It allows for the "Reset" button to behave differently for filter confirmation. Instead of resetting to a blank state, a user can reset back to the initial state of the dialog (ie, the current state the filters are in). 
* It's simpler than serverSideFilterList and requires no management of a filterList.
* It's future-proof since it works with the existing filterList.
* It's backwards compatible and works along side serverSideFilterList.

I'm not advocating that serverSideFilterList be removed or deprecated (that would probably annoy people). I'm pointing these out since there's some overlap for these features for the first use-case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 MUI-Datatables is a data tables component built on [Material-UI](https://www.material-ui.com).  It comes with features like filtering, resizable + view/hide columns, search, export to CSV download, printing, selectable rows, expandable rows, pagination, and sorting. On top of the ability to customize styling on most views, there are three responsive modes "stacked", "scrollMaxHeight", and "scrollFullHeight" for mobile/tablet devices.
 
 <div align="center">
-	<img src="https://user-images.githubusercontent.com/19170080/38026128-eac9d506-3258-11e8-92a7-b0d06e5faa82.gif" />
+  <img src="https://user-images.githubusercontent.com/19170080/38026128-eac9d506-3258-11e8-92a7-b0d06e5faa82.gif" />
 </div>
 
 ## Install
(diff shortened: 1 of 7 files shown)


patorjk:
One other thing this PR does: To avoid confusion found in issues like https://github.com/gregnb/mui-datatables/issues/1105 and https://github.com/gregnb/mui-datatables/issues/1102, this PR makes it so serverSideFilterList only overrides the internal filterList if it's set (see: https://github.com/gregnb/mui-datatables/pull/1093/files#diff-5c2a6a9d873eac896a7241fed4bf9778R102). That way the table only uses the special serverSideFilterList if it's actually being provided.

patorjk:
Closing in favor of https://github.com/gregnb/mui-datatables/pull/1317

PR #1119, opened 2019-12-13
Fix for multiselect customFilterListOptions render

Legogris:
This fixes a bug where `customFilterOptions.render` gets called twice for non-custom filters that override render. For `multiselect`, this beaks, as one of them passes all optoins as array and the second calls it once per option.

Example that breaks with current `master` (just added custom filter render to customize-filer): https://gist.github.com/Legogris/e63efeb24b766a76a26221c03df0c404

### DIFF
--- a/src/components/TableFilterList.js
+++ b/src/components/TableFilterList.js
@@ -122,21 +122,22 @@ class TableFilterList extends React.Component {
               return item.map((data, colIndex) => filterChip(index, data, colIndex));
             })
           : filterList.map((item, index) => {
-              const customFilterListRenderersValue = filterListRenderers[index](item);
-
-              if (columnNames[index].filterType === 'custom' && customFilterListRenderersValue) {
-                if (Array.isArray(customFilterListRenderersValue)) {
-                  return customFilterListRenderersValue.map((customFilterItem, customFilterItemIndex) =>
-                    customFilterChipMultiValue(
-                      customFilterItem,
-                      index,
-                      customFilterItemIndex,
-                      item,
-                      customFilterListRenderersValue,
-                    ),
-                  );
-                } else {
-                  return customFilterChipSingleValue(index, item);
+              if (columnNames[index].filterType === 'custom') {
+                const customFilterListRenderersValue = filterListRenderers[index](item);
+                if (customFilterListRenderersValue) {
+                  if (Array.isArray(customFilterListRenderersValue)) {
+                    return customFilterListRenderersValue.map((customFilterItem, customFilterItemIndex) =>
+                      customFilterChipMultiValue(
+                        customFilterItem,
+                        index,
+                        customFilterItemIndex,
+                        item,
+                        customFilterListRenderersValue,
+                      ),
+                    );
+                  } else {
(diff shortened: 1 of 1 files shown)


Legogris:
When writing more tests this seems to still be broken unfortunately.

Legogris:
Not sure what intended behavior for multichips are. If an array is passed in, the same function is still run first on the array itself and then on each individual item. 
https://github.com/gregnb/mui-datatables/blob/12b2d2baadd2fe4815ea04fe9f9240a1687876d7/src/components/TableFilterList.js#L104
https://github.com/gregnb/mui-datatables/blob/12b2d2baadd2fe4815ea04fe9f9240a1687876d7/src/components/TableFilterList.js#L53

gabrielliwerant:
Hey @Legogris, can you point me to the exact diff that I need to add to the `customize-filer` example to break it? Your gist appears to contain the entire thing, but I'd like to see just the new part. Thanks.

Legogris:
Line 16:

```
customFilterListOptions: {
            render: v => { console.log(v); return v.toUpperCase();  }
          },
```

gabrielliwerant:
@Legogris The behavior for custom filters with arrays was written specifically to handle use cases where people are building their own custom multiselects, which entails passing in arrays of filter values as filter values. So when we loop through and find an array, we need to apply the render logic again to actually render the label correctly. Your fix does seem to prevent running this unnecessarily when there is no custom filter supplied, though custom filters will still need to run this twice. So your fix seems good for cases where there is no custom filter being supplied. Is there some other way it is broken?

Legogris:
> @Legogris The behavior for custom filters with arrays was written specifically to handle use cases where people are building their own custom multiselects, which entails passing in arrays of filter values as filter values. So when we loop through and find an array, we need to apply the render logic again to actually render the label correctly. Your fix does seem to prevent running this unnecessarily when there is no custom filter supplied, though custom filters will still need to run this twice. So your fix seems good for cases where there is no custom filter being supplied. Is there some other way it is broken?

The problem, both before and after my fix, is that there are cases when the same callback (`customFilterListOptions`) gets called in the same path with the first argument being the whole array in one call and individual elements for the rest. My example above should illustrate how this breaks (only workaround would be to add checks inside the callback to check if the input is an array and have separate logic for separate cases). Is this really as intended?

I've read through both the docs and the source multiple times and I am still not sure what the intended behavior is here? It's not just that it's unnecessarily run multiple times, it's that the calls differ in input.

gabrielliwerant:
@Legogris There do appear to be some issues here. I will take a closer look.

gabrielliwerant:
@Legogris I've opened a PR here https://github.com/gregnb/mui-datatables/pull/1140 to fix the issue with the double renderings, and to simplify the logic. Custom multiselect filter lists require handling of array data instead of just strings, but the logic before was separating them in a confusing manner, and resulted in bugs. There were also some prior bugs I didn't catch until I looked through this logic again.

Part of the fix included your changes, so thank you very much for first catching the issue and then attempting a fix. Feel free to take a look at the PR and let me know what you think, and if it fixes your issue.

PR #1120, opened 2019-12-13
Feature/1111 inline filters

### LINKED ISSUE #1111: FR: Add option to show filters inline in the table
My users requested that it would be great to have filters inlined in table so they are easier to find.

I'm thinking that it could be done as in Excel, with an icon next to header name opening a popup:
Screenshot 2019-12-09 at 16 50 06

or as another row just below header (either a real HTML row, or just visually below) with a full filter input as it is now shown in filter popups. 

I'm wondering what do you think about it? I could commit some time to work on a PR. I believe this should be optional, to be set by table option `filterListInline: bool` or similar.

KrzysztofMadejski:
Resolves #1111

What I did:
1. copied `TableFilter.js` as `TableFilterInline.js`
2. modified its render method to show `TableHeadRow` instead of a grid
3. modified rendering of filter inputs (removed labels, made font smaller, etc.)
2. added `filtersInline` configuration option
3. Added option to README and added example
4. tested manually
1. Added unit tests
    - cloned `TableFilter.test.js` to `TableFilterInline.test.js` and adapted it (in separate commit) 
    - added test cases for `TableHead.test.js` and `TableToolbar.test.js`


src/components/TableToolbar.js
```diff
@@ -312,7 +312,7 @@ class TableToolbar extends React.Component {
               }
             />
           )}
-          {options.filter && (
+          {options.filter && options.filtersInline !== true && (
```
gabrielliwerant: I actually would like to see us keep the filters on, even if using inline filters. I can think of use cases for this, but if someone wants to hide the toolbar filter, they already have the filter option here to do so.

gabrielliwerant: I actually would like to see us keep the filters on, even if using inline filters. If someone wants to hide the toolbar filter, they already have the filter option here to do so.

test/MUIDataTableFilter.test.js
```diff
@@ -16,10 +16,10 @@ describe('<TableFilter />', function() {
 
   beforeEach(() => {
     columns = [
-      { name: 'firstName', label: 'First Name', display: true, sort: true, filter: true, sortDirection: 'desc' },
-      { name: 'company', label: 'Company', display: true, sort: true, filter: true, sortDirection: 'desc' },
-      { name: 'city', label: 'City Label', display: true, sort: true, filter: true, sortDirection: 'desc' },
-      { name: 'state', label: 'State', display: true, sort: true, filter: true, sortDirection: 'desc' },
+      { name: 'firstName', label: 'First Name', display: 'true', sort: true, filter: true, sortDirection: 'desc' },
```
gabrielliwerant: Nice catch!

test/MUIDataTableFilterInline.test.js
```diff
@@ -0,0 +1,330 @@
+import Checkbox from '@material-ui/core/Checkbox';
```
gabrielliwerant: I think these tests may be redundant. I'll explain more in the review overview.

test/MUIDataTableToolbar.test.js
```diff
@@ -134,6 +134,17 @@ describe('<TableToolbar />', function() {
     assert.strictEqual(actualResult.length, 0);
   });
 
+  it('should render a toolbar with no filter icon if option.filtersInline = true', () => {
```
gabrielliwerant: Per above comment about regular filters, I think we should remove this.

src/components/TableHead.js
```diff
@@ -30,7 +31,9 @@ class TableHead extends React.Component {
   };
 
   render() {
-    const { classes, columns, count, options, data, setCellRef, selectedRows } = this.props;
+    const { classes, columns, count, options, data, setCellRef, selectedRows,
```
gabrielliwerant: If you run `npm run prettier` it will handle all the stylistic conventions.

src/components/TableFilterInline.js
```diff
@@ -0,0 +1,266 @@
...
+  handleTextFieldChange = (event, index, column) => {
+    this.props.onFilterUpdate(index, event.target.value, column, 'textField');
+  };
+
+  handleCustomChange = (value, index, column) => {
+    this.props.onFilterUpdate(index, value, column.name, column.filterType);
+  };
+
+  // TODO extend TableFilter
+  renderCheckbox(column, index) {
```
gabrielliwerant: I'd like to see each of these have the `filterList` icon (https://material-ui.com/components/material-icons/) as an input adornment (https://material-ui.com/components/text-fields/#input-adornments) prefix.

KrzysztofMadejski:
> I do have some reservations about the approach where we copy the TableFilter file and duplicate all the logic. My concern is that if there's a bug in the filters or a tweak that needs to be made, etc., the changes will need to be made in two different places, and there will be a high chance of the files diverging and creating more bugs. What we could do instead is add both inline and regular components in the TableFilter file, and then render the appropriate one based on the inline option. What do you think?

I had those doubts as well, so I actually started by doing base type and extending it. In the end though the bulk of the file contents is different and may diverge in the future: different CSS, different render methods for fields, different main render method. The only thing that is actually left duplicated are the `handleXxxChange` methods, which is very simple logic. So in my opinion it is more clear to keep it as two components that might still diverge in the future.

src/components/TableToolbar.js
KrzysztofMadejski: Will do that, it doesn't hurt I agree. Though setting `options.filter` to `false` will disable them in both places.

test/MUIDataTableFilterInline.test.js
KrzysztofMadejski: I'd keep it as separate components, but indeed the tests for them are almost identical, except testing labels & footer. Maybe the tests for both components could be aggregated in one file. What do you think?

src/components/TableFilterInline.js
KrzysztofMadejski: As below? For checkboxes I skipped it.

Screenshot 2020-01-15 at 11 41 13

KrzysztofMadejski:
@gabrielliwerant We agreed that a lot of test cases is duplicated for both components (filters in toolbar and inline). I aggregated tests for both in one file, making a few test cases separate for one or the other. Would it be ok now?

gabrielliwerant:
I can't sign off on the duplication approach taken here in general. I've worked on this myself some, and I have something messy that uses an HOC approach that returns a filter component based on whether or not the filters are inline. I did this to verify that it could work well, and it does, and cleans up most of the duplication, but it still needs work and I haven't had time to finish it. You're welcome to use this kind of approach if you want to move this forward.

KrzysztofMadejski:
Could you share your approach (the Code) ?

czw., 5 mar 2020, 16:34 użytkownik Gabriel Liwerant <
notifications@github.com> napisał:

patorjk:
@KrzysztofMadejski Just as a heads up, I've taken on maintainer responsibilities and am working on [v3 of this table](https://github.com/gregnb/mui-datatables/pull/1300). I tried out your fork and I think it's an awesome feature! However, I have the same reservations about code duplication. 
 
For example, your [renderSelect](https://github.com/KrzysztofMadejski/mui-datatables/blob/feature/1111-inline-filters/src/components/TableFilterInline.js#L202) method is only slightly different from the [v3 one](https://github.com/gregnb/mui-datatables/blob/v3/src/components/TableFilter.js#L170), and it's missing a few small updates that have happened in v3 (hence the concern about updating code in 2 places).

It seems like the only key difference is the FilterInline version not needing the wrapping GridListTile component. One possible solution would be to have a [conditional wrapping component](https://blog.hackages.io/conditionally-wrap-an-element-in-react-a8b9a47fab2). When viewed in the Popover, the filters would get the wrapping they currently have, and when viewed inline, they wouldn't have the wrapping.

KrzysztofMadejski:
@patorjk these are very helpful suggestions, thank you!

Sadly in the meantime I stopped to develop this part of the app and I don't think I'll have the time to continue to work on this PR. Feel free to close it.

PR #1122, opened 2019-12-13
Fix require path in test

Legogris:
`npm run dev` fails on current master due to invalid casing in filename. This fixes that.

### DIFF
--- a/examples/serverside-sorting/index.js
+++ b/examples/serverside-sorting/index.js
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import MUIDataTable from '../../src';
 import { CircularProgress, Typography } from '@material-ui/core';
-import Cities from './Cities';
+import Cities from './cities';
 
 class Example extends React.Component {
   state = {


gabrielliwerant:
Hm, I don't experience this issue. Can you provide details of your setup like version for npm, node, operating system, etc., so that I might gain some insight?

Legogris:
I don't suppose you're on Windows @gregnb (paths are case-insensitive on Windows but not on Linux/OS X)? There is no `Cities.json` or `Cities.js` in the folder.

patorjk:
I can verify this, it doesn't work on OS X or on codesandbox: https://codesandbox.io/s/github/gregnb/mui-datatables

I also added a fix as part of this PR: https://github.com/gregnb/mui-datatables/pull/1087

gabrielliwerant:
I don't see the problem on Windows, but I don't see any harm in changing it to lowercase, as that is how it is used elsewhere, and that is the exact filename, so it should be fixed. Thanks for pointing this out @Legogris and thanks for helping to verify @patorjk.

Legogris:
> I don't see the problem on Windows, but I don't see any harm in changing it to lowercase, as that is how it is used elsewhere, and that is the exact filename, so it should be fixed. Thanks for pointing this out @Legogris and thanks for helping to verify @patorjk.

Apart from Windows on NTFS/FAT, paths are generally case-sensitive on all filesystems and OS's in mainstream use today.

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/issues/1157.

PR #1134, opened 2020-01-04
Feature/issue 1063 responsive options and table height

gabrielliwerant:
Closes https://github.com/gregnb/mui-datatables/pull/1075, https://github.com/gregnb/mui-datatables/issues/1063 and continues by adding two new options for responsive cases:

- scrollFullHeightFullWidth
- stackedFullWidth

Both use absolute position on the paper container element to allow window width to be surrounded by the container to fix visual breakage with overflowing widths, but also allow fixed header options to operate.

Height was decoupled in the code, but not created as a separate option as height will only be supported by certain responsive modes, which is likely to create confusion. Overriding styles allows altering height values, so this functionality exists regardless, but the decoupling will make it easier to modify the API in the future if there's a good way to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ The component accepts the following props:
 |**`customFilterDialogFooter `**|function||Add a custom footer to the filter dialog. `customFilterDialogFooter(filterList: array) => React Component`
 |**`elevation`**|number|4|Shadow depth applied to Paper component
 |**`caseSensitive `**|boolean|false|Enable/disable case sensitivity for search
-|**`responsive`**|string|'stacked'|Enable/disable responsive table views. Options: 'stacked', 'scrollMaxHeight' (limits height of table), 'scrollFullHeight' (table takes on as much height as needed to display all rows set in rowsPerPage) **('scroll' option has been deprecated in favor of `scrollMaxHeight`)**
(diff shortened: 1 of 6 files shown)


maxkarkowski:
seem like that it is still an issue that scrollFullHeight has no overflow. 
setting: responsive: 'scrollFullHeight' as any,
`<div className="vpp-order-list" data-testid="vppTable">
      <MUIDataTable
        title={<Title onRefresh={handleRefresh} />}
        data={rows}
        columns={TableColumns}
        options={TableSettings}
      ></MUIDataTable>
    </div>`
overflow

gabrielliwerant:
@maxkarkowski If you have an overflowing width, then you'd probably want to use `scrollFullHeightFullWidth`. The reason there's a new option for that instead of modifying the old one is that they don't quite work the same way, so they will lead to different appearances in different situations, and `scrollFullHeight` might be the best option for someone that doesn't have width issues because the table is small (hence, it makes sense to have both).

maxkarkowski:
@gabrielliwerant thanks for the hint. i tried `scrollFullHeightFullWidth` but came across another issue as described here. [https://github.com/gregnb/mui-datatables/pull/1075#discussion_r375730961](https://github.com/gregnb/mui-datatables/pull/1075#discussion_r375730961)

so both settings won't really fit for large tables

gabrielliwerant:
@maxkarkowski The table is not guaranteed to display properly on the inside of a grid. The responsive options assume that the table is directly affected by the window dimensions. This may be something you have to experiment with by overriding styles on your end.

AndrewBenzSW:
I'm curious why the hard coded 499px for scrollMaxHeight moved from the style to the element itself.  We'd overridden the style value to good success but when we upgraded to the newest version, the override no longer worked since styled on the element is more specific than class styles, so the override no longer worked.

I was able to work around it by override with "!important" so it's not a huge deal.  I just feel icky using that :P

Thanks!

PR #1136, opened 2020-01-06
Propagate filterData to custom filters

Legogris:
This forwards `filterData` to custom filters `display` callback.

Reason: Users who want to do custom rendering but are happy with the options already calculated by mui-datatables.


patorjk:
Sorry for the delay, PR has been merged into the v3 branch and will release when v3 is ready.

PR #1144, opened 2020-01-09
Fix for columns resizing

domind:
Just limited move of columns based on padding and fixed minimum width

### DIFF
--- a/src/components/TableResize.js
+++ b/src/components/TableResize.js
@@ -94,45 +94,32 @@ class TableResize extends React.Component {
   };
 
   onResizeStart = (id, e) => {
-    this.setState({ isResize: true, id, startPosition: e.clientX });
+    this.setState({ isResize: true, id });
   };
 
   onResizeMove = (id, e) => {
-    const { startPosition, isResize, resizeCoords } = this.state;
-    const fixedMinWidth = 50;
-    let cellPositions = [];
+    const { isResize, resizeCoords } = this.state;
+    const fixedMinWidth = 100;
+    const idNumber = parseInt(id);
     const finalCells = Object.entries(this.cellsRef);
-    const elStyle2 = window.getComputedStyle(finalCells[id][1], null);
-    let previous = 0;
-    finalCells.forEach(([key, item]) => {
-      if (!item) return;
-      const elStyle = window.getComputedStyle(item, null);
-      const cellWidth = fixedMinWidth + parseFloat(elStyle.paddingLeft) + parseFloat(elStyle.paddingRight);
-      cellPositions.push(previous + cellWidth);
-      previous = previous + cellWidth;
-    });
-    const minLeft = cellPositions[id].toFixed(2);
+    if (idNumber >= finalCells.length) return;
+    if (isResize) {
+      let leftPos = e.clientX;
+      if (leftPos > resizeCoords[idNumber + 1].left - fixedMinWidth)
+        leftPos = resizeCoords[idNumber + 1].left - fixedMinWidth;
+      if (
+        (idNumber === 0 && typeof resizeCoords[0] != 'undefined') ||
+        (idNumber === 1 && typeof resizeCoords[0] == 'undefined')
+      ) {
+        if (leftPos < fixedMinWidth) leftPos = fixedMinWidth;
+      } else if (leftPos < resizeCoords[idNumber - 1].left + fixedMinWidth)
+        leftPos = resizeCoords[idNumber - 1].left + fixedMinWidth;
 
-    try {
-      const maxRight = cellPositions[parseInt(id) + 1].toFixed(2);
-      const minWidth = maxRight - minLeft;
-
-      if (isResize) {
-        let leftPos = startPosition - (startPosition - e.clientX);
-        if (leftPos > resizeCoords[parseInt(id) + 1].left - minWidth)
-          leftPos = resizeCoords[parseInt(id) + 1].left - minWidth;
-        if (parseInt(id) > 0) {
-          if (leftPos < resizeCoords[parseInt(id) - 1].left + minWidth)
-            leftPos = resizeCoords[parseInt(id) - 1].left + minWidth;
-        } else if (leftPos < minWidth) leftPos = minWidth;
-        const curCoord = { ...resizeCoords[id], left: leftPos };
-        const newResizeCoords = { ...resizeCoords, [id]: curCoord };
-        this.setState({ resizeCoords: newResizeCoords }, this.updateWidths);
-      }
-    } catch (error) {
-      return;
+      const curCoord = { ...resizeCoords[id], left: leftPos };
+      const newResizeCoords = { ...resizeCoords, [id]: curCoord };
+      this.setState({ resizeCoords: newResizeCoords }, this.updateWidths);
     }
-  };
+  }; 
 
   onResizeEnd = (id, e) => {
     this.setState({ isResize: false, id: null });
--- a/src/components/TableResize.js
+++ b/src/components/TableResize.js
@@ -99,7 +99,7 @@ class TableResize extends React.Component {
   onResizeMove = (id, e) => {
     const { isResize, resizeCoords } = this.state;
     const fixedMinWidth = 100;
-    const idNumber = parseInt(id);
+    const idNumber = parseInt(id,10);
     const finalCells = Object.entries(this.cellsRef);
     if (idNumber >= finalCells.length-1) return;
     if (isResize) {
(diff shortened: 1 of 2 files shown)


src/components/TableResize.js
```diff
@@ -94,21 +94,32 @@ class TableResize extends React.Component {
...
-
+    const { isResize, resizeCoords } = this.state;
+    const fixedMinWidth = 100;
+    const idNumber = parseInt(id);
+    const finalCells = Object.entries(this.cellsRef);
+    if (idNumber >= finalCells.length) return;
     if (isResize) {
-      const leftPos = startPosition - (startPosition - e.clientX);
+      let leftPos = e.clientX;
+      if (leftPos > resizeCoords[idNumber + 1].left - fixedMinWidth)
```
gabrielliwerant: Looking for `resizeCoords` at an array position larger than the id sometimes causes it to be undefined. Can you explain what situations would cause us to have a `resizeCoords` array that is a larger index than the number of columns we have?

src/components/TableResize.js
```diff
@@ -94,21 +94,32 @@ class TableResize extends React.Component {
...
   };
 
   onResizeMove = (id, e) => {
-    const { startPosition, isResize, resizeCoords } = this.state;
-
+    const { isResize, resizeCoords } = this.state;
+    const fixedMinWidth = 100;
+    const idNumber = parseInt(id);
+    const finalCells = Object.entries(this.cellsRef);
+    if (idNumber >= finalCells.length) return;
```
gabrielliwerant: Can you explain what situation would cause us to have an id column that is the same or greater than our total number of columns?

src/components/TableResize.js
```diff
@@ -94,45 +94,32 @@ class TableResize extends React.Component {
...
-    this.setState({ isResize: true, id, startPosition: e.clientX });
+    this.setState({ isResize: true, id });
   };
 
   onResizeMove = (id, e) => {
-    const { startPosition, isResize, resizeCoords } = this.state;
-    const fixedMinWidth = 50;
-    let cellPositions = [];
+    const { isResize, resizeCoords } = this.state;
+    const fixedMinWidth = 100;
```
gabrielliwerant: Can you explain the use of `100` here instead of `50`?

src/components/TableResize.js
```diff
@@ -94,21 +94,32 @@ class TableResize extends React.Component {
...
-    this.setState({ isResize: true, id, startPosition: e.clientX });
+    this.setState({ isResize: true, id });
   };
 
   onResizeMove = (id, e) => {
-    const { startPosition, isResize, resizeCoords } = this.state;
-
+    const { isResize, resizeCoords } = this.state;
+    const fixedMinWidth = 100;
+    const idNumber = parseInt(id);
```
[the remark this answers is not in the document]
domind: id is string type.

src/components/TableResize.js
```diff
@@ -94,21 +94,32 @@ class TableResize extends React.Component {
...
+    const finalCells = Object.entries(this.cellsRef);
+    if (idNumber >= finalCells.length) return;
     if (isResize) {
-      const leftPos = startPosition - (startPosition - e.clientX);
```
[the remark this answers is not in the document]
domind: Simple math x= a - (a -b )  is same as x= a-a +b which is same as x=b;

src/components/TableResize.js
domind: It is arbitrary number. In my previous solution was smaller, as was adding right padding of cell. However after second thought on this as anyway could not connect this with width of cell decided to skip looping through all cells to get padding array and  decided to have it fixed, but then it needed bit more space.

src/components/TableResize.js
domind: changed `if (idNumber >= finalCells.length)` to `if (idNumber >= finalCells.length-1)` - as this line of code was added to prevent error with undefined id when trying to move last resizer. Last resizer should not be moved at all.

src/components/TableResize.js
```diff
@@ -94,21 +94,32 @@ class TableResize extends React.Component {
...
+      let leftPos = e.clientX;
+      if (leftPos > resizeCoords[idNumber + 1].left - fixedMinWidth)
+        leftPos = resizeCoords[idNumber + 1].left - fixedMinWidth;
+      if (
+        (idNumber === 0 && typeof resizeCoords[0] != 'undefined') ||
+        (idNumber === 1 && typeof resizeCoords[0] == 'undefined')
```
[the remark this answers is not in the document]
domind: resizeCoords[0] is undefined when you set `selectableRows: "none"`
=== - changed this.

src/components/TableResize.js
domind: made mistake here - should be length-1 - we should not process moving for last resizer.

src/components/TableResize.js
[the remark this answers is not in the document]
gabrielliwerant: True. It didn't matter before because there weren't any calculations that depended on it being int, but I guess there are now. Let's use the radix argument here, since we can't be too careful with `parseInt`. When radix is not specified, it's behavior can vary.

src/components/TableResize.js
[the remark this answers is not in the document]
gabrielliwerant: Makes sense, thanks for explaining it to me. Since we're no longer using `startPosition`, it should be safe to remove it our state initialization as well.

src/components/TableResize.js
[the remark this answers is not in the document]
gabrielliwerant: Thanks for explaining. That's definitely something that is going to be difficult for me to remember if I need to look at this again in a few months. I think it would be better if we could code it in such a way that it documents itself a little more. Perhaps by actually using the `selectableRows` options, since this is the situation we care about?

src/components/TableResize.js
gabrielliwerant: Thanks for explaining. Your catch here raises an interesting question about the last resizer. We really don't want users even attempting to resize on the edge of the table, so I think it would be better to not render a resizer for the last column at all. That way, we don't invite confusion with resize cursors and we can simplify the logic here a bit since we know we can't possibly resize on the last column.

src/components/TableResize.js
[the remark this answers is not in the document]
domind: at present moment options and selectableRows is not passed to TableResize as props. Added comment explaining this line. Let me know if this is sufficient, or if I should pass props for selectableRows in MUIDataTable.js

src/components/TableResize.js
domind: Problem is, that with present logic, movement of second last resizer is limited by position of last one. If we will remove last one - will need to rebuild code for 2nd last to be limited by total width of table. If you want to test removing of last resizer - this can be easily done by adding `finalCells.pop();` after `finalCells `declaration in `setDividers = () => {`

src/components/TableResize.js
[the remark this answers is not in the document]
domind: removed

src/components/TableResize.js
domind: Removed last resizer - it complicated a bit checks as needed to eliminate case when last resizer is the only one, so it would be first as well. Hope this split to functions makes code bit more readable.

src/components/TableResize.js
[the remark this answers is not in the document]
domind: Forwarded all options props that are passed to MUIDataTable

src/components/TableResize.js
gabrielliwerant: Another way to do this would be to keep `resizeCoords` as is, and create a new array from it that has the final element removed, and use that array to draw the coordinates. The advantage there would be that we would not have to rethink the logic that uses `resizeCoords`. However, what you have also works!

domind: I already rewrote this logic, plus did not see a point to store this last resizer only for that purpose.
Instead to position of last resizer, prefered to keep reference to table width.

domind: @gabrielliwerant - You wait something from my end, or you are still testing and I should wait?

gabrielliwerant:
Hey @domind, I continued your PR here https://github.com/gregnb/mui-datatables/pull/1172 and would love for you to take a look to let me know what you think and to see if everything works for you on your end. I liked what you did by adding a separate functions to help make the logic a bit more readable, but I took the idea a little further to build on your work. I thought that might be simpler than a lot of back and forth over the exact structure of the logic.

Thank you so much for all of your effort. This bug has been around for a long time, so it's great to finally have something that resolves it!

kelzuriel-mihai:
Hey!
Any updates on this?

patorjk:
Closing in favor of https://github.com/gregnb/mui-datatables/pull/1172

PR #1154, opened 2020-01-15
Prevent skipping of heading level for Accessibility and standard semantic HTML purposes

lalong13:
Change Typography from `h6` to `p`.
Put text inside `strong` to keep emphasis and bold text.
http://web-accessibility.carnegiemuseums.org/foundations/semantic/

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -280,8 +280,8 @@ class TableBody extends React.Component {
               options={options}
               colIndex={0}
               rowIndex={0}>
-              <Typography variant="subtitle1" className={classes.emptyTitle}>
-                {options.textLabels.body.noMatch}
+              <Typography variant="body1" className={classes.emptyTitle}>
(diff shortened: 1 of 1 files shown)


src/components/TableBody.js
```diff
@@ -280,8 +280,8 @@ class TableBody extends React.Component {
               options={options}
               colIndex={0}
               rowIndex={0}>
-              <Typography variant="subtitle1" className={classes.emptyTitle}>
-                {options.textLabels.body.noMatch}
+              <Typography variant="body1" className={classes.emptyTitle}>
+                <strong>{options.textLabels.body.noMatch}</strong>
```
gabrielliwerant: Lets lose the `strong` tags to keep better backwards compatibility with the appearance, but otherwise looks good.

patorjk:
Thank you for the PR. Added in for v3. However, I agree with the comment on removing the strong tags, as the text is not currently bold, so I've removed them to look the look consistent.

Issue #351, opened 2019-01-07
More Custom Column Header


loganasherjones:
Thanks again for a great library. I'm interested in providing this feature and would submit a PR for it, but I think there might be API additions/changes in order to support this, so I thought I'd start with an issue.

## Expected Behavior

What I would really like to be able to do is to create a header that acts as a filter as well as a sortable header. I have a basic implementation, done right now, but the filtering is not possible with the current `customHeadRender` feature. I need more access to more of the props from the `MUIDataTable` component.

I'm not really sure what the best way to achieve this is. One thought I had was to be able to allow the API to completely replace components. i.e. I could re-write my custom `TableHead` component and then inject it something like:

```
components = {
  'TableHead': <MyCustomTableHead />
}
<MuiDataTable components={components}
```

Again, what I really want to do is to be able to create a header for a column that is capable of filtering. However you think is best to achieve this, I'm open to implementing.

gregnb:
I would like to see how you would turn this feature on and off? Need more API information. I think it's a great feature it just has to fit in nicely. Also something to consider if filter is turned on for the column header does it mean the filters in the toolbar get hidden? etc

Maaartinus:
I'd love to see it working. Some ideas:
- To switch it on, use `options.filterStyle = "header"` (better naming?).
- The filter symbol in the toolbar could be used to hide (and disable) the filters. And to show (and enable) them again.
- Doesn't make the permanently shown filter the chips redundant? I could imagine adding a cross near the filter word so it can be removed as easily as when using the chip.

gregnb:
For sure, I will keep a watch here for anymore updates

gregnb:
@loganasherjones Let me know if you make anymore updates

chupee120:
Hello,

I will also need this feature?

Any chance to have a Pull Request soon?

Thanks,

patorjk:
[Custom components](https://github.com/gregnb/mui-datatables#custom-components) have been added as of version 3.0.0.

rahul-network:
What is the repository name to get above column header changes ?

GitMurf:
Did this evolve any in the last year or two? What is the best option for folks to replicate the above sort/filter in column header logic? Do we need to build our own logic? Is the code / links from this comment the best resource to use? https://github.com/gregnb/mui-datatables/issues/351#issuecomment-453785382

Thanks!

PR #1167, opened 2020-01-23
Allow to pass custom table components

vbartusevicius:
Currently there is no way to change built-in table components. This PR enables some customization.
This is a bit different that customizing with custom Material theme as this allows to react to given props.

In example: you can change the color of `Chip` in `TableFilterList` depending on current value. Useful for filtering by some status-like fields - you will have one `Chip` indicating status with visual feedback in filter list instead of duplicating this feedback throughout entire table.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The component accepts the following props:
 |**`columns`**|array|Columns used to describe table. Must be either an array of simple strings or objects describing a column
 |**`data`**|array|Data used to describe table. Must be either an array containing objects of key/value pairs with values that are strings or numbers, or arrays of strings or numbers (Ex: data: [{"Name": "Joe", "Job Title": "Plumber", "Age": 30}, {"Name": "Jane", "Job Title": "Electrician", "Age": 45}] or data: [["Joe", "Plumber", 30], ["Jane", "Electrician", 45]]) **Use of arbitrary objects as data is not supported, and is deprecated. Consider using ids and mapping to external object data in custom renderers instead e.g. `const data = [{"Name": "Joe", "ObjectData": 123}] --> const dataToMapInCustomRender = { 123: { foo: 'bar', baz: 'qux', ... } }`**
 |**`options`**|object|Options used to describe table
+|**`components`**|object|Custom components used to render the table
 
 #### Options:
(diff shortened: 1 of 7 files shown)


vbartusevicius:
@gregnb, @gabrielliwerant - could you take alook at this?

gabrielliwerant:
@vbartusevicius There is a rather long backlog of PRs and issues, and they should be taken somewhat in order, unless it's a quick bug fix or something. This doesn't qualify, and so I'm afraid it may be a while. Additionally, I'm not too sure about this approach. The way customization tends to be handled now is through custom renderers. There's no precedent in the library for customization that works in  the way you propose, so I'm not too sure about adding it.

vbartusevicius:
@gabrielliwerant, yes there is a long backlog, but I don't see too much of activity there. I think this PR could help with cases like https://github.com/gregnb/mui-datatables/issues/351 In addition to that I intentionally introduced separate prop `components` as I think it brings more separation between purposes of props - now there is a bit of a mess: handlers, customComponents, options, configurations, etc. in one `options` prop.

If you think that it would be more reasonable to move `components` to individual `options.custom*` - I'm fine with that.

patorjk:
This is an exceptionally cool idea. Currently much of what's described here can be done with the table's various customRender functions, however, this type of API (to me at least) is much easier to understand. This approach would also solve the Tooltip component problem described in https://github.com/gregnb/mui-datatables/pull/1217 (though this PR doesn't have Tooltip available as an override so I may look into that). It also lends itself well to the new [plug-in selection](https://github.com/gregnb/mui-datatables#plug-ins), and would allow alternative forms of various components for common use cases.

The downsides I see are that if any prop name to one of these overridable components changes, the custom component will break. But these names have been pretty stable, and maybe saving such a change for a major release would be the way to go if such a situation arose. There could also be issues if any data formats changed, but customRender functions have this same issue. I may add something in the documentation about recommending the use of ~versions for apps that do lots of custom internal overrides.

Anyway, I've merged this into the upcoming v3 branch. Sorry for the delay and thank you for putting this together!

vbartusevicius:
@patorjk - thanks for appreciation!

PR #1169, opened 2020-01-27
Check for undefined search button.

### LINKED ISSUE #1168: Cannot read property focus of null
Pressing the escape key causes an undefined error `this.searchButton.focus();` cannot read property focus of null.

## Expected Behavior
No error is thrown if the search button has a null reference, handle null checking.

## Current Behavior

An error is thrown.

Type some text into a search bar, press escape, it throws an error for me. It could be due to the fact that we refetch data on changes in search text and that a ref is lost somewhere. I have no idea.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |3.93 |
| MUI-datatables  |  2.14.0       |
| React        | 16.8.6      |
| browser      |  latest chrome     |

jdetle:
Closes #1168

### DIFF
--- a/src/components/TableToolbar.js
+++ b/src/components/TableToolbar.js
@@ -217,8 +217,9 @@ class TableToolbar extends React.Component {
       showSearch: false,
       searchText: null,
     }));
-
-    this.searchButton.focus();
+    if (this.searchButton) {
+      this.searchButton.focus();
+    }
   };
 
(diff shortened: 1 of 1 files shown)


patorjk:
Looks like this is covered by https://github.com/gregnb/mui-datatables/pull/1207 which will be in the next release.

Issue #909, opened 2019-09-10
Data as array of objectswith properties which have a "dot"


rubenp9:
Hello,
I think that there is a bug when you use data as array of object with properties which have a "dot":

`render() {
    const { tableData } = this.props;

    let cols = [
      {
        name: '1',
        label: 'ID 1'
      },
      {
        name: '1_1',
        label: 'ID 1_1'
      },
      {
        name: '9.1',
        label: 'ID 9.1'
      }
    ];

    let data = [
      {
        '1': 'it works',
        '1_1': 'it works',
        '9.1': 'it does not work'
      },
      {
        '1': 'it works again',
        '1_1': 'it works again',
        '9.1': 'it does not work yet'
      }
    ];

    return (
      <MUIDataTable data={data} columns={cols} options={this.getOptions()} />
    );
  }`

To solve that I have to replace using `'9.1'.replace('.', '_')` or something similar.

Any suggestions?

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.9.3     |
| MUI-datatables  |    2.9.0     |
| React        |    16.8.6     |
| browser      |    Google Chrome Version 76.0.3809.132 (Official Build) (64-bit)     |

gabrielliwerant:
@rubenp9 Nice find. Yes, this is an unintended consequence of enabling nested data access through the standard dot notation. I'm open to suggestions about how to work around this. It's a little messy, but we could add an option like `enableNestedDataAccess` which could be flagged on or off to enabled/disable parsing the dot.

PR #1181, opened 2020-02-04
enableNestedDataAccess option to disable nested data when needed #909

domind:
also added 2 tests for  option `enableNestedDataAccess : false` :
1. making sure that table is not generated same way like when with  `enableNestedDataAccess : true` 
2. table is correctly omitting nested data.


gabrielliwerant:
Hello @domind. Can you explain the use case for shutting off this functionality via an option? It was my thinking that one would simply not use the dot notation if he/she did not want to use nested data.

domind:
Hi @gabrielliwerant you [proposed to add this as an option](https://github.com/gregnb/mui-datatables/issues/909#issuecomment-530145855)  

> I'm open to suggestions about how to work around this. It's a little messy, but we could add an option like enableNestedDataAccess which could be flagged on or off to enabled/disable parsing the dot.

Right now, by default nested data are enabled, so if there is a dot in name data are not displayed.

gabrielliwerant:
@domind Ok, thanks for reminding me. So, another way to do this would be to allow an option for choosing which character to use as a separator for objects. The advantage of this approach would be that there would be a way for someone to use this feature even if there are dots in the names, instead of having to choose between enabling/disabling the feature, What do you think?

domind:
Sounds reasonable. Should we still keep name "enableNestedDataAccess" in such case? Guess default will be enabled with dot as separator, but should we allow to disable it completely? 
Might be to rename it to nestedDataSeparator and it would take string value, and "disable" string would turn it off. Should I rewrite it that way?

domind:
@gabrielliwerant - arranged following:
- changed enableNestedDataAccess to string, so now nested data can use any separator,
- default value in case option is omitted is set to `'.'` (processing of nested data enabled)
- to disable - you need to pass empty string `enableNestedDataAccess : ''`

patorjk:
Thank you for putting this together!

I've gone ahead and merged this into v3 and added some documentation in the README.

This feature was previously mentioned only in one example, and that was for something else and wasn't even linked on the README page. So it was kind of a hidden feature. I actually think having it on by default is bad since it leads to confusing bugs as seen in https://github.com/gregnb/mui-datatables/issues/909. I think I'm going to set it to be off by default in v3. It's a breaking change, but I'll have it documented in the write-up for v2 to v3.

PR #1217, opened 2020-03-09
Injectable Tooltip component

butchmarshall:
We use an modified version of MuiTooltip within our codebase but were unable to pass through this component without completely rebuilding the header functionality.

This pull request allows for an injectable Tooltip component

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -1331,7 +1331,7 @@ class MUIDataTable extends React.Component {
   }
 
   render() {
-    const { classes, className, title } = this.props;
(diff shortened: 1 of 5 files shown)


patorjk:
Sorry there has been a delay in addressing this. This is a pretty neat idea and I agree that users should be able to add this kind of customization. However, I prefer the implementation of https://github.com/gregnb/mui-datatables/pull/1167 since it allows for many different components to be overridden. That PR didn't have Tooltips, so I've gone ahead and added them in. It's been merged into our [v3 branch](https://github.com/gregnb/mui-datatables/pull/1300) and should be released in maybe a week or two. Sorry again about the delay.

Issue #1216, opened 2020-03-09
Deleting custom filter manually displays empty chips


Xunrel:
## Expected Behavior
Removing the filter from the filter dialog should not display any chips

## Current Behavior
Removing the filter from the filter dialog shows empty chips

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/muidatatables-custom-toolbar-jk556

1. Opening filter dialog (top right)
2. Deleting the age filters by using Backspace key

## Possible workaround
If I return an empty array within the render function of customFilterListOptions as the fallback, it works as expected.
Don't know if this was intended or not.

| Tech         | Version |
|--------------|---------|
| Material-UI  | 3.2.2 |
| MUI-datatables  | 2.14.0 |
| React        | 16.4.0 |

wdh2100:
@Xunrel  Hello~
https://codesandbox.io/s/muidatatables-custom-toolbar-umbig

return false -> return []

render: v => {
              console.log("Render: ", v[0], v[1], this.state.ageFilterChecked);
              if (v[0] && v[1] && this.state.ageFilterChecked) {
                console.log("Min max");
                return [`Min Age: ${v[0]}`, `Max Age: ${v[1]}`];
              } else if (v[0] && v[1] && !this.state.ageFilterChecked) {
                console.log("Min max");
                return [`Min Age: ${v[0]}, Max Age: ${v[1]}`];
              } else if (v[0]) {
                console.log("Min");
                return `Min Age: ${v[0]}`;
              } else if (v[1]) {
                console.log("Max");
                return `Max Age: ${v[1]}`;
              }
              return [];
            },

PR #1224, opened 2020-03-14
Fix customize-filter example

wdh2100:
https://github.com/gregnb/mui-datatables/issues/1216

Hello?
I really like this library.
But the example is wrong
So I tried to fix it. 😃


wdh2100:
I think this should also be merged

https://github.com/gregnb/mui-datatables/pull/1192

PR #1244, opened 2020-04-06
Adding support for non English characters

hybmg57:
Added BOM to support non English characters

### DIFF
--- a/examples/data-as-objects/index.js
+++ b/examples/data-as-objects/index.js
@@ -4,6 +4,22 @@ import MUIDataTable from "../../src/";
 
 class Example extends React.Component {
 
+  state = {
+    counter: 0,
+    data: [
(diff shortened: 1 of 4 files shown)


wdh2100:
that's good~
👍 👍 
https://stackoverflow.com/questions/18925210/download-blob-content-using-specified-charset

patorjk:
A BOM is Excel specific and can be added using the onDownload option:

    onDownload: (buildHead, buildBody, columns, data) => {
        return "\uFEFF" + buildHead(columns) + buildBody(data); 
    } 

The problem with always adding a BOM is that it will break non-Excel spreadsheet programs, example: https://github.com/react-csv/react-csv/issues/89

For the upcoming v3 release, I've put a note in the documentation about how to add a BOM since it is something Excel user like (though it limits CSV files to Excel).

The rest of this PR seems related to table state, which should be fixed as of version [2.15.0](https://github.com/gregnb/mui-datatables/releases/tag/2.15.0) of the table. I'm going to close this PR, but if you feel like I'm missing something please open another one.

Issue #1230, opened 2020-03-20
Reversing the sorting direction


patryknawolski:
Is there an option in mui-datatables API to reverse the sorting directions in a way that after clicking sortable header column for the first time the column will be sorted in an ascending order?

## Expected Behavior
After clicking header of any sortable column for the first time should be sorted in ascending order. after clicking it second time it should be sorted in descending order.

## Current Behavior
After clicking header of any sortable column for the first time should be sorted in descending order. after clicking it second time it should be sorted in ascending order.

wdh2100:
sortDirection | string |   | Set default sort order enum('asc', 'desc', 'none') (null option has been deprecated, use 'none' instead)
-- | -- | -- | --

patryknawolski:
> sortDirection	string	 	Set default sort order enum('asc', 'desc', 'none') (null option has been deprecated, use 'none' instead)

This is for changing the default sorting, not changing the order of sorting directions.

patorjk:
Updated as of version 3.0.0

PR #1274, opened 2020-04-29
Changed first time column sort order to be ascending

skcs10:
Ref: #1230 

I agree with this issue that the first time a column header is clicked for sorting, it should use ascending order. I suspect the current behaviour is a left-over of introducing the "none" `sortDirection` option recently.

Hopefully this PR will provide an acceptable solution, if maintainers agree with the change of behaviour.


patorjk:
I agree, I tried out several other datatables and they all follow the standard you mentioned. I've merged this into the v3 branch, which should hopefully be releasing in a week or two.

PR #1304, opened 2020-06-02
Checkbox in TableHead shows checked if data is empty which is confusing on load

### LINKED ISSUE #1301: Checkbox in TableHead shows checked if data is empty which is confusing on load
## Expected Behavior
If the table has no data, the default checkbox state should be unchecked rather than checked.

## Current Behavior
Checkbox shows as checked while table data loads, but once data loads, becomes unchecked because no rows are selected.

https://codesandbox.io/s/infallible-archimedes-ulsmb?file=/src/App.js

https://github.com/gregnb/mui-datatables/blob/master/src/components/TableHead.js#L37

I think this line should become something like:

`    
let isChecked = numSelected > 0 && numSelected === count ? true : false;
`

mattcorner:
Fixes #1301

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7955,7 +7955,8 @@
           "version": "2.1.1",
           "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


wdh2100:
#1301 

>I agree, and I think your proposed fix would probably solve it. If you put in a PR I can get it into the v3 branch that's currently being worked on.

I think you should meger at v3 branch.

patorjk:
Thanks! Merged into v3 (which should release later this month).

PR #1306, opened 2020-06-03
fix webpack config

wdh2100:
Build has too many logs. 😭 
So reduce the log... change stats options(errors-warings)
ref : https://webpack.js.org/configuration/stats/

devserver host set localhost #1042 

+ remove unused import (path)

What do you think?

thanks~

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,3 @@
-const path = require('path');
 const webpack = require('webpack');
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


patorjk:
I agree with removing the unused import, are you also suggesting we change the stat param too? I'm think that would be good too. Moving it to "normal" from "verbose". None of the info "verbose" seems to be providing seems very useful.

wdh2100:
> Moving it to "normal" from "verbose". None of the info "verbose" seems to be providing seems very useful.

"verbose" provide you useful information, right?

Is this right?

Then can I just change devServer(stat : error-waring)?
Too many logs on the development server.

patorjk:
Sure, "errors-warnings" works. Also, I agree that the logs are a little much.

PR #1310, opened 2020-06-04
Added sortOrder option (deprecating sortDirection on columns object)

patorjk:
In the current API, the table's sort direction is held in one of the column options objects. This has the following problems:

* It leads to ambiguity if multiple columns have this field set.
* It's makes it hard to implement multiple level sorting (where a table is first sorted by 1 column, and then for values that are the same a second column is used to sort). You can see an example of this in [jQuery Datatables](https://datatables.net/) - click a column header, and then shift+click another header. The second column is used to break ties.
* It adds unnecessary complexity to users who use the serverSide option, since a sort value is typically passed to the server separately, and in the current implementation it burred in one of the column's options.

This PR removes sortDirection from the column's object, and creates a "sortOrder" property on the options, examplet:

    var options = {
        sortDirection: {
            name: ${selectedColumn},
            direction: ${direction}
        }
    };

<del>I may change columnName to "name" and sortDirection to "direction". </del>

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ The component accepts the following props:
 |**`onTableInit`**|function||Callback function that triggers when table state has been initialized. `function(action: string, tableState: object) => void`
 |**`setRowProps`**|function||Is called for each row and allows you to return custom props for this row based on its data. `function(row: array, dataIndex: number) => object` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-styling/index.js)
 |**`setTableProps`**|function||Is called for the table and allows you to return custom props for the table based on its data. `function() => object` [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-styling/index.js)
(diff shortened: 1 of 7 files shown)


wdh2100:
serverSide should change the option, right?

>I may change columnName to "name" and sortDirection to "direction".

me, too

```js
var options = {
    sortDirection: {
        name: ${selectedColumn},
        direction: ${direction}
    }
};
```

patorjk:
@wdh2100 I think you're right, backward compatibility for something like this is a good idea. I've updated the PR to account for "sortDirection" if no sortOrder value is present. I also updated the field names.

mxmlnglt:
@patorjk you said earlier "It's makes it hard to implement multiple level sorting": can you confirm this is not yet implemented?

patorjk:
Correct, it's not yet implemented.

broncha:
Didn't this disable sort by multiple columns? Now you can only sort by one column

Issue #1264, opened 2020-04-20
Unable to override scroll max height


maxmann74:
I'm using Material UI 4.9.2 , MUI-DataTables 2.14, React 16.13.1.

After reading other posts it is mentioned several times that you can set the max scrollheight override like this:
                      MUIDataTable: {
                        responsiveScroll: {
                          maxHeight: '400px'
                        }
                      }

I have done this, and also have options.responsive ="scrollMaxHeight"

The MUIDataTable-responsiveScrollMaxHeight div always gets rendered with height 499px regardless.   Is this still a bug, or is there another way to set the actual height of the scrolling area?

DangerousDetlef:
Create the override in the theme like so:

```
const getMuiTheme = () => createMuiTheme({
  overrides: {
    MUIDataTable: {
      responsiveScrollMaxHeight: {
        maxHeight:'400px !important' 
      }
    },
  }
})
```
In your (functional) component, load the the theme:
`const theme = getMuiTheme();`

Then create your MUI DataTable and wrap it in a MuiThemeProvider, passing the loaded theme like this:
```
    <MuiThemeProvider theme={theme}>
        <MUIDataTable
          ...
        />
      </MuiThemeProvider>
```

maxmann74:
Thanks, that worked. None of the examples I saw referenced the "responsiveScrollMaxHeight" class, and none of them used "important".   I figured override meant override :)  Good to know you still have to use "important".

Issue #1227, opened 2020-03-16
Max height in scrollMaxHeight mode


honzasram:
Hi,

First off, I'd like to thank you for everyone's hard work. It is amazing component and most of the time is is pleasure to work with it.
However there is one small and really frustrating problem. [This line](https://github.com/gregnb/mui-datatables/blob/4f836096a56be93ee1322f3c98cbaca485f0b064/src/MUIDataTable.js#L1371) contains very static value. So far i do not know how to overwrite it as it is applied as element style. Which means I can overwriting hell of styles, but this will remain there as last applied value. 

Is this intentional? How am i supposed to stretch it to given height of view? I managed to get it working in version 2.13.3, but on latest 2.14 it is with this value 

This is newer version:

![image](https://user-images.githubusercontent.com/9151345/76798188-67778a00-67cf-11ea-85f1-8d3c0d7514a6.png)

This is older one with same overrides and settings:
![image](https://user-images.githubusercontent.com/9151345/76798725-68f58200-67d0-11ea-8579-12a5ad2968de.png)

wdh2100:
ref

responsive | string | 'stacked' | Enable/disable responsive table views. Options: 'stacked', 'scrollMaxHeight' (limits height of table), 'scrollFullHeight' (table takes on as much height as needed to display all rows set in rowsPerPage), 'scrollFullHeightFullWidth' (same as 'scrollFullHeight' except that paper container wraps the table and the width takes the full browser window), 'stackedFullWidth' (same as stacked with the addition of the paper container changes with 'scrollFullHeightFullWidth') ('scroll' option has been deprecated in favor of scrollMaxHeight)
-- | -- | -- | --

honzasram:
scrollFullHeight won't work. It will allow to overflow and create scroll on whole table, not on rows section. I tried that.

scrollFullHeight - will show all rows
scrollMaxHeight - limit height and allow scrollable rows portion (keeps footer visible)

I want scrollMaxHeight to max out on my own value.

wdh2100:
That's right. The user can't change the value with the options.

wdh2100:
@gabrielliwerant  
Can I add MaxHeight to the options?

honzasram:
@wdh2100 
I went thru all issues regarding this and I had feeling like this is intentional. Reasoning is not clear to me, tough. 
Setting style after render seems to me like really ugly hack. I'm sure this is either a bug or an feature.

gabrielliwerant:
[...]

PR #1311, opened 2020-06-04
Add maxHeight option

wdh2100:
How are you?

I would appreciate it if you could contribute to the project.

This is the way to solve the problem above. (ref #1264, #1227 )

I think there's still enough to improve.

Added maxHeight option.

**Only works with 'scrollMaxHeight'**

I made it based on v3 branch


wdh2100:
I think so, too.
The response to the table is ambiguous.

>responsive:
>- "standard": Table always in what we currently know as "scrollMaxHeight" mode.
>- "vertical": The table's current "stacked" mode.

agree this

>tableBodyHeight: This is a CSS string value (ex: '500px', 'none', '100%', etc). This is just another name for the maxHeight option you suggested.

I don't know the css style well, but I think it would be good to keep the height(100%) as it is and limit the max-height.

ref : https://developer.mozilla.org/en-US/docs/Web/CSS/max-height
ref : https://developer.mozilla.org/en-US/docs/Web/CSS/height

wdh2100:
What I'm worried about is that we keep it **compatible** and apply a new concept of response.

patorjk:
> What I'm worried about is that we keep it **compatible** and apply a new concept of response.
> 
> or
> 
> Is it right to take these problems because it's a major update(3.0.0)?

I think I lean towards the first option. I always get upset when I update a library and it stops working. So a sort of middle ground has to be walked where things still work but the problems are removed.

>I don't know the css style well, but I think it would be good to keep the height(100%) as it is and limit the max-height.

I've gone ahead and put in a PR for the updates, when you have a chance I'd be really interested in your thoughts. I opted to go with adding both options in because they're subtly different:

* tableBodyMaxHeight: This caps the table at a certain height, but it's still possible for it's height to shrink. Example: A page you're looking at has 10 items, and then the next page just has 1 item.
* tableBodyHeight: The table body maintains its height regardless of the content of the table.

I've personally needed the latter option, but I imagine there are people who prefer the first option. I made the maxHeight override height, but now I think I got that backwards. Maybe height should take precedence and there shouldn't be any warning if both are set.

Issue #1323, opened 2020-06-11
ViewColumn Panel’s Close button does not work


pranavtheway:
ViewColumn Panel’s Close button does not work.

On inspecting, its found that the “X” button viewColumn Panel does not have button element.

wdh2100:
@pranavtheway  thank reporting!

There's a problem with **z-index.** 
found a bug.

I'll have to fix it soon.

patorjk:
Fixed in version 3.0.1

PR #1325, opened 2020-06-11
Fix popover bug 

wdh2100:
ref : #1323

viewColumn closeIcon not working

Fix
viewColumns

### DIFF
--- a/src/components/Popover.js
+++ b/src/components/Popover.js
@@ -13,18 +13,11 @@ const Popover = ({ className, trigger, refExit, content, ...providedProps }) => 
     open(true);
   };
 
-  const handleRequestClose = callback => {
+  const handleRequestClose = () => {
     open(false);
-    if (callback && typeof callback === 'function') {
-      callback();
-    }
   };
 
-  let closeIconClass;
-  if (providedProps.classes && providedProps.classes.closeIcon) {
-    closeIconClass = providedProps.classes.closeIcon;
-    delete providedProps.classes.closeIcon;
-  }
+  const closeIconClass = providedProps.classes.closeIcon;
 
   const transformOriginSpecs = {
     vertical: 'top',
--- a/src/components/Popover.js
+++ b/src/components/Popover.js
@@ -1,110 +1,84 @@
-import React from 'react';
+import React, { useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import MuiPopover from '@material-ui/core/Popover';
 import IconButton from '@material-ui/core/IconButton';
-import { findDOMNode } from 'react-dom';
 import CloseIcon from '@material-ui/icons/Close';
 
-class Popover extends React.Component {
-  state = {
-    open: false,
+const Popover = ({ className, trigger, refExit, content, ...providedProps }) => {
+  const [isOpen, open] = useState(false);
+  const anchorEl = useRef(null);
+
+  const handleClick = event => {
+    anchorEl.current = event.currentTarget;
+    open(true);
   };
 
-  UNSAFE_componentWillMount() {
-    this.anchorEl = null;
-  }
-
-  componentDidMount() {
-    if (this.props.refClose) {
-      this.props.refClose(this.handleRequestClose);
-    }
-  }
-
-  componentDidUpdate(prevProps, prevState) {
-    /*
-     * Update Popover position if a filter removes data from the table because
-     * it affects the window height which would cause the Popover to in the wrong place
-     */
-    if (this.state.open === true) {
-      this.anchorEl = findDOMNode(this.anchorEl);
-      if (this.popoverActions) {
-        this.popoverActions.updatePosition();
-      }
-      const shouldHide = typeof this.props.hide === 'boolean' ? this.props.hide : false;
-      if (shouldHide) {
-        this.handleRequestClose();
-      }
-    }
-  }
-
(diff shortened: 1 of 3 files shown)


src/components/Popover.js
```diff
@@ -1,110 +1,84 @@
...
 import MuiPopover from '@material-ui/core/Popover';
 import IconButton from '@material-ui/core/IconButton';
-import { findDOMNode } from 'react-dom';
 import CloseIcon from '@material-ui/icons/Close';
 
-class Popover extends React.Component {
-  state = {
-    open: false,
-  };
+const Popover = ({ className, trigger, refExit, content, ...providedProps }) => {
```
wdh2100, line 7: Changed class to function

src/components/Popover.js
```diff
@@ -1,110 +1,84 @@
...
-            <CloseIcon />
-          </IconButton>
-          {content}
-        </MuiPopover>
-        {triggerEl}
-      </React.Fragment>
-    );
-  }
-}
+Popover.propTypes = {
```
wdh2100, line 78: Define propTypes

src/components/Popover.js
```diff
@@ -1,110 +1,84 @@
...
-      key: 'content',
-      ref: el => (this.anchorEl = el),
-      onClick: () => {
-        if (trigger.props.onClick) trigger.props.onClick();
-        this.handleClick();
-      },
-    });
+  return (
+    <>
+      <span {...triggerProps}>{trigger}</span>
```
wdh2100, line 55: ref : https://reactjs.org/docs/react-api.html#cloneelement

src/components/Popover.js
```diff
@@ -1,110 +1,84 @@
...
-  };
+const Popover = ({ className, trigger, refExit, content, ...providedProps }) => {
+  const [isOpen, open] = useState(false);
+  const anchorEl = useRef(null);
 
-  UNSAFE_componentWillMount() {
-    this.anchorEl = null;
-  }
+  const handleClick = event => {
+    anchorEl.current = event.currentTarget;
```
wdh2100, line 12: ref : https://material-ui.com/api/popover/
anchorEl | HTML element\| func |   | A HTML element, or a function that returns it. It's used to set the position of the popover.
-- | -- | -- | --


The setting time(**Assign Html to anchorEl [useRef]**) has been changed by using A.

src/components/Popover.js
```diff
@@ -1,110 +1,84 @@
...
-      this.anchorEl = findDOMNode(this.anchorEl);
-      if (this.popoverActions) {
-        this.popoverActions.updatePosition();
-      }
-      const shouldHide = typeof this.props.hide === 'boolean' ? this.props.hide : false;
-      if (shouldHide) {
-        this.handleRequestClose();
-      }
-    }
+  let closeIconClass;
```
wdh2100, line 23: It seems unnecessary, but I'll keep it for now.
Because, I do not know the exact reason.

src/components/TableToolbar.js
```diff
@@ -149,7 +149,7 @@ class TableToolbar extends React.Component {
 
       // now, check columns:
       if (options.downloadOptions.filterOptions.useDisplayedColumnsOnly) {
-        columnsToDownload = columns.filter((_, index) => _.display === 'true');
+        columnsToDownload = columns.filter(_ => _.display === 'true');
```
wdh2100, line 152: unused

src/components/TableViewCol.js
```diff
@@ -14,6 +14,7 @@ export const defaultViewColStyles = theme => ({
   },
   title: {
     marginLeft: '-7px',
+    marginRight: '24px',
```
wdh2100, line 17: Refer to the image

src/components/TableToolbar.js
```diff
@@ -353,6 +352,7 @@ class TableToolbar extends React.Component {
           {options.viewColumns && (
             <Popover
               refExit={this.setActiveIcon.bind(null)}
+              classes={{ closeIcon: classes.filterCloseIcon }}
```
wdh2100, line 355: Add missing props

patorjk:
Looks like this bug snuck in with the X icon that was added to the filter dialog (I didn't realize it got added to the ViewColumns popover too).

I like the hook refactor, however, I noticed a bug. If you open up the serverside-filters example, the filter dialog no longer closes when you push "Apply Filters". 

It looks like this refactor doesn't include the componentDidUpdate code. I'm actually not sure what the "popoverActions" code does, and I can't get the filter dialog to move (like the comment says), so I'm wondering if that code is out of date.

However, the second part of the componentDidUpdate routine is the code that allows the dialog to be closed. You should be able to add a useEffect hook to rectify this. I'd look something like this: (you'd need to pull "hide" from the props)

```
  useEffect(() => {
    if (isOpen) {
      const shouldHide = typeof hide === "boolean" ? hide : false;
      if (shouldHide) {
        open(false);
      }
    }
  }, [hide, isOpen, open]);
```

Other than that, nicely done!

wdh2100:
@patorjk Great!!

I missed it because I was working with the old version. Thank you.

patorjk:
This looks good to go. I think I'll wait a few more days to see if any other bugs are found and then do a 3.0.1 release then, maybe Monday or Tuesday this coming week.

Issue #1307, opened 2020-06-03
Filter data by multiple columns values


qweluke:
Let's say I have a table with 2 columns: Age 1 and Age 2.
![image](https://user-images.githubusercontent.com/193701/83638886-2317b100-a5aa-11ea-820c-7f43f9cd2fd6.png)

Now I need to filter rows that have some value in Age 1 or Age 2

https://codesandbox.io/s/muidatatables-custom-toolbar-9n5fq?fontsize=14&hidenavigation=1&theme=dark

I know I can use `filterOptions` for certain column and use `logic(value, filterVal) {..}` - but it contains value from current cell. How to get data from other column in my filter?

patorjk:
The way to do this would be to create another column with the values you want to filter on and hide it's display, and then hide the filtering for the two other columns. Example:

    const columns = [{name: "Name"}, {name: "Title"}, {name: "Location"}, 
    {
      name: "Age",
      options: {
        filter: false,
      }
    }, {
      name:"Age 2",
      options: {
        filter: false,
      }
    },
    {
      name: "Ages",
      options: {
        display: false,
      }
    }];

https://codesandbox.io/s/muidatatables-custom-toolbar-7fvfk?file=/index.js

wdh2100:
ref : https://github.com/gregnb/mui-datatables/blob/master/examples/customize-filter/index.js

andrew-vogulkin:
@wdh2100 i have a similar issue but i am thinking that the best way to do that is extend 
`column.filterOptions.filterOptions` call with row and rowIndex, 
i would like to push my branch with changes, how i can get access ?

wdh2100:
@andrew-vogulkin  

>how i can get access ?

Is this means, You want to push your code?

Of course it's possible.
Read the reference below.

thanks,

ref : 
- https://opensource.com/article/19/7/create-pull-request-github 
-https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

patorjk:
Resolved in version 3.0.1 - a new "rows" field has been added to the logic callback which allows access to all data within the row.

PR #1328, opened 2020-06-12
Add row param for column.filterOptions.logic

andrew-vogulkin:
With all entire row object in logic fn filtering can be more powerful 

Related with: https://github.com/gregnb/mui-datatables/issues/1307


patorjk:
This seems to be a good update. Since it's small, I'll add it into the 3.0.1 release, which should happen in the next few days.

PR #1329, opened 2020-06-12
Draggable Columns

patorjk:
This PR adds the ability to drag and drop columns. I wanted a nice column swap effect while dragging. You can try it out here: (see the **Draggable Columns** example)

https://codesandbox.io/s/github/gregnb/mui-datatables/tree/draggable_cols

I am debating if draggableColumns should be enabled by default - as otherwise people may not know it's there. But for the initial release I think it should default to off.

**API updates:**

Option to turn on and customize draggable columns:

```
draggableColumns: {
  enabled: boolean,
  transitionTime: number
}
```

Option indicating the column order (defaults to the order in the columns prop):

```columnOrder: array```

A "draggable" option to a column's option object indicating if the column can be dragged:

```columns.options.draggable: boolean```

**Todo:**

* So far I've only tested on Chrome/Safari/Firefox on Mac. I'll do Windows testing next week.
* Get it working with resizableColumns feature.
* Tests.

### CODE REVIEW

File: src/components/TableHead.js [hunk 1]
```diff
@@ -19,115 +19,133 @@ const defaultHeadStyles = theme => ({
...
+        [classes.responsiveStacked]:
+          options.responsive === 'vertical' ||
+          options.responsive === 'stacked' ||
+          options.responsive === 'stackedFullWidth',
+        [classes.responsiveSimple]: options.responsive === 'simple',
+        [classes.main]: true,
+      })}>
+      <TableHeadRow>
+        <TableSelectCell
+          ref={el => setCellRef(0, findDOMNode(el))}
```
wdh2100, line 98: findDOMNode is oldStyle and heavy

ref : 
- https://reactjs.org/docs/react-dom.html#finddomnode
- https://medium.com/trabe/getting-rid-of-finddomnode-method-in-your-react-application-a0d7093b2660

src/components/TableHead.js
```diff
@@ -19,115 +19,133 @@ const defaultHeadStyles = theme => ({
...
+    reorderColumns,
+    updateColumnOrder,
+    columnOrder,
+    headCellRefs,
+    timers,
+    toggleSort,
+    tableRef,
+    sortOrder = {},
+    components = {},
+  } = props;
```
wdh2100, line 45: It would be better to apply PropTypes. 😄

src/components/TableBody.js
```diff
@@ -210,8 +210,21 @@ class TableBody extends React.Component {
...
+    var ret = [];
+
+    for (var ii = 0; ii < row.length; ii++) {
+      ret.push({
+        value: row[ columnOrder[ii] ],
+        index: columnOrder[ii]
+      });
+    }
+
+    return ret;
```
wdh2100, line 223: It would be better to reduce the use of (var, let).

Especially in the case of **'var'**, you should avoid using it.

In the case below, it seems to be unnecessary use (var).
The behavior is the same even if you change to **const**.
Because the reference point of the variable has not changed.

```js
const ret = [];
ret.push(1); // [1]
console.log(ret) // [1]

var ret = [];
ret.push(1); // [1]
console.log(ret) // [1]
```

I think it will disappear if webpack is applied, 
but I think it would be better to use it properly from the beginning. 

thanks

examples/draggable-columns/index.js
```diff
@@ -0,0 +1,111 @@
...
+    ["Jack Jackson", "Business Analyst", "El Paso","1","555-4321"],
+    ["Joe Jones", "Computer Programmer", "El Paso","1","555-4321"],
+    ["Jacky Jackson", "Business Consultant", "Baltimore","4","555-1234"],
+    ["Jo Jo", "Software Developer", "Washington DC","4","555-1122"],
+    ["Donna Marie", "Business Manager", "Annapolis","5","555-5555"],
+
+  ];
+
+  return (
+    <React.Fragment>
```
wdh2100, line 54: What do you think?
```html
<React.Fragment/>

</>
```
ref : https://reactjs.org/docs/fragments.html#short-syntax

src/MUIDataTable.js
```diff
@@ -19,6 +19,8 @@ import DefaultTableToolbarSelect from './components/TableToolbarSelect';
 import MuiTooltip from '@material-ui/core/Tooltip';
```
wdh2100, line 19: This component is getting very complicated. (Because it's the **main**...) 😭

What do you think about this?

Is there a better way?

wdh2100:
```js
  const options = {
    resizableColumns: true,
    draggableColumns: {
      enabled: true,
      transitionTime
    }
  };
```

I think it's not working.
Please check.

src/components/TableHead.js
patorjk: Good catch, I wasn't aware this was on its way out. This code is a hold over from back when Greg originally did the table resize stuff. I'll try and refactor it out as a see it. Looks like there are 8 instances of its use in the code base total.

src/components/TableHead.js
patorjk: Looks like this component never had any, I'll add some in for it - however, I'll probably do it last just because things may change.

src/components/TableBody.js
patorjk: Agreed, I'll go back and replace the vars, old habits die hard :P

examples/draggable-columns/index.js
patorjk: Agreed

src/MUIDataTable.js
patorjk: It is, though it'll inherently be the most complex piece. I think I'll try and move some of what I wrote out of main and only keep what's essential there.

patorjk:
> ```js
>   const options = {
>     resizableColumns: true,
>     draggableColumns: {
>       enabled: true,
>       transitionTime
>     }
>   };
> ```
> 
> I think it's not working.
> Please check.
> 
> OR
> 
> Can't they(resize, drag) be used at the same time?

I knew I was forgetting something! The tableresize stuff is really the only area of the table I'm not that familiar with. Worst case, draggable columns can be disabled for it. However, I don't really see any reason why it shouldn't be able to work with it, so I'll investigate this. Thanks for catching this!

patorjk:
It should now work with resizable columns.

src/components/TableHead.js
```diff
@@ -98,7 +99,7 @@ function TableHead(props) {
       })}>
       <TableHeadRow>
         <TableSelectCell
-          ref={el => setCellRef(0, findDOMNode(el))}
+          ref={el => setCellRef(0, 0, findDOMNode(el))}
```
wdh2100, line 102: I think we should try to get rid of A findDOMNode

wdh2100: I think I found a way. ( ref callback )
I'll commit it soon.

patorjk: I think that may be the last one too, if you figure out a way just submit a PR and I'll merge it into this branch.

TalhaK26:
Hey @patorjk, 
Draggable Columns not working with customHeadRenders. Is there any solution?

patorjk:
Because customHeaderRender replaces [TableHeadCell](https://github.com/gregnb/mui-datatables/blob/master/src/components/TableHeadCell.js) (see: https://github.com/gregnb/mui-datatables/blob/master/src/components/TableHead.js#L132), you have to implement the logic inside of TableHeadCell if you want draggable columns.

Alternatively, what are you using customHeaderRender for? Note that there is now a customHeadLabelRender option that will allow you to render for the header label without having to override the TableHeadCell component. If you use this method, you can still use draggable columns.

TalhaK26:
Thanks @patorjk,
customHeadLabelRender works for me.

Issue #1280, opened 2020-05-08
left coordinate with resizableColumns option affects dividers position


kiryl-fively:
## Expected Behavior

When you set resizableColumns option,  moving left coordinate of the table or its wrapper shouldn't affect dividers position and columns width. 

## Current Behavior

Currently, when you set resizableColumns option and for example add left margin/padding to the table wrapper, dividers are rendered in wrong positions so the last one overflows container.

## Steps to Reproduce (for bugs)

1. Set resizableColumns option
2. Set either left margin/padding or change left position of the table component or add some block to the left of the table. Basically you need to change left coordinate of the table rectangle.

## Your Environment

https://codesandbox.io/s/muidatatables-resizeable-columns-nk0yt
| Tech         | Version |
|--------------|---------|
| Material-UI  |     4.9.0    |
| MUI-datatables  |  2.14.0       |
| React        |     16.12.0    |
| browser      |    Google Chrome Version 81.0.4044.129 (Official Build) (64-bit)     |

patorjk:
I can confirm this is a bug, I'll try and dig deeper if I get some time.

patorjk:
I've created a PR to resolve this issue. It can be viewed here: https://github.com/gregnb/mui-datatables/pull/1330

patorjk:
Resolved in version 3.0.1

PR #1330, opened 2020-06-13
Bug fix for #1280 / Fix issues with resizableColumns

patorjk:
The resizableColumns feature currently doesn't work when there is left margin on the container for the table (see https://github.com/gregnb/mui-datatables/issues/1280).

This PR makes some adjustments so that the resizableColumns work with leftMargin. The PR can be viewed below: (see the "Resizable Columns" example)

https://codesandbox.io/s/github/gregnb/mui-datatables/tree/resizeLeftMargin

### DIFF
--- a/src/components/TableResize.js
+++ b/src/components/TableResize.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { findDOMNode } from 'react-dom';
 import { withStyles } from '@material-ui/core/styles';
 
 const defaultResizeStyles = {
@@ -54,6 +53,7 @@ class TableResize extends React.Component {
   };
 
   componentDidMount() {
+    this.minWidths = [];
     this.windowWidth = null;
     this.props.setResizeable(this.setCellRefs);
     this.props.updateDividers(() => this.setState({ updateCoords: true }, () => this.updateWidths));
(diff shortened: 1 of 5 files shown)


patorjk:
I believe this PR now fixes all of the issues surrounding the resizableColumns feature. It can be tried out in the codesandbox in my above post in the "Resizable Columns" example.

I'm a little surprised there hasn't been more interest in this feature as it's really cool. After some adjustments it seems to work flawlessly now.

src/components/TableResize.js
```diff
@@ -98,7 +95,18 @@ class TableResize extends React.Component {
     const { resizeCoords, tableWidth } = this.state;
 
     Object.entries(resizeCoords).forEach(([key, item]) => {
-      let newWidth = Number(((item.left - lastPosition) / tableWidth) * 100).toFixed(2);
+      let newWidth = Number(((item.left - lastPosition) / tableWidth) * 100);
```
wdh2100, line 98: Decimal points are likely to cause bugs.

Can't we just throw away the decimal point?
I think we can throw away the decimal point.

```js
const newWidth = Number(((item.left - lastPosition) / tableWidth) * 100).toFixed(0)
```

patorjk: I've done a little more research on this and decimal places will [not cause issues](https://cruft.io/posts/percentage-calculations-in-ie/). I was mostly just being overly cautious. What perplexed me is why the code was rounding to 2 decimal places. It appears that in the old days browsers rounded to 2 decimal places, so possibility for consistency it was set to round to 2. 

For me (Chrome/Firefox/Safari) on Mac, there is a noticeable jitteriness when moving the dividers when only 2 decimal places are used. When there is no cap (and the browser does the rounding), it looks nice and smooth. Anyway, since a longer decimal place won't cause an issue, I think I'm going to leave it be. Certain browsers will round the number differently, but a larger decimal point allows newer browsers to be more precise.

PR #1338, opened 2020-06-16
Convert Functional component with hooks

wdh2100:
**Change from Class Component to Functional component**

Hello~ 😃 

React seems to be changing to a functional component(= pure component). (my opinion)

We need to be a functional component to use a lot of cool **Hooks**.

I'll go ahead with the test to see if I have any performance gains.

But in my opinion ❓ , I feel more comfortable with the functional component than the class component. 

I'm going to need a lot of tests because I'm going to change the whole component 

thanks!

happy coding~

p.s I'd like to make a **CodeSandBox** that the changes here, what should I do?

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -5,70 +5,54 @@ import MuiTableBody from '@material-ui/core/TableBody';
 import TableBodyCell from './TableBodyCell';
 import TableBodyRow from './TableBodyRow';
 import TableSelectCell from './TableSelectCell';
-import { withStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/core/styles';
 import cloneDeep from 'lodash.clonedeep';
 import { getPageValue } from '../utils';
 import classNames from 'classnames';
 
-const defaultBodyStyles = theme => ({
-  root: {},
-  emptyTitle: {
-    textAlign: 'center',
-  },
-  lastStackedCell: {
-    [theme.breakpoints.down('sm')]: {
-      '& td:last-child': {
-        borderBottom: 'none',
+const useStyles = makeStyles(
+  theme => ({
+    root: {},
+    emptyTitle: {
+      textAlign: 'center',
+    },
+    lastStackedCell: {
+      [theme.breakpoints.down('sm')]: {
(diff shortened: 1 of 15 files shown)


patorjk:
I agree that hooks are the way to go, I'd be curious if there was any performance improvement. You might try looking at the Large Data Set example, There is a render time that prints to the console.

I think I'd like to see this done in multiple PRs though (maybe with 3 refactored files per PR). I hate to ask you to resubmit, but this is a really big PR and covers a lot of surface area. It'd be easier to test and accept in chunks. Also, I'd avoid refactoring these components: tablebody, tablehead, tableheadcell, tableresize, and tabletoolbar - because major changes occur to them in the draggable columns PR, and it'd be easier to refactor them into hooks after that update was merged in (I think some of them become hooks in that PR too).

I think certain components will need to stay classes (ex: MUIDataTable), but where it makes sense we should definitely use hooks. They seem a lot easier to work with. Anyway, thanks for putting this together and sorry for asking you to break it into chunks.

patorjk:
> p.s I'd like to make a CodeSandBox that the changes here, what should I do?

CodesSandbox will auto create a sandbox for you based on your branch name, all you have to do is use this base URL:

http://codesandbox.io/s/github/

And combine it with the branch info:

wdh2100/mui-datatables/tree/FunctionComponent/withUseHooks

To get this:

http://codesandbox.io/s/github/wdh2100/mui-datatables/tree/FunctionComponent/withUseHooks

I always just copy the link at the top of the PR (where it says "wdh2100 wants to merge 15 commits into gregnb:master from wdh2100:FunctionComponent/withUseHooks"), and then modify it to use http://codesandbox.io/s/github/

patorjk:
Actually, the more I think about it, it's probably best to do a single PR for a single refactored file. That way if there's an issue with 1 file, things don't get held up.

wdh2100:
>Actually, the more I think about it, it's probably best to do a single PR for a single refactored file. That way if there's an issue with 1 file, things don't get held up.

@patorjk 

thanks your answer

If there's no relations, I'll separate pull request.

wdh2100:
Change #1346

Issue #1040, opened 2019-10-30
"Deprecated: Passing objects in as data is not supported, and will be prevented in a future release."


dandrei:
Just upgraded to version `2.12.4` and I get the following:

`Deprecated: Passing objects in as data is not supported, and will be prevented in a future release. Consider using ids in your data and linking it to external objects if you want to access object data from custom render functions.`

I am currently using this option as a work-around in order to access the original object from the rendering function, [as described in my reply to issue 475](https://github.com/gregnb/mui-datatables/issues/475#issuecomment-471047711).

Other recently opened issues (like [](https://github.com/gregnb/mui-datatables/issues/1038)) are also asking for a way to access the original row object.

Is this feature being considered for inclusion in the future?

Thanks.

gabrielliwerant:
As discussed here https://github.com/gregnb/mui-datatables/issues/1011, the table has never officially supported passing object data for table cells, and doing so leads to a number of errors if not handled precisely in custom renders. If you're using a custom render, you're stepping into custom territory, and there's not any strong case I can see for accepting objects for this case. As explained in the issue, you can map something like ids back to arbitrary data/objects. They don't need to be passed into the table to get you the information you need.

gabrielliwerant:
Closing this, as there is no planned support for passing arbitrary data structures into the table. As discussed in https://github.com/gregnb/mui-datatables/issues/1038, I would accept something that added `dataIndex` to the `customBodyRender` callback which would help alleviate some of these issues.

nathancharles:
Since #1011 is closed, leaving a comment here.
I agree that supporting objects for cell data would really create a more flexible and powerful framework.

It is not uncommon for a table cell to include more than just a number or string. For example, having a user avatar that has an image url, name and email is a pretty common use-case. Conceptually, and practically, it makes sense to have this data in one place, instead of some placeholder that will need to be mapped back. Mapping to some external data can be very clunky to implement. It makes sense that a user will want to be able to search any visible data in the table too, but this puts limitations on searching and require accessing an external data source to search the table.

Not having to re-map and restructure the data you need for every render can give some performance benefits as well. If there is some expensive operation, having to only do it once while constructing the table data is much less overhead rather than managing this on every cell re-render.

[...]

Issue #915, opened 2019-09-11
customBodyRender bug


Lilliance:
Hello!

When i trying to provide `customBodyRender` function in column options its broke filter with that column.

Errors heading looks like:
1. "Failed prop type: Invalid prop `primary` supplied to `ForwardRef(ListItemText)`, expected a ReactNode."
2. "Failed prop type: Invalid prop `children` supplied to `ForwardRef(Typography)`, expected a ReactNode."
3. "Invariant Violation: Objects are not valid as a React child (found: object with keys {id, name}). If you meant to render a collection of children, use an array instead."

I trying to use this `customBodyRender` function:
```
(field: { id: number; name: string }) => {
  return (
    <span style={{ color: STATUS_MAP[field.name] || 'inherit' }}>
      {field.name}
    </span>
  );
};
```

When i trying to return only `field.name` string, without `<span></span>` wrapper all works fine.

Also, your [example](https://github.com/gregnb/mui-datatables/blob/master/examples/component/index.js) works.

What i missed?

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.3.3     |
| MUI-datatables  |    2.1.8     |
| React        |    16.8.6     |
| browser      |    any     |

gabrielliwerant:
Hello @Lilliance. You list your version of this library here as `2.1.8`, is that correct? Can you confirm that the issue still exists in the latest version, which is `2.10.3`?

Lilliance:
Thank you for advice, @gabrielliwerant. I updated library to `2.10.3`, but i have same issue. :(

gabrielliwerant:
It looks like be using incorrect arguments in the function. Check out the `component` example in the library. The arguments are `value`,, `tableMeta`, `updateValue`. `value` is a string, and it is the value intended for the cells in the column you are customizing.

Lilliance:
When i don't use any args, i have same issue. Like this:
```
customBodyRender: () => {
  return (
    <span>
      hello! I am hardcoded text. 
    </span>
  );
};
```

gabrielliwerant:
I'll need to see an example of your problem with all the (minimal) code necessary to create the issue. Can you create a codesandbox (https://codesandbox.io) and share it here?

Lilliance:
Sure, you can check my [example](https://codesandbox.io/embed/heuristic-brook-rzp7d)
I noticed one more thing:
[...]

PR #1339, opened 2020-06-17
Added customBodyRenderLite / removed object deprecation for column values

patorjk:
This PR:

* Adds customBodyRenderLite to a column's option object.
* Removes the deprecation warning for object data, as it will now be allowed.

## customBodyRenderLite

customBodyRenderLite is a way of rendering data in a more performant way. Using it can lead to tables that render quicker. Before going further, here is a brief technical description of how customBodyRender works under the covers:

- When setting the table data, it is executed once for every row of data so that the table can get the values for the filterData.
- When computing the data for the display, it is again executed once for every row (the results are cached in the displayData). 

If the data changes, both of the above happen again. This means if you have a table of 1000 rows and 10 columns (all with customBodyRender), A total of 1000 * 10 * 2 = 20,000 function calls get executed related to rendering.

Unlike customBodyRender, the value returned from customBodyRenderLite is **not** used for filtering. So it's return value has no affect on filters. Instead, filters use the value in the data array. Additionally, unlike customBodyRender, customBodyRenderLite is only called for cells that are visible on the current page. This means for the same table, if you have 10 rows per page, a total of 10 * 10 = 100 function calls get made. A test showing this dynamic has been added.

Another trade off for customBodyRenderLite is that it doesn't take in `(value, tableMeta, update)`, instead it simply takes in `(dataIndex, rowIndex)`. This means you have to lookup a value yourself (though personally I've found having the dataIndex is much more useful).

## Objects no longer deprecated

Back in release 2.12.2, a deprecation warning was added stating that objects would no longer be accepted as values for data (with the exception of arrays). This occurred because of this issue: https://github.com/gregnb/mui-datatables/issues/915. Essentially what was happening was the filter dialog would cause the table to blow up when given object data.

Forbidding objects is overkill though, as being able to input them is extremely useful. I've patched the error that occurred for the filter dialog. There are two basic problems that remain:

* customBodyRender or customBodyRenderLite is required for object data to be displayed (otherwise the table throws an error). People previously did not seem to have an issue with this (https://github.com/gregnb/mui-datatables/issues/1040). I've added a note to the readme about it.
* Filtering does not work out of the box for object data. Users who want to filter with object data will need to write their own logic. I think I'm fine with this. No one prior to 2.12.2 seemed to inquire about it, and typically these kind of fields are non-text based, so filtering makes no sense anyway.


src/components/TableBodyCell.js
```diff
@@ -134,7 +134,7 @@ function TableBodyCell(props) {
         className,
       )}
       {...otherProps}>
-      {children}
+      {typeof children === 'function' ? children(dataIndex, rowIndex) : children}
```
wdh2100, line 137: Maybe children could be a function.
I thought it was only a react node(element)

Does children mean columns?

patorjk: Yeah, it's a pattern called "Function as Child Component". It's useful in certain situations. Here the children equal either the value that will be displayed in the column, or the customBodyRenderLite function, which will generate the value for the column.

wdh2100: @patorjk  thanks!

Issue #1334, opened 2020-06-15
How to order filters independently from the columns?


mxmlnglt:
Hello,

We're using MUIDataTable for a while now, because that's the best table-oriented component the previous front developer had chosen.
Now he's not here, we have to try our best to make it meet our users' expectation.

That's why I'm opening a few issues regarding various aspects of customization of MUIDataTable.
For this purpose, I've forked your Codesandbox and created a little showcase of all our issues:
https://codesandbox.io/s/muidatatables-misc-requests-4vkxk

First of all: is there a way to reorder the filters in the filter dialog?
In fact, we noticed the order of the filters is linked to the order of the columns, so we tried to reorder our columns; unfortunately this isn't always wanted.

In our example, we'd like to have both "Former Role" and "New Role" filters side-by-side. How can we do this?

Current:
![MUIDataTable_filters_not_nearby](https://user-images.githubusercontent.com/29626929/84711663-25451c80-af67-11ea-83b9-2966754bdddb.png)

Expected:
![MUIDataTable_filters_now_nearby](https://user-images.githubusercontent.com/29626929/84711678-2a09d080-af67-11ea-96d6-87cbd39d1d51.png)

Note: to be able to "create" the 2nd screenshot, I "hacked" the live document and set `width: 100%` (instead of 50%) for the `<li class="MuiGridListTile-root"` element wrapping the Name filter.

patorjk:
Hello,

I'm the maintainer of this project. My plan for the summer is to try to do a release every week on Sunday evenings (either patch or minor, depending on what's ready). As you stated, currently the order of filters is determined by the column order:

https://github.com/gregnb/mui-datatables/blob/master/src/components/TableFilter.js#L377

However, for what you're doing, I think a simpler solution may work. All you really need is for the "Name" filter to stretch across the dialog. Right now the filter dialog is a grid 2 cells wide. I could add another parameter to [filterOptions](https://github.com/gregnb/mui-datatables#column-options) that would be a boolean variable that specified if the filter should stretch across the dialog. Example:

      {
        name: 'Name',
        options: {
          filter: true,
          filterOptions: {
            fullWidth: true
          },
          filterType: 'textField'
        },
      },

If this works let me know. You can do the PR or I can take a look into it later this week.

mxmlnglt:
Thanks for you answer, this looks interesting. But for the moment, I have little time to investigate in the MUIDataTable code, and as this does not have a high priority, I'll try to have a look later on.

[...]

PR #1340, opened 2020-06-17
Added filterOptions.fullWidth

patorjk:
Closes https://github.com/gregnb/mui-datatables/issues/1334

Also updated documentation for filterOptions.


wdh2100:
> const cols = (column.filterOptions && column.filterOptions.fullWidth) === true ? 2 : 1;

Good feature. but, I see a lot of duplicated code.

patorjk:
It could probably be optimized some, I'll leave it as is for now though. Ideally, the component render methods could be re-used for something like what's done here - https://github.com/gregnb/mui-datatables/pull/1120. That feature is pretty cool but it duplicates this file, which would just lead to headaches (I would have had to edit two files for this PR).

Issue #1364, opened 2020-06-22
Column Drag and Drop issue with excluded columns


mcmastermind:
I am experiencing an issue when dragging columns when there are columns with the display property set to exclude. It looks like, in useColumnDrop.js the headCellRefs array indexes do not align with the columnOrder argument.

## Expected Behavior

Columns should be able to drag and drop smoothly regardless of display property.

## Current Behavior

useColumnDrop.js throws and error on line 38 when columns with the property `display: 'exclude'` is set.

Error: `TypeError: Cannot read property 'offsetWidth' of undefined`

## Steps to Reproduce

[Sandbox Example](https://codesandbox.io/s/muidatatables-resize-columns-example-oyzwx)

1. Define your columns
2. Set one column with `display: 'exclude'`
3. Render datatable
4. Try to drag a column

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.10.2     |
| MUI-datatables  |    3.1.1     |
| React        |    16.13.1     |
| browser      |    FireFox     |

patorjk:
Thank you for the report. I have a fix working in dev. I'm going to test it a little more and see if I can resolve an issue another person is having but hopefully a patch will be out soon.

patorjk:
If you have a moment, can you try out this beta version I just published:

npm install mui-datatables@3.1.2-beta.0

I believe it should resolve this issue.

mcmastermind:
I am still experiencing the same behavior. I also updated the [code sandbox](https://codesandbox.io/s/muidatatables-resize-columns-example-oyzwx) to use the beta version as well

patorjk:
Should be fixed in mui-datatables@3.1.2-beta.1, let me know if this works for you:

https://codesandbox.io/s/muidatatables-drag-columns-error-nzf03

It looks like the issue that time was the "exclude" is an invalid value for display (display takes in "true", "false" or "excluded"). However, this problem was hidden because the table only displayed columns if display equaled "true", making it seem like "exclude" worked. However, if you were to click the ViewColumn button in the toolbar, the column would be there and you could show it (since that dropdown menu was based on the input "excluded").

In mui-datatables@3.1.2-beta.1 I've updated the code to account for this. "true" and "false" work as expected, and anything other than those two values basically acts as the "excluded" input.

mcmastermind:
Works like a charm! Keep up the good work, and thank you for tackling this so quickly.

patorjk:
[...]

Issue #1363, opened 2020-06-22
Dnd with 2 tables, 2 routes, redux. "Cannot have two HTML5 backends at the same time"


AkselAllas:
I have 2 routes: route1 and route2, which both render a MUI-datatable.
In 3.1.1, when I switch from route1 to route2, i get an error:

`Cannot have two HTML5 backends at the same time`

Stacktrace shows that this is a dnd error. As such, I downgraded to 3.0.1 and now the error doesn't happen.

I hope this is useful. I can provide more details if necessary on Thursday. 

Full stack trace of error:
```

HTML5BackendImpl.setup
node_modules/react-dnd-html5-backend/dist/esm/HTML5BackendImpl.js:403
DragDropManagerImpl.handleRefCountChange
node_modules/dnd-core/dist/esm/DragDropManagerImpl.js:38
dispatch
node_modules/redux/es/redux.js:222
HandlerRegistryImpl.addTarget
node_modules/dnd-core/dist/esm/HandlerRegistryImpl.js:101
registerTarget
node_modules/react-dnd/dist/esm/common/registration.js:3
registerHandler
node_modules/react-dnd/dist/esm/hooks/internal/drop.js:54
commitHookEffectListMount
node_modules/react-dom/cjs/react-dom.development.js:19731
commitLifeCycles
node_modules/react-dom/cjs/react-dom.development.js:19787
commitLayoutEffects
node_modules/react-dom/cjs/react-dom.development.js:22803
HTMLUnknownElement.callCallback
node_modules/react-dom/cjs/react-dom.development.js:188
invokeGuardedCallbackDev
node_modules/react-dom/cjs/react-dom.development.js:237
invokeGuardedCallback
node_modules/react-dom/cjs/react-dom.development.js:292
commitRootImpl
node_modules/react-dom/cjs/react-dom.development.js:22541
unstable_runWithPriority
node_modules/scheduler/cjs/scheduler.development.js:653
runWithPriority$1
node_modules/react-dom/cjs/react-dom.development.js:11039
commitRoot
node_modules/react-dom/cjs/react-dom.development.js:22381
finishSyncRender
node_modules/react-dom/cjs/react-dom.development.js:21807
performSyncWorkOnRoot
node_modules/react-dom/cjs/react-dom.development.js:21793
(anonymous function)
node_modules/react-dom/cjs/react-dom.development.js:11089
unstable_runWithPriority
node_modules/scheduler/cjs/scheduler.development.js:653
[...]

Issue #1359, opened 2020-06-22
Uncaught TypeError: Cannot read property 'offsetParent' of undefined when draggableColumns is enabled


achmadk:
## Expected Behavior

Draggable columns can work properly

## Current Behavior

Get error `Uncaught TypeError: Cannot read property 'offsetParent' of undefined`

## Steps to Reproduce (for bugs)

1. Open [https://dashboard-preview.web.app/timah](https://dashboard-preview.web.app/timah), please wait until all charts and tables has been loaded.
2. Click inspect element, then open console tab.
3. Select table you want to drag its column, e.g. Tender Amount by Vendor.
4. Drag one column inside that table.
5. When that column is dragging, check the console that displays `Uncaught TypeError: Cannot read property 'offsetParent' of undefined`. 

## Your Environment

| Tech         | Version |
|--------------|---------|
| `@material-ui/core`  | 4.10.2        |
| `mui-datatables`  | 3.1.0        |
| `react`        | 16.13.1        |
| browser      | Google Chrome on Linux (Ubuntu 20.04)       |

patorjk:
Yikes, that's a bad bug :/. It only happens when selectableRows: "none".  I've got a fix in development. I'm going to do a little more testing and hopefully push a 3.1.1 version a little later tonight.

achmadk:
thanks for your quick response, @patorjk .

patorjk:
No problem, thank you for the quick report! I've push out a 3.1.1 patch that has draggable columns working when selectableRows="none". If you see anything else please let me know.

achmadk:
I have updated `mui-datatables` to 3.1.1. I have forgot that the website use iframe. Drag and drop feature still doesn't work for me. Only display warning both [Violation] 'dragstart' and [Violation] 'drop' handler into console tab.

patorjk:
Can you show me an example? I can't seem to re-create this.

patorjk:
I'm aiming to do another patch release tonight. I'm not sure if this will solve your problem, but if you have a moment can you try this beta version:

npm install mui-datatables@3.1.2-beta.0

achmadk:
sorry, that beta version still doesn't fix this issue.

patorjk:
Can you show me an example? I'm unable to reproduce the problem.

wdh2100:
@patorjk 

I think I need to pass `window` to use it in `iframe`

ref : 
[...]

PR #1366, opened 2020-06-22
version 3.1.2 (fixes)

patorjk:
This release would fix the following:

* Allow draggable columns to work with excluded columns (https://github.com/gregnb/mui-datatables/issues/1364).
* Make it so cursor isn't a hand if the column isn't draggable.

### DIFF
--- a/src/components/TableHeadCell.js
+++ b/src/components/TableHeadCell.js
@@ -116,11 +116,14 @@ const TableHeadCell = ({
     item: {
       type: 'HEADER',
       colIndex: index,
+      headCellRefs: draggableHeadCellRefs
     },
     begin: monitor => {
-      setHintTooltipOpen(false);
-      setSortTooltipOpen(false);
-      setDragging(true);
+      setTimeout(() => {
+        setHintTooltipOpen(false);
+        setSortTooltipOpen(false);
+        setDragging(true);
+      }, 0);
       return null;
     },
     end: (item, monitor) => {
@@ -151,7 +154,6 @@ const TableHeadCell = ({
 
   const isDraggingEnabled = () => {
     if (!draggingHook) return false;
-
     return options.draggableColumns && options.draggableColumns.enabled && column.draggable !== false;
   };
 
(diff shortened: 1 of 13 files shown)


src/components/TableHeadCell.js
```diff
@@ -116,11 +116,14 @@ const TableHeadCell = ({
     item: {
       type: 'HEADER',
       colIndex: index,
+      headCellRefs: draggableHeadCellRefs,
     },
     begin: monitor => {
-      setHintTooltipOpen(false);
-      setSortTooltipOpen(false);
-      setDragging(true);
+      setTimeout(() => {
```
wdh2100, line 122: Can you tell what this is❓ 

why use setTimeout..

ref : https://react-dnd.github.io/react-dnd/docs/api/use-drag

begin(monitor): Optional. Fired when a drag operation begins. Nothing needs to be returned, but if an object is returned it will override the default item property of the spec.

src/components/TableHeadCell.js
```diff
@@ -116,11 +116,14 @@ const TableHeadCell = ({
...
+        setHintTooltipOpen(false);
+        setSortTooltipOpen(false);
+        setDragging(true);
+      }, 0);
```
wdh2100, line 126: Can you tell what this is❓ 

why use setTimeout..

ref : https://react-dnd.github.io/react-dnd/docs/api/use-drag

begin(monitor): Optional. Fired when a drag operation begins. Nothing needs to be returned, but if an object is returned it will override the default item property of the spec.

patorjk: An attempt to solve https://github.com/gregnb/mui-datatables/issues/1359 - I cannot reproduce that issue. I found this thread: https://github.com/react-dnd/react-dnd/issues/1085 - it had a workaround that seemed related to what the guy was describing. Wrapping in setTimeout doesn't really have ill effects, it's just odd. I also can't reproduce https://github.com/gregnb/mui-datatables/issues/1363. I tried creating an example with other react-dnd componets but everything seems to work fine. I found this issue: https://github.com/react-dnd/react-dnd/issues/1571 and it had a harmless solution so I added that in. Hopefully that will resolve the problem that guy is seeing.

wdh2100: @patorjk 

Thanks for the kind explanation.

wdh2100: It would be okay to change to `Immediately-invoked function expression`.

I'll test it and tell you again

PR #1385, opened 2020-07-02
Update onViewColumnsChange to onColumnViewChange in API doc

GuanYangCLU:
Used customized function onColumnViewChange searched from web but in doc it shows onViewColumnsChange and not found this in module.


patorjk:
What version of the library are you using? This was updated in version 3 and both methods should work:

https://github.com/gregnb/mui-datatables/blob/master/src/MUIDataTable.js#L1094

I'm debating removing the deprecation warning though and just adding a note in the documentation that onColumnViewChange is now an alias of onColumnViewChange.

This was a name change that was requested since the feature is referred to as viewColumns on the columns object.

PR #1387, opened 2020-07-03
Added setFilterChipProps

patorjk:
New option that allows you to add props to the Filter Chip component. Works identically to the other setXProps methods. Example:

      setFilterChipProps: (colIndex, colName, data) => {
        //console.log(colIndex, colName, data);
        return {
          color: 'primary',
          variant: 'outlined',
          className: 'testClass123',
        };
      }

Useful for easily customizing to a filter chip. This can already be done through the custom component feature, but doing it that way is a bit of work. Using setFilterChipProps would make customization a lot easier.

### DIFF
--- a/examples/customize-filter/index.js
+++ b/examples/customize-filter/index.js
@@ -235,6 +235,14 @@ class Example extends React.Component {
       filter: true,
       filterType: 'multiselect',
       responsive: 'standard',
+      setFilterChipProps: (colIndex, colName, data) => {
+        //console.log(colIndex, colName, data);
+        return {
+          color: 'primary',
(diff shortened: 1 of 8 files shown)


examples/customize-filter/index.js
```diff
@@ -235,6 +235,14 @@ class Example extends React.Component {
       filter: true,
       filterType: 'multiselect',
       responsive: 'standard',
+      setFilterChipProps: (colIndex, colName, data) => {
+        //console.log(colIndex, colName, data);
```
wdh2100, line 239: I would rather delete it than comment it

Issue #1261, opened 2020-04-17
React.StrictMode prevent removing filter


RedFish:
I am testing some features of this library into my app, and I have noticed that the closing icon of active filter is not working.

Some warnings appear as well in the console:
- Warning: Using UNSAFE_componentWillMount in strict mode is not recommended and may indicate bugs in your code. See https://fb.me/react-unsafe-component-lifecycles for details.

- Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of t which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
    in th (created by ForwardRef(TableCell))
    in ForwardRef(TableCell) (created by WithStyles(ForwardRef(TableCell)))

## Expected Behavior
Remove the filter

## Current Behavior
Nothing append

## Steps to Reproduce
https://codesandbox.io/s/muidatatables-custom-toolbar-0nhte

1. <React.StrictMode><App /></React.StrictMode>
2. Add a filter
3. Remove added filter from the X icon

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |     4.9.9    |
| MUI-datatables  |    2.14.0     |
| React        |     16.13.1    |
| browser      |     chrome Version 80.0.3987.163    |

wdh2100:
https://github.com/gregnb/mui-datatables/issues/679 
😢

patorjk:
Hmm, I think both these issues can be taken care of without much of an issue. I think the last of the findDOMNode calls will be removed after the [draggable columns](https://github.com/gregnb/mui-datatables/pull/1329) PR is complete, and refactoring out UNSAFE_componentWillMount seems pretty straight forward: There are 2-3 methods that need to return objects rather than setting the state. They can then be called from the constructor.

I'll put this on my list.

johnsonsamuel:
Hi team, I am seeing the same issue.

```
Using UNSAFE_componentWillMount in strict mode is not recommended and may indicate bugs in your code
Move code with side effects to componentDidMount, and set initial state in the constructor.
```
Please advise.

johnsonsamuel:
@RedFish Were you able to solve this issue?

patorjk:
The findDOMNode code was refactored away in the last release. The UNSAFE_componentWillMount part will be tackled in the next release (probably in a week or so).

patorjk:
[...]

PR #1388, opened 2020-07-03
Refactored out unsafe_componentWillMount

patorjk:
The table currently uses the method unsafe_componentWillMount in the main component. React is removing this lifecycle method in version 17 and it currently gives a warning message to developers using strict mode (https://github.com/gregnb/mui-datatables/issues/1261). The method is deemed unsafe because it can lead to memory leaks if not used properly.

React recommends using the constructor is initialize state or componentDidMount for initialization that requires DOM nodes. Only the former is relevant to mui-datatables, so the code has been refactored to use the constructor.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -259,34 +259,8 @@ class MUIDataTable extends React.Component {
     },
   };
 
-  state = {
-    activeColumn: null,
-    announceText: null,
-    count: 0,
-    columns: [],
-    expandedRows: {
-      data: [],
-      lookup: {},
-    },
-    data: [],
-    displayData: [],
-    filterData: [],
-    filterList: [],
-    page: 0,
-    previousSelectedRow: null,
-    rowsPerPage: 0,
-    searchProps: {},
-    searchText: null,
-    selectedRows: {
-      data: [],
-      lookup: {},
-    },
-    showResponsive: false,
-    sortOrder: {},
-  };
-
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.tableRef = React.createRef();
     this.tableContent = React.createRef();
     this.draggableHeadCellRefs = {};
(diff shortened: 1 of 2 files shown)


src/MUIDataTable.js
```diff
@@ -259,52 +259,56 @@ class MUIDataTable extends React.Component {
...
     this.tableContent = React.createRef();
     this.draggableHeadCellRefs = {};
     this.resizeHeadCellRefs = {};
     this.timers = {};
     this.setHeadResizeable = () => {};
     this.updateDividers = () => {};
-  }
 
-  UNSAFE_componentWillMount() {
-    this.initializeTable(this.props);
```
wdh2100, line 300: initializeTable has changed the position of componentDidMount.

src/MUIDataTable.js
```diff
@@ -259,52 +259,56 @@ class MUIDataTable extends React.Component {
...
+    this.setTableData(props, TABLE_LOAD.INITIAL, true, null, true);
   }
 
   componentDidMount() {
     this.setHeadResizeable(this.resizeHeadCellRefs, this.tableRef);
 
     // When we have a search, we must reset page to view it unless on serverSide since paging is handled by the user.
     if (this.props.options.searchText && !this.props.options.serverSide) this.setState({ page: 0 });
+
+    this.setTableInit('tableInitialized');
```
wdh2100, line 311: But shouldn't this be the first thing to do?

```js
componentDidMount() {
  this.setTableInit('tableInitialized');

  ...
}
```

src/MUIDataTable.js
```diff
@@ -677,14 +670,15 @@ class MUIDataTable extends React.Component {
     return transformedData;
   };
 
-  setTableData(props, status, dataUpdated, callback = () => {}) {
+  setTableData(props, status, dataUpdated, callback = () => {}, fromConstructor = false) {
```
wdh2100, line 673: It doesn't seem necessary to add fromConstructor.
Because it can be judged by the existence of callback.

And I think it's better to separate the function.

I think it has too heavy functions.

wdh2100:
It is good to remove **unsafe_componentWillMount**.

it is such a great work!

wdh2100:
Now that I've been thinking about it, 

 it's a **refectoring**, so you'd better focus logic.

You can take the above reviews lightly.

thanks!

src/MUIDataTable.js
patorjk: Is this an issue?

src/MUIDataTable.js
patorjk: To keep consistent, it probably should. I'll move it up.

src/MUIDataTable.js
patorjk: The constructor sets a default function for that callback. I also wanted to make it explicit at what was happening, since the constructor is the only place where the unique behavior happens.

src/MUIDataTable.js
wdh2100: Nope,
I left a trace while reading the code. Sorry. 🙏

src/MUIDataTable.js
wdh2100: >The constructor sets a default function for that callback. I also wanted to make it explicit at what was happening, since the constructor is the only place where the unique behavior happens.

Oh, I see. There is a default. hmmm...

PR #1398, opened 2020-07-09
Added jumpToPage option

patorjk:
This option, if true, adds a "Jump to Page" dropdown to the table footer that allows users to navigate to a specific page.


src/components/JumpToPage.js
```diff
@@ -0,0 +1,107 @@
...
+    textLabels,
+    rowsPerPage,
+    page,
+    changePage
+  } = props;
+
+  const textLabel = textLabels.pagination.jumpToPage;
+
+    let pages = [];
+  let lastPage = Math.min( 1000, getPageValue(count, rowsPerPage, 1000000));
```
wdh2100, line 63: In the case of ServerSide, it seems that we need to find the value in another way.

For example, Add  `lastPage` props, options

src/components/JumpToPage.js
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {  
+  InputBase,  
+  MenuItem,  
+  Select,  
+  Toolbar,  
+  Typography
+} from '@material-ui/core';
+import { withStyles } from '@material-ui/core/styles';
```
wdh2100, line 10: Method Higher-order component(`withStyles`)  looks better than method Hooks(`makeStyles`).

Especially in the case of functional component

`withStyles` to `makeStyles`

However, it is not an unconditional task.
It has nothing to do with behavior | logic.

src/components/JumpToPage.js
patorjk: It'll still work. When using serverSide, you're supposed to have the total count of the table (but you only display the current page of entries). This feature is actually ported over from a fork I was using and I have a bunch of serverSide tables.

src/components/JumpToPage.js
patorjk: Great catch! I first had the component as a class but then decided to make it a hook, however, I must have gotten distracted and forgot the withStyles/makeStyles part. Kind of interesting that everything still worked fine. I'll push an update here shortly though.

PR #1400, opened 2020-07-10
update link for custom component example

atwayne:


patorjk:
Nice find, thank you!

Issue #1394, opened 2020-07-08
Sorting on Salary doesnt work


pranavtheway:
I have a table in which there is a salary field which is rounded and formatted to use comma separated figures within like 1,23,344

Now when I try to apply sort it doesnt work.

My sandbox url is: https://codesandbox.io/s/quiet-feather-q4jn4

patorjk:
It's sorting by string. The table will look at the data, and if it's a number, it'll sort it like a number, and if it's a string, it'll sort it like a string. If you want a custom sort, you have to use the customSort option:

https://codesandbox.io/s/quizzical-kalam-eeeqx?file=/src/App.js

Note: That's just a quick and dirty example, and probably doesn't sort exactly how you want. But that's how you override how sorting works.

Also, I noticed your Start Date wasn't sorting correctly either. This is because the table will use the value returned from customBodyRender for sorting (which is a string). To avoid this, you can use customBodyRenderLite (which doesn't effect filtering or sorting), I updated the example above to show this.

pranavtheway:
Hey That was good lead though, I missed something in my example which I noticed later, that I didnt saved my sandbox after change by implementing the formatNumber function to all the salaries, the solution or the direction didnt worked with it...

This is the updated sandbox link with your solution intact, not working: 

https://codesandbox.io/s/elated-lehmann-u00pr

patorjk:
You have to process your input, parseInt can't handle commas. Here's a version where the commas are removed:

https://codesandbox.io/s/inspiring-lederberg-nde7o?file=/src/App.js

Essentially, inside of customSort you'll want to transform your data to numbers so they can be compared, or transform them to strings that would sort correctly when compared.

pranavtheway:
Thats an excellent solution!!! Thank You So Much!!

Issue #1336, opened 2020-06-15
How to sort empty values at the end the table? (via customSort?)


mxmlnglt:
Hello,

We're using MUIDataTable for a while now, because that's the best table-oriented component the previous front developer had chosen.
Now he's not here, we have to try our best to make it meet our users' expectation.

That's why I'm opening a few issues regarding various aspects of customization of MUIDataTable.
For this purpose, I've forked your Codesandbox and created a little showcase of all our issues:
https://codesandbox.io/s/muidatatables-misc-requests-4vkxk

Some columns can have no data (either empty string or null), and nothing get display: it's OK for us.

We have set a default (ascendant) sort on the "New Role" column.
In our example, the first 2 rows with "New Role" = empty appear at the top of the table.
We want them to appear at the end of the table (i.e. on page 2), the same way it's possible in SQL (at least on Oracle) to ask for `NULLS LAST`.

I suppose this is not feasible "by default", so probably will require implementing a `customSort`.
But how can we implement such a function without implementing a full default sorting function from scratch? I mean, would it be possible to simply re-use the default MUIDataTable sort function?

patorjk:
There's an [example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-sorting/index.js) on how to do customSort, however, I kind of like your use case as the example just does a miscellaneous sort. A customSort for what you're trying to do would look something like this:

      customSort: (data, colIndex, order) => {
        return data.sort((a, b) => {
          var aVal = a.data[colIndex];
          var bVal = b.data[colIndex];

          if (!aVal) aVal = "ZZZ";
          if (!bVal) bVal = "ZZZ";

          if (aVal < bVal) {
            return -1 * (order === "desc" ? -1 : 1);
          } else if (aVal > bVal) {
            return 1 * (order === "desc" ? -1 : 1);
          } else {
            return 0;
          }
        });
      }

Sandbox example:

https://codesandbox.io/s/muidatatables-misc-requests-j68nt?file=/index.js

wdh2100:
OR https://codesandbox.io/s/muidatatables-misc-requests-6c1rj?file=/index.js:2495-2587

replace Data

```jsx
    <MUIDataTable
        title={"ACME Employee list"}
        data={data.map(item =>
          item.flatMap(innerItem => (innerItem ? innerItem : ""))
        )}
        columns={columns}
        options={options}
      />
```

mxmlnglt:
[...]

PR #1424, opened 2020-07-19
Added support for custom sort function for columns

patorjk:
This PR is a continuation of https://github.com/gregnb/mui-datatables/pull/951. I've simply updated the code for the latest branch, expanded upon the example, and made it so sortCompare takes precedence over customSort.

This feature has been requested/come up several times recently (https://github.com/gregnb/mui-datatables/issues/1011#issuecomment-645651798, https://github.com/gregnb/mui-datatables/issues/1336#issuecomment-648682275, https://github.com/gregnb/mui-datatables/issues/1394). 

sortCompare Interface
---------
```typescript
(order: 'asc'|'desc') => ({data: val1}, {data: val2}) => number
```

example:
```javascript
columns = [
  {
    name: "Name",
    options: {
      sort: true,
      sortCompare: order => ({data: val1}, {data: val2}) => (val1.length - val2.length) * (order === 'asc' ? 1 : -1)
    }
  }
]
```

If both customSort and sortCompare are defined, sortCompare will take precedence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ The component accepts the following props:
 |**`customRowRender `**|function||Override default row rendering with custom function. `customRowRender(data, dataIndex, rowIndex) => React Component`
 |**`customSearch `**|function||Override default search with custom function. `customSearch(searchQuery: string, currentRow: array, columns: array) => boolean`
 |**`customSearchRender `**|function||Render a custom table search. `customSearchRender(searchText: string, handleSearch, hideSearch, options) => React Component`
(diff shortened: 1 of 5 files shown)


wdh2100:
Can I suggest a name?

I think it would be more intuitive to change the name to compareFunction  from this [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

patorjk:
There already is a sort parameter though (it’s Boolean for if the column can be used for sorting).

I’d be open to another name. sortFn doesn’t really line up with the libraries names. What do you think of “customSort”? It’d make it inline with the name on options, though I’m not sure if that would be confusing. Another idea: “sortCompare” - though it’s kind of odd that order is required. This could be refactored out though. I think I like sortCompare, what are your thoughts?

PR #1428, opened 2020-07-21
Reset sort

danielbyun:
Resets column sort on the third click.


patorjk:
I think this idea, but this PR modifies 44 files and changes 18k+ lines of code. I think you may have done the change in an older branch. Can you do the change from the current code base?

danielbyun:
Oops, will do! Will create a new PR also.

PR #1430, opened 2020-07-22
Reset sorting

danielbyun:
Resets sorting on the third click

### DIFF
--- a/examples/simple/index.js
+++ b/examples/simple/index.js
@@ -20,7 +20,6 @@ function Example() {
     responsive,
     tableBodyHeight,
     tableBodyMaxHeight,
-    sortOrder: {},
     onTableChange: (action, state) => {
       console.log(action);
       console.dir(state);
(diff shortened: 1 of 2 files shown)


patorjk:
Thank you for putting this together! Just tried it out and it works well. After thinking a bit, this will probably need to be manually turned on through an option (with the default behavior being toggling between asc and desc). It occurred to me that users using tables with serverSide=true may not be anticipating a "none" value and it could break people.

Looking at other tables, some ideas are thirdSortClick, sortThreeClickReset, threeStateSort. Any thoughts on a name for this option?

danielbyun:
I think `sortThreeClickReset` or `sortThirdClickRest` clearly states what the purpose of this prop is so... Any of those sound good!

I'm glad I can contribute even a little bit to the library that has helped us so much in organizing and then displaying our chaotic data haha

Thank you!

Issue #1435, opened 2020-07-24
Set manually "allSelectedRows" checkbox true/false on table loading


tecnoher:
I want to be able to set selectAllRows checkbox (in header) on/off when datatable is loaded (as response of some initial condition)

This is my code, I use customRowRender prop.

`<Grid container className={classes.root}>
<Grid item xs={12} className={classes.tableJobPositions}>
  <MUIDataTable
    title={'Puestos'}
    data={dataJob}
    columns={columnsJob}
    options={optionsJp()}
  />
</Grid>

<Grid item xs={12} className={classes.organizationalUnits}>
  <MUIDataTable
    title={'Centros de costos'}
    data={dataOrg}
    columns={columnsOrg}
    options={optionsOu()}
  />
</Grid>
</Grid>

const optionsOu = () => ({
  download: false,
  print: false,
  rowsPerPage: 10,
  selectableRows: 'multiple',
  sort: false,
  fixedHeader: true,
  filter: false,
  selectableRowsHeader: true,
  onTableChange: (action, tableState) => {
    //this.props.outerState().isDeselect = true;
    onSelectAllChange(tableState.selectedRows.data.length);
  },
  customRowRender: organizationalUnitsRowRender,
  textLabels: {
    body: {
      noMatch: "Lo sentimos, no hay registros",
      toolTip: "Orden",
      columnHeaderTooltip: column => `Orden para ${column.label}`
    },
    selectedRows: {
      text: "filas(s) seleccionadas"
    },
  }
});

const organizationalUnitsRowRender = (row, index) => {
  let [id, code, description] = row;
  let isSelected = isOuRowSelected(id);
  return (
    <TableRow
      hover
      role="checkbox"
      aria-checked={isSelected}
      tabIndex={-1}
      key={'ou' + id}
      selected={isSelected}
    >
      <TableCell
        padding="checkbox"
        onClick={() => { handleClickOu({ id, code, description }) }}
      >
        <Checkbox checked={isSelected} color={'primary'} />
      </TableCell>
      <TableCell align="left">{id}</TableCell>
      <TableCell align="left">{code}</TableCell>
      <TableCell align="left">{description}</TableCell>
    </TableRow>
  );
};`

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    3.0.1     |
| MUI-datatables  |    2.14.0     |
| React        |    16.13.1     |
[...]

PR #1438, opened 2020-07-24
Enhancement/rowSelected options

wdh2100:
#1435 

People seem to make a lot of mistakes when using it.
In addition, in case of `undefined`, it behaves abnormally.

So I'm trying to fix it this way.

how do you think?


patorjk:
I think this is a good idea, I'll add it into the 3.3.0 release.

PR #1456, opened 2020-08-02
feat($general): make TableFilter swappable

andreiglingeanu:
This pull request makes the `TableFilter` component swappable from client code. This is very useful in cases when special customizations needs to be done in the filter dialog. The alternative to this is overriding the `TableToolbar` component and swap `TableFilter` from there, which is not desirable at all.


patorjk:
Looks good, I'll include this in the next release.

PR #1472, opened 2020-08-15
feat: add option to pass tableId as a prop

### LINKED ISSUE #1452: Warning: Prop `data-tableid` did not match. Server: "08323920907465618" Client: "06902255999981364"
## Expected Behavior
No warning about prop mismatch should appear when doing SSR (e.g. with next.js).

## Current Behavior

A warning about prop mismatch appears when using mui-datatables in a next.js context:
screenshot

See https://codesandbox.io/s/exciting-dhawan-fsrl0 for reproduction.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |  4.11       |
| MUI-datatables  |  3.3.1       |
| React        |     16.13    |

The issue is related to the prop `data-tableid`, which is only used for dragging of columns. Turning off the `draggable` option for all columns does not solve the issue.

I took a look into the code and did not fully grasp why `data-tableid` has to be a dynamic value. The only case that would make sense to me is that multiple tables are rendered at once. If that is the reason, it might a possibility to hand over responsibility to the developer and require him to pass a unique identifier to each table.

rfermann:
This PR adds the option to pass a `tableId` prop to the `<MUIDataTable />` component.

Additionally it sets a default `tableId` when `options.draggableColumns.enable` is set to `true` and the `tableId` prop hasn't been provided by the user.

It fixes #1452.

@patorjk i'm not used to enzyme as a testing library and I struggled finding a way to properly test for the availability of the `data-tableid` attribute. Hence I haven't added test cases for now. Could you guide me a bit in this?

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -218,6 +218,7 @@ class MUIDataTable extends React.Component {
       selectableRowsHideCheckboxes: PropTypes.bool,
       selectableRowsOnClick: PropTypes.bool,
       serverSide: PropTypes.bool,
+      tableId: PropTypes.string,
       tableBodyHeight: PropTypes.string,
       tableBodyMaxHeight: PropTypes.string,
       renderExpandableRow: PropTypes.func,
@@ -268,7 +269,6 @@ class MUIDataTable extends React.Component {
 
   constructor(props) {
     super(props);
-    this.tableId = (Math.random() + '').replace(/\./, '');
     this.tableRef = React.createRef();
     this.tableContent = React.createRef();
     this.draggableHeadCellRefs = {};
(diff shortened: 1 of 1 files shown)


patorjk:
I haven't tried it out yet but it looks good.

As for a test, you could probably add a small one that included using the tableId option, and then did a check to see if it got added to the cells.

Example: https://github.com/gregnb/mui-datatables/blob/master/test/MUIDataTable.test.js#L437

```
    let count = mountWrapper
      .find('[data-tableid=' + options.tableId + ']')
      .length;

    assert.strictEqual(count, 10); // I'm not sure what the right number would be, the parameter would be added to THs and TDs though
```

Issue #1479, opened 2020-08-17
filterPos incorrect when filterList contains object types


JeffJankowski:
## Expected Behavior
When `filterList` is populated with non-primitive object types, `customFilterListOptions.update: function(filterList, filterPos, index) => filterList` should provide a proper `filterPos` index to the custom update function.

My current use-case has a filter list with the following data shape for complex logic when fetching remote data:
```
interface Filter {
  label: string;
  value: string | number | boolean;
  key: string;
}

filterList: Filter[][]
```

I need `update` to return a correct `filterPos` to properly remove chips in a drill-down filtering scenario (eg. country -> state).

## Current Behavior
`filterPos` is always returning `-1`. I assume this is because of the naive equality check of reference types via `indexOf`:
https://github.com/gregnb/mui-datatables/blob/a382f3e8e850c7fcdd395f8594a5e0afefcaf905/src/MUIDataTable.js#L1372

I can't figure out a way around this issue without forking the repo (which I'd like to avoid). Possible solutions:
1. Add additional prop that allows me to set my own, custom comparator
1. Return filter `value` instead of array index (`filterPos`) to do my own comparison logic in the `update` function
1. Deep-object equality check instead of naive reference equality (probably not ideal)

## Steps to Reproduce (for bugs)
I can provide a codesandox environment if requested.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.10.2     |
| MUI-datatables  |    3.3.1     |
| React        |    16.3.1     |
| browser      |         |
| etc          |         |

patorjk:
Can you provide a codesandbox? 

I've been terribly busy recently, I would accept a PR to fix this.

JeffJankowski:
@patorjk PR created: #1481

JeffJankowski:
Merged into master in v3.4.1 #1487

PR #1481, opened 2020-08-18
Bugfix/1479: Return Correct filterPos using Deep-Object Equality

JeffJankowski:
Issue Ticket: #1479 

The library is already pulling in `lodash.isequal`, so I decided to leverage that function for a deep-object comparison in the `Array.findIndex()` function predicate.

Personally, I think it would be better to either: pass in a custom comparator as a prop, or return the filter `value` instead of `filterPos` to allow the user to define their own logic in `update()`. However, the deep-object comparison method does not break any existing contracts/functionality or require adding an additional prop.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -5,6 +5,7 @@ import clsx from 'clsx';
 import assignwith from 'lodash.assignwith';
 import cloneDeep from 'lodash.clonedeep';
 import find from 'lodash.find';
+import isEqual from 'lodash.isequal';
 import isUndefined from 'lodash.isundefined';
 import merge from 'lodash.merge';
 import PropTypes from 'prop-types';
@@ -1369,7 +1370,7 @@ class MUIDataTable extends React.Component {
   };
 
   updateFilterByType = (filterList, index, value, type, customUpdate) => {
-    const filterPos = filterList[index].indexOf(value);
+    const filterPos = filterList[index].findIndex(filter => isEqual(filter, value));
 
     switch (type) {
       case 'checkbox':


wdh2100:
ref : https://codesandbox.io/s/github/JeffJankowski/mui-datatables/tree/bugfix/issue-1479-incorrect-filterpos

wdh2100:
That's a great idea.

However, you should consider search(Search is based on the entire data. `But for now, only the features in the filter have been modified.`), Deep-Object Equality is good. However, it becomes a problem when there are  mixture of and data types

```js
  const sampleData = [
    ["Gabby George", 1, "Minneapolis"],
    [
      "Aiden Lloyd",
      "Business Consultant for an International Company and CEO of Tony's Burger Palace",
      "Dallas"
    ],
    ["Jaden Collins", "1", "Santa Ana"],
    ["Franky Rees", "Business Analyst", "St. Petersburg"],
    ["Aaren Rose", null, "Toledo"],
    ["Johnny Jones", "Business Analyst", "St. Petersburg"],
    ["Jimmy Johns", "Business Analyst", "Baltimore"],
    ["Jack Jackson", "Business Analyst", "El Paso"],
    ["Joe Jones", "Computer Programmer", "El Paso"],
    ["Jacky Jackson", "Business Consultant", "Baltimore"],
    ["Jo Jo", 1, "Washington DC"],
    ["Donna Marie", "1", "Annapolis"]
  ];
```

JeffJankowski:
@wdh2100 Can you expand on your comment a little more? This is just a fix for a bug I encountered with `customFilterListOptions.update`.

wdh2100:
@JeffJankowski 

Yes. I know you tried to fix the problem.
But another bug arises(mixture of and data types, Search result and filter result are different ...).

JeffJankowski:
@wdh2100 Does it manifest in the codesandbox of my PR you posted? I'm just trying to understand the new issue. I would assume that `lodash.isEqual` does the naive equality check (`===`) on passed values that are primitive instead of objects; so I'm not sure why this would reduce a regression.

wdh2100:
@JeffJankowski 

Sorry. 
I think I misunderstood the problem.. 😢 

This PR works fine.

PR #1482, opened 2020-08-19
Update display to accept either boolean or string

w3bdesign:
Typescript definitions have been updated for the display property, and it accepts either a boolean (true or false) or a string ("excluded").

As a result the documentation should reflect this.


w3bdesign:
@patorjk

PR #1506, opened 2020-09-04
Fallback for falsy customRowRender

petrkrejcik:
If you want to apply a customRowRender only to a specific row you can fallback to the default renderer by returning falsy result from customRowRender. It might be useful when you want to edit a specific line (render inputs instead of values).

Maybe the PR should also introduce a new option `fallbackCustomRowRender` to not create a breaking change. Currently returning `null` simply doesn't render the row.

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -241,7 +241,10 @@ class TableBody extends React.Component {
             const { data: row, dataIndex } = data;
 
             if (options.customRowRender) {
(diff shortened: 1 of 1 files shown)


petrkrejcik:
Closing PR because the desired result is possible to achieve with `customBodyRender`.

iamkk11:
> Closing PR because the desired result is possible to achieve with `customBodyRender`.

How?

PR #1518, opened 2020-09-16
Add support for 'any match' filter for array values

AntoineLecacheurInsee:
This adds a "filterArrayFullMatch" option to the table that defaults to true, to decide how to handle array filtering.

Default remains the same : if you have several filter values selected, for a row to show, the column value needs to include every filter values.
If filterArrayFullMatch is false though, then we check if at least one of the filter values is in the column value.

I made some changes to the array-value-columns example to include both use cases, I changed the example component to be a function which is why the commit is a bit long.

I thought this use case I added should be pretty common, and would be interesting to add support for.
Also this is my first PR, so let me know if you're interested with this option, and if so tell me if anything is wrong.


patorjk:
This looks good to me, seems to work well. I can add this to the next release. Another user said they would try out the grouping feature for me, so I may do a release of that and this later this week or early next week. 

Ref: https://codesandbox.io/s/github/AntoineLecacheurInsee/mui-datatables/tree/filter-array-fullmatch

PR #1535, opened 2020-10-02
Adding title props in both backIconButtonProps and nextIconButtonProps

Ruriel:
Hello, guys. This is my first Github contribution. I noticed that the textLabel for 'next' and 'previous' were not being applied in the titles for the 'nex't and 'previous' buttons in the pagination area. Only the aria-label were being applied. So, I made this correction to add both titles. Please, let me know if you need anymore information on this PR.

### DIFF
--- a/examples/text-localization/index.js
+++ b/examples/text-localization/index.js
@@ -49,6 +49,10 @@ class Example extends React.Component {
         body: {
           noMatch: "Sorry we could not find any records!",
         },
+        pagination:{
+          next: "Following page",
+          privous: "Preceding page"
+        },
         filter: {
           all: "All Records",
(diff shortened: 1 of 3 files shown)


patorjk:
Thank you for your contribution, I'll merge it in for the next release (should be within the next hour or so).

fluxquantum:
@Ruriel @patorjk Thanks for helping support this library. 

May I know if there's a way to directly access the nextIconButtonProps? 

I want to be able to override the 'disable' attribute for it. As one of my use cases is that I want to disable the pagination forward button when data is being fetched, this way I am not allow triggering of additional api calls until the initial call has returned.

Thank you for your time.

PR #1536, opened 2020-10-02
Allow custom react-dnd backend

butchmarshall:
This pull request actually has two desired outcomes

1) I'm using react-dnd in other components with a different backend for IE performance reasons - i'd like to use my custom backend everywhere

2) I was having issues initially and wanted to just disable loading the drag-and-drop component from even loading - this now enabled that by passing  components={{DragDropBackend: null}}


patorjk:
This looks good to me, thank you for the contribution! I'll get it into the next release.

gaut-b:
Hello all. It seems that the conditional calls to the "useDrag" and "useColumnDrop" hooks have been removed from the code during commit e75d2fe. Conditional calls to hooks are indeed not a good practice, but they allowed in this case to avoid an error when no drag and drop backend is provided ( components={{{DragDropBackend: null}}), which is quite a handy feature! Any chances that you might reconnect this feature ?

Issue #1604, opened 2020-12-11
Failed prop type: Invalid prop `isRowSelectable` of type `boolean` supplied to `TableSelectCell`, expected `function`.


tynindima:
> 

> 
When I use your example in console appears error:
![image](https://user-images.githubusercontent.com/58040069/101915531-68a6dd80-3bce-11eb-9b31-5775d1e93ff6.png)
## Expected Behavior

Console without errors
## Current Behavior

"Warning: Failed prop type: Invalid prop `isRowSelectable` of type `boolean` supplied to `TableSelectCell`, expected `function`." appears in console
## Steps to Reproduce (for bugs)

1. npm install mui-datatables --save
2. npm install @material-ui/core
3. npm install @material-ui/icons
4. adding simple table 
![image](https://user-images.githubusercontent.com/58040069/101916498-ae17da80-3bcf-11eb-9ce1-24f27e8f0974.png)

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.11.2     |
| MUI-datatables  |     3.7.2    |
| React        |     17.0.1    |
| browser      |    Chrome, Mozilla     |
| etc          |         |

lounsbrough:
We are having the same issue as of version 3.7.2

lounsbrough:
@tynindima - I have a PR up that should resolve this since it's causing issues for us as well.

scarabcoder:
I'm getting this issue with the latest version as well. 
I don't know if it's related, but I'm getting this error immediately before it:
![image](https://user-images.githubusercontent.com/4871271/101963508-c1bd4400-3bc3-11eb-87ff-e542eb42a5bb.png)

@lounsbrough Shouldn't it be a function though, and it's just being set to a bool incorrectly?

lounsbrough:
Yeah could be, not sure which is the expected behavior, but currently sibling components are providing booleans.

scarabcoder:
> Yeah could be, not sure which is the expected behavior, but currently sibling components are providing booleans.

I say that because `isRowSelectable` is in the API documentation as accepting a function. Seems like it should accept either a function or a boolean, but a more correct fix would probably be to supply `isRowSelectable` with something like `() => false`

lounsbrough:
True. I can make that change.

wdh2100:
@lounsbrough @scarabcoder 

Thanks!

Looking at the code, in `TableSelectCell` it should be set to Boolean.
[...]

PR #1605, opened 2020-12-11
Issue 1604: isRowSelectable should be PropTypes.bool

lounsbrough:
It seems that this would resolve issue https://github.com/gregnb/mui-datatables/issues/1604

### DIFF
--- a/src/components/TableBody.js
+++ b/src/components/TableBody.js
@@ -287,7 +287,7 @@ class TableBody extends React.Component {
                     selectableOn={options.selectableRows}
                     selectableRowsHideCheckboxes={options.selectableRowsHideCheckboxes}
                     isRowExpanded={this.isRowExpanded(dataIndex)}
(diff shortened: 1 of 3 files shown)


lounsbrough:
Runs fine locally, tests seem to maybe be in a bad state, I get exceptions when running. Looking at the tests I don't see anything I would expect to have broken with this change.

wdh2100:
@lounsbrough 

TableSelectCell is an internal component, so boolean is correct.

However, it may be confusing because it looks different from the document.
What do you think?

lounsbrough:
I think either way is likely ok, but boolean internally may be simpler.

wdh2100 (approved):
LGTM 👍

jeanatpi:
Maybe I'm wrong but this bug is still present in last release 3.7.6.

Is there any workaround?

Issue #1591, opened 2020-11-30
selectableRowsHideCheckboxes hides check boxes but still takes up space


aryo:
When `selectableRowsHideCheckboxes` is set to true, the first column containing the check boxes is still there (even though the checkboxes are no longer visible).

Is there a way to remove the whole column? or at least make it take up 0 space?

## Expected Behavior
Checkbox column does not take up space

## Current Behavior
Checkbox column takes up space even though checkboxes are hidden

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/muidatatables-custom-toolbar-kwemy?file=/index.js

Play around with `selectableRowsHideCheckboxes` option

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  | 4.11.0        |
| MUI-datatables  |   3.7.1      |
| React        |   16.13.1      |
| browser      |         |
| etc          |         |

aryo:
I've had to override the theme based on whether the `selectableRowsHideCheckboxes` option is true to get this to work. Ideally this should happen out of the box if the option is true.

```
MUIDataTableSelectCell: {
      root: {
        display: "none",
      }
    }
``

wdh2100:
https://github.com/gregnb/mui-datatables/blob/fde3497a552040505f70c274466715fce1d7e8ab/src/components/TableSelectCell.js#L165

I think it's in the wrong place.

wdh2100:
@aryo 

Fixed 3.7.2

aryo:
@wdh2100 Good stuff!

PR #1623, opened 2020-12-30
Fixed @material-ui/icons imports to allow tree-shaking

deloreanz:
## The Issue
Per Material UI's guide on minimizing bundle size, It appears a recent mui-datatables update [#1591](https://github.com/gregnb/mui-datatables/issues/1591) broke tree-shaking by importing material icons incorrectly.  This PR attempts to remedy that so projects importing mui-datatables aren't bloated by importing the full icon set.

Example:
```javascript
// uses tree-shaking, only imports the Remove icon
import RemoveIcon from '@material-ui/icons/Remove';
// imports all of @material-ui/icons (requires a Babel plugin to only include the Remove icon)
import { Remove } from '@material-ui/icons';
```

## Reasoning
I came upon this issue when trying to reduce the bundle size of one of my projects and found that the full icon set was still being included even though I followed the Material UI guide.  I searched through the entire bundle and found the cause.  The import syntax in  [#1591](https://github.com/gregnb/mui-datatables/issues/1591) can still allow for tree-shaking with the inclusion of a Babel plugin, but I figured the rest of mui-datatables used the explicit syntax which does not require any help from Babel to keep the bundle size minimal.

## Reference
Commit using unoptimized import syntax:
https://github.com/gregnb/mui-datatables/issues/1591

Material UI guide reference:
https://material-ui.com/guides/minimizing-bundle-size/

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,6 @@
   "version": "3.7.2",
   "description": "Datatables for React using Material-UI",
   "main": "dist/index.js",
-  "files": [
-    "dist"
-  ],
   "scripts": {
     "dev": "cross-env NODE_ENV=development webpack-dev-server -d --progress --colors",
     "test": "cross-env NODE_ENV=test mocha --require @babel/register test/**/*.test.js",
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "3.7.2",
   "description": "Datatables for React using Material-UI",
   "main": "dist/index.js",
-  "files": [	
-    "dist"	
+  "files": [
+    "dist"
   ],
   "scripts": {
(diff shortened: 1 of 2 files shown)


wdh2100:
thanks!

wdh2100 (approved):
LGTM 👍

package.json
```diff
@@ -3,9 +3,6 @@
   "version": "3.7.2",
   "description": "Datatables for React using Material-UI",
   "main": "dist/index.js",
-  "files": [
-    "dist"
-  ],
```
wdh2100, line 8: What's the reason?

deloreanz: Sorry I accidentally pushed an update to this branch when I was testing something else.

wdh2100:
@deloreanz 

thanks!

Will be released soon.

PR #1630, opened 2021-01-08
Fix sort header: duplicate tabindex, focus/hover style issues

### LINKED ISSUE #1629: Sortable headers have tabindex=0 on both span containing header/button and button, causing you to have to tab twice through the headers
## Expected Behavior
You should only have to tab through each header once to be able to get to a different header to sort. Additionally, the headers should have a focus style so that it is clear which is focused when using keyboard navigation.

## Current Behavior
You have to hit tab on each header twice, since both the MUIDataTableHeadCell-toolButton and the MUIDataTableHeadCell-sortLabelRoot have `tabindex="0"`. When tabbing, it is unclear which header you are tabbed to (except that you can see the tooltip).

https://codesandbox.io/s/muidatatables-custom-toolbar-rvmcj?file=/index.js

1. Ensure the example is displaying the non-responsive view
2. Click into the example area
3. Press Tab until you see the _Sort_ tooltip. Press enter to sort by the first column.
4. Press Tab once more. Notice there is no tooltip. Press enter to sort by the second column.
5. Press Tab once more. Notice the _Sort_ tooltip. Press enter and notice that you again sort by the second column.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  | 4.10.2 |
| MUI-datatables  | 3.4.1 |
| React        | 16.9.0 |
| browser      | Chrome 87.0.4280.88 |

pkmnct:
Fixes #1629
- Use Material UI Button for sort (adds focus, hover styles)
- Fix help and arrow alignment
- Center tooltip on button
- Fix duplicate `tabindex`

### DIFF
--- a/src/components/TableHeadCell.js
+++ b/src/components/TableHeadCell.js
@@ -113,6 +113,7 @@ const TableHeadCell = ({
 
   const sortLabelProps = {
     classes: { root: classes.sortLabelRoot },
+    tabindex: -1,
     active: sortActive,
     hideSortIcon: true,
     ...(ariaSortDirection ? { direction: sortDirection } : {}),
--- a/src/components/TableHeadCell.js
+++ b/src/components/TableHeadCell.js
@@ -1,3 +1,4 @@
+import Button from '@material-ui/core/Button';
 import clsx from 'clsx';
 import HelpIcon from '@material-ui/icons/Help';
 import MuiTooltip from '@material-ui/core/Tooltip';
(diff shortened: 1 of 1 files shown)


src/components/TableHeadCell.js
```diff
@@ -197,37 +203,38 @@ const TableHeadCell = ({
...
-          onClick={handleSortClick}
-          className={classes.toolButton}
-          data-testid={`headcol-${index}`}
-          ref={isDraggingEnabled() ? dragRef : null}
-          tabIndex={0}>
+        <span className={classes.contentWrapper}>
           <Tooltip
             title={getTooltipTitle()}
-            placement={'bottom-start'}
+            placement="bottom-center"
```
wdh2100, line 209: For what reason?

pkmnct: This was to better line up the tooltip with the new Material-UI button added for the focus/hover effects. I can revert this if we want to leave default alignment to bottom-start.

wdh2100: It looks good to me too

wdh2100 (approved):

Issue #1609, opened 2020-12-15
Using a different icon package?


quinnter:
Is it possible to use a different icon library? I see that material/icons is a peer dependency, but my project uses mdi/js (so, nearly the same). I am currently using material-table and I am looking to switch over to something different, mostly for bundle size reduction but the package has also been abandoned! 

Thank you

Issue #1048, opened 2019-11-04
How to change action icons


zilkenberg22:
How to change action icons 

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.3.3     |
| MUI-datatables  |   2.10.3      |
| React        |  16.9.0       |
| browser      |    google Chrome     |
| etc          |         |

gabrielliwerant:
Unless you specify all your own actions via `customToolbar`, there isn't yet a way to do it. I'm open to proposals for how to do so here https://github.com/gregnb/mui-datatables/issues/198.

BjoernRave:
@gabrielliwerant if I would create a custom toolbar, is there a way to get access to the download, filter etc. functionality ? Meaning can I create my own button and pass it an onClick handler to download the csv?

gabrielliwerant:
@BjoernRave Unfortunately, individual functions and components aren't importable on their own at the moment, but I'm open to PRs that would provide this functionality.

Issue #198, opened 2018-09-20
Is there anyway I could change the icons in the action panel?


susmita-vps:
I would like to change the 'export to csv' icon to a standard download icon. Currently it uses the download from cloud icon. Thanks.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |         |
| MUI-datatables  |         |
| React        |         |
| browser      |         |
| etc          |         |

gregnb:
Not currently but I am working on changing the API around a bit to allow for more flexibility

susmita-vps:
Hi Greg, for a short-term is it possible to change the download icon to a standard one (with a simple downarrow) rather than the "cloud_download"? Thanks.

PR #1639, opened 2021-01-25
Added ability to use custom action toolbar icons

lancerael:
We really want to use this library but we need the icons to match the rest of our UI. I have seen multiple requests for this feature including a request for help. So here is my proposed solution. With this PR we can now pass custom Icons which will optionally render whatever component is passed in its place.

* Added ability to use custom icons.
* Added unit test suite for new functionality.
* Updated documentation in readme.
* Added example to the dev build.

image

* Fixes https://github.com/gregnb/mui-datatables/issues/1609
* Fixes https://github.com/gregnb/mui-datatables/issues/198
* Fixes https://github.com/gregnb/mui-datatables/issues/1048


wdh2100:
@lancerael  Thank you. I'll take a look

amanicom:
I would like to use another icon for the help icon which is used in this project for the column. Is there a way to use the custom icon for the hint icon used ?

amanicom:
I would like to use another icon for the hint option instead of the HelpIcon(?). Is there a way to use the custom icon for the hint option ?

PR #1677, opened 2021-03-07
No dnd backend provided

gaut-b:
This PR follows #1536 which allows to manage cases where no DND backend is provided. 

The original PR conditionally called the hooks "useDrag" and "useColumnDrop" in the TableHeadCell component, but it seems that these changes have been overwritten in commit e75d2fe. Therefore, a bug appears in the TableHeadCell component if no DnDbackend is provided (which is quite a neat feature).

So I propose a quickfix in which I created a new component named TableHeadCellNoDnd which is called if dragNDrop is not explicitly enabled.

I added a working example of a table to which is passed "components={{DragDropBackend: null}}".

### DIFF
--- a/examples/examples.js
+++ b/examples/examples.js
@@ -35,6 +35,7 @@ import CustomComponents from './custom-components';
 import InfiniteScrolling from './infinite-scrolling';
 import Themes from './themes';
 import LargeDataSet from './large-data-set';
+import NoDND from './no-drag-and-drop';
 
 /**
  * Here you can add any extra examples with the Card label as the key, and the component to render as the value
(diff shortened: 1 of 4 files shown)


wdh2100:
@gaut-b 

That's a good opinion.
However, if there is one thing that takes, there are a lot of duplicate codes.
How about making it with Higher-order Component?

wdh2100:
I will consider whether it is possible to change it in the current structure (TableHeadCell - TableHeadCellNoDnD)

gaut-b:
Thanks for the review. This PR was a quickfix and indeed not very "DRY". HOC is indeed a good idea and I will try to get back to you in the next few days with a proposal.

However, the cleanest way in my opinion would be to put back the conditional calls on the hooks following the value of isDraggingEnabled.I don't think there are cases where a backend is provided or not dynamically, so a conditional call on hooks should not be a problem.

Issue #1686, opened 2021-03-24
Hide search Icon and leave search bar on


limsc1017:
My table options

options:{
    search: false, 
    searchOpen: true,
    ...
}
I want the hide the icon only and display the search bar

## Expected Behavior
search: false + searchOpen: true

Should hide the icon + display the search bar 

## Current Behavior
search bar and icon are both removed

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/vigorous-haslett-40jlf?file=/src/App.js

This sandbox is similar to my table, I have no other buttons so there is nothing to close the search bar.  hence I don't need the icon to toggle it.

im forced to put search: 'disabled' at the moment.

Issue #1125, opened 2019-12-16
Always-on Search field


twpopcornchkn:
When "searchOpen" option is set to true, is it possible to truly replace the title with the search field (title is not in use in my use case). 
Because when the "viewColumn" or "filter" pop-over are opened, search field will be hidden, and it will have to turn back on by clicking the search icon again, which can be annoying for the users. 

## Expected Behavior

search field always-on. 

## Current Behavior

search field will be hidden when other action icon (filter, view column) are clicked.

## Steps to Reproduce (for bugs)
code sandbox
https://codesandbox.io/embed/muidatatables-custom-toolbar-q078r?fontsize=14&hidenavigation=1&theme=dark

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.6.1     |
| MUI-datatables  |   2.12.4      |
| React        |     16.9.0    |
| browser      |  Chrome       |

gabrielliwerant:
So, the search bar does stay open if there's text in it. If there's nothing in it, it closes when opening up another field, since you can't interact with both at the same time.

The table currently isn't designed for a search bar that is always open, no matter what, since this will hide the table title, which is not ideal. If we add this behavior, we should reconsider the placement of the search bar and/or title so that both can be visible.

FYI, you can fake always open behavior of the search bar by setting `searchText` to equal a space character. I know this is not ideal for you, but could work in the mean time.

Brmm:
I have the exact same use case. Can confirm setting searchText to ' ' never closes the search field, this combined with a customSearchRender() that handles the search functionality works for me.

honi:
Somewhat related, I would argue that handling the title as a required prop is incorrect. Sometimes you already have a title elsewhere in the page and you just want to display the table. 

In this scenario the search can easily be always visible to use that space, regardless if it's unusable while opening the filters or other elements. That's going to happen with any text field, if you open a drawer, popup, etc. you won't be able to use a text field that's underneath, and in those cases you probably are not hiding the text field anyways.

Dragomir-Ivanov:
I second what @honi said. I do have full screen table, with title already presented in anothert UI control. Just want the search field to be always on. Thanks!

gabrielliwerant:
The title field is not required, it is optional, but the behavior I want to have is the ability to have both a title and an always on search field, so I'm open to pull requests which will handle both of those things.

honi:
[...]

PR #1691, opened 2021-03-30
Search always open

lbressler13:
Related to #1125 and #1686

This adds the option to have the search bar always open. This sets the search bar to be open and removes the close option. It also removes the search icon from the toolbar, since that would allow the search bar to be closed. An alternative would be to keep the icon but disable its on click, would be happy to make that change if that's preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ The component accepts the following props:
 |**`searchPlaceholder`**|string||Search text placeholder. [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-search/index.js)
 |**`searchProps`**|object|{}|Props applied to the search text box. You can set method callbacks like onBlur, onKeyUp, etc, this way. [Example](https://github.com/gregnb/mui-datatables/blob/master/examples/customize-search/index.js)
 |**`searchOpen`**|boolean|false|Initially displays search bar.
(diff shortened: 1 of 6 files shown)


tea-lover-418:
Happy to see this PR as this is a functionality I've been looking for.

LuisAlejandro:
I hope this gets merged soon.

MendelBak:
Is this library still maintained? It looks like no new PRs have been closed in a while.

entozoon:
@lbressler13  This still closes the search bar when you open filters (in which case you're screwed as there's no search button either).

Regardless, it's a shame not to have been merged

wdh2100 (approved):
LGTM 👍 

@lbressler13

Forgive me for being late.
Thanks

entozoon:
Nice one @wdh2100 ! Perhaps leave the issue open for now, due to the bug I mentioned above? It's not _quite_ perfect but workable :)

lbressler13:
> @lbressler13 This still closes the search bar if it's blank when you open filters (in which case you're screwed as there's no search button either).
> There's a bit of a workaround, setting `searchText: " ",` in options so that it's never blank
> Regardless, it's a shame not to have been merged.

Thanks for the catch, I'll work on fixing this

PR #1692, opened 2021-03-31
fix issues with always rendering

brandonseydel:
Lots of  unecessary expensive re-renders are happening because of always using () =>


wdh2100:
> Lots of unecessary expensive re-renders are happening because of always using () =>

@brandonseydel 

Could you explain a little more about why?

PR #1698, opened 2021-04-09
Removing top level imports to reduce bundle sizes.

tylerg-smt:
I performed a text search for anything importing `from '@material-ui/core` and replaced the results with second level imports, so that the entire @material-ui/core directory is not pulled into bundles.


wdh2100:
thanks!

tylerg-smt:
@wdh2100 What are next steps? Does someone need to review the PR?

wdh2100:
@tylerg-smt 

I'll release it this week With a other fix bugs

PR #1711, opened 2021-05-05
cut out space in data-testid for automation tools

gaberoo322:
my company's QA team couldnt select the download csv button using the cypress automation testing tool because there was a space in the localization config of downloadCsv.

this fix makes it so that you can use whatever localization text you want and remove the whitespaces only in the data-testid attributes.


wdh2100 (approved):
LGTM 👍

Issue #1770, opened 2021-09-01
search Field disappears with option searchAlwaysOpen


Cantasura:
## Expected Behavior
Search field should always be displayed when the option is set to "true"

## Current Behavior
Search field disappears as soon as "viewColumns" or "filter" are clicked

## Steps to Reproduce (for bugs)
https://codesandbox.io/s/muidatatables-custom-toolbar-forked-gc95n?file=/index.js
1. click on the e.g. "viewColumn" - Button

## Your Environment

| Tech         | Version |
|--------------|---------|
| Material-UI  |    4.12.3     |
| MUI-datatables  |  3.7.8       |
| React        |   16.9.0      |

PR #1771, opened 2021-09-01
fix: searchAlwaysOpen when clicking other filters

Cantasura:
#1770 
I think this line should do the trick, but not sure if  I missing something


ashfaqnisar:
@Cantasura @wdh2100 can you close this issue as it is fixed!

PR #1879, opened 2022-03-04
place component in isolated stacking context

berzi:
If the table is placed near tooltips or other elements with z-indexes, some parts of the table (notably the header) can overlap in unexpected ways. Placing the component in an isolated stacking context with `isolation: isolate` makes sure that the z-indexes inside the component only interact with other indexes inside the component, causing the component as a whole to be treated as a single object, which is more coherent and desirable in practically all situations.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -27,7 +27,9 @@ import { load, save } from './localStorage';
 
 const defaultTableStyles = theme => ({
   root: {},
-  paper: {},
+  paper: {
+    isolation: 'isolate',
+  },
   paperResponsiveScrollFullHeightFullWidth: {
     position: 'absolute',
   },


berzi:
@gregnb sorry for the notification, but I'm not sure whether this PR has been seen at all and I don't know how much time I should expect before acknowledgement: this is my first contribution here.

patorjk:
Looks like there's quite a few uses of z-index:

https://github.com/gregnb/mui-datatables/search?q=zIndex

This is a good PR, as it would help against z-index inflation.

patorjk:
I'm going to merge this. I think this is a good idea, and I checked it to be sure there were no side effects. Not sure when it'll get deployed, but if and when there's another release it'll be in there. Thank you for the PR.

PR #1892, opened 2022-03-17
ToolbarSelect should always shown

MarkSinD:
Is there a way to show TableToolbarSelect always?
I redefined the TableToolbarSelect component, now I have my own logic for this component, but I also need it to always be on the screen, and it appears only when selecting an object from the table.
I saw the selectToolbarPlacement property that is responsible for showing TableToolbarSelect. Added the always property there. Please do so that I can customize my TableToolbarSelect and be able to show it always.


patorjk:
This looked fine to me, I added it into the current release.

PR #1947, opened 2022-08-24
Update to tss-react v4

garronej:
Hi,  
With `tss-react` v4, the fact that mui-datatables uses `tss-react` is completely transparent for the end user 🥳.


patorjk:
Thank you for this PR, added it for the current release.

PR #1995, opened 2023-03-25
Allow user to customize the hardcoded wrapper Component.

ensemblebd:
Allow user to customize the hardcoded wrapper Component.
Without this, user has no control over the overflow wrapper, and thus cannot implement custom scroll systems, or drag systems, etc.

ie: 

```
	...

	return (
		<MUIDataTable
			data={props.data.Entries}
			columns={columns}
			options={mui_table_options}
			components={{
				ReactiveWrapper: DragScrollable,
			}}
		/>
	);
}

const DragScrollable = (props) => {
	const classes = useStyles();
	const ref= React.useRef(null); 
	const { events } = useDraggable(ref); // https://www.npmjs.com/package/react-use-draggable-scroll

	const {style, className} = props; // these are fed from MuiDataTable

	return (
		<div 
			ref={ref} 
			{...events} 
			className={cx(classes.dragContainer, className, classes.scrollHide)} 
			style={{width: 'calc(100% - 9px)', height: 'calc(100% - 9px)', ...style}}
		/>
	);
};
```

For me this was of particular interest so as to implement https://www.npmjs.com/package/react-custom-scrollbars-2 and https://www.npmjs.com/package/react-use-draggable-scroll. 
Now when I set `tableBodyHeight`, the browser default scroll isn't used, rather my own system.
But this could of course be used for any number of purposes.

### DIFF
--- a/src/MUIDataTable.js
+++ b/src/MUIDataTable.js
@@ -11,6 +11,7 @@ import isUndefined from 'lodash.isundefined';
 import merge from 'lodash.merge';
 import PropTypes from 'prop-types';
 import React from 'react';
+import DefaultReactiveWrapper from './components/ReactiveWrapper';
 import DefaultTableBody from './components/TableBody';
 import DefaultTableFilter from './components/TableFilter';
 import DefaultTableFilterList from './components/TableFilterList';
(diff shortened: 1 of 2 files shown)


ensemblebd:
Closing as I've begun merging PR's myself into my fork, and thus my Diff no longer applies to this PR. 
Additionally it does not appear that this project will ever see another update, so what's the point of PR's.

patorjk:
I'm a maintainer, but I'm no longer active (I don't have the time and I don't use this project anymore). If you're actively working on a fork, I would recommend contacting the repo owner on Twitter/X and asking if you could be a maintainer. That might be better than working on a fork (others may be interested in helping you develop it).

PR #2046, opened 2025-01-19
Upgrade `react-dnd` to v16 and import type of `<TableHeadCell />`

### LINKED ISSUE #108: Custom Row/Footer
Question:

Is it possible to add a custom footer to a datatable? At the bottom of a datatable I need to add a row with "Totals" that needs to be fixed (always be visible on every page even when pagination is on), but not be filterable/sortable. I've tried creating a separate datatable component below the main table (and styling it to look as part of the main table), but there is a scrolling issue in responsive: when the main table is scrolled horizontally, last row is not scrolled with it - so not a good UI experience. Any suggestions on how it can be achieved?

This is what I need:
image

Thanks!

dominosaurs:
Fix #108


dominosaurs:
wrong PR

